PR #4, opened 2018-05-07
Changes round end restart timer to 2 minutes

GrayRachnid:
3 minutes felt a bit too long.


GrayRachnid:
I doubt this is getting touched by any mirror PR so i'll just add a comment instead of modularizing.

PR #24, opened 2018-05-15
[MIRROR] Adds Micro mechs

CitadelStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/3603
--------------------
adds in construction paths for the 3 current micro mechs along with the parts
also adds the ability to build the equipment for micro mechs (req tech and rnd levels can change upon suggestion)
adds in the crates for the relevant circuits along with adding them to the circuit printer
the gopher (mining micro mech) has the same building steps as the ripley and the 2 security micro mechs (polecat and weasel) are the same as the durand or gygax with the polecat using a special armor plate

if you want me to change any of the material or research values just ask


deathride58:
Build failed due to RNG in map generation

PR #28, opened 2018-05-15
Minor map edits

GrayRachnid:
- Makes Research Surface exit smaller, it was too big and that made it take too long to drain all the phoron, causing people to force doors open in impatience when phoron was at a low amount(hint, ZAS doesn't care.)
- Gives atmospherics one more scrubber and vent, I found it kind of weird that atmos only has 3 pairs with all the area it covers and two floors(with ZAS in mind)
- Adds a guest pass terminal in front of medbay.


GrayRachnid:
reeeeeeeeeeeeeeee

PR #151, opened 2018-06-19
Changes the three fully visible leg-based holsters to be usable with webbing

jakeramsay007:
Honestly there's no point using holsters except for carrying a single gun, when webbing already does what they do but better. This at least lets the non-concealed holsters be usable with webbing for more realism and utility, especially for security and explorers.

The three non-concealed holsters are the Leg, Hip and Waist holsters.


Cw3040:
yes

deathride58:
Did you test this to make sure it doesn't cause any weird bugs when equipping webbing and a non-concealed holster at the same time?

jakeramsay007:
@deathride58 It uses the exact same method as the machete holster, no weird bugs except for a message saying you can't holster this item when you click on your jumpsuit to put an item in your webbing

PR #160, opened 2018-06-28
Meteor Change (Speedmerge this, please)

Cw3040:
-Removes meteors from rotation, they crash the serb


CameronWoof:
This affects the wrong file. Check the _vr override.

Cw3040:
shhhhhshhhhshhhh i'm retarded

PR #161, opened 2018-07-05
Fixes a typo with cups of lime tea.

avabee:
It would erroneously be named "cup of berry tea" despite being full of lime tea. This fixes that.


avabee:
I made the PR upstream at Polaris instead. Should reach here soon.

PR #182, opened 2018-07-10
Removes whitelist restrictions from clowns and mimes

deathride58:
They're still latejoin-only. Honk.


Cw3040:
please no

deathride58:
The events that unfolded as a result of this PR's testmerge serve as an example of why even immortal machines should never be trusted to play god

PR #184, opened 2018-07-10
Manual sync with upstream

deathride58:
Closes all the active mirror PRs
Reverts our specific map changes, though they should be easy to readd later since they're minor


deathride58:
This PR has stagnated pretty badly. Replaced by #255

PR #185, opened 2018-07-12
Re-adds EAL language implants.

avabee:
They now cost double the points to purchase, because apparently VORE had issues with them, and I'm assuming in good faith they had at least somewhat of a decent reason for removing them.
(Though it is rather ironic that Virgo removed a widely-liked feature for balance issues.)


avabee:
Ah, I used the web UI. Good catch, though.

Mewchild:
The translator earpiece was also removed at the same time as the EAL implant. Removal reason was wanting to make languages more niche, and make it so people couldn't understand everything.

jakeramsay007:
@Mewchild translator earpieces can be produced in RnD

Mewchild:
They used to be selectable in the loadout, for 4 points.

PR #186, opened 2018-07-14
Supply Changes

Cw3040:
Changelogs in the Commits. This one's for you, Hyena.


Cw3040:
wish i could add meme pr label but DNM

Cw3040:
so this is to remove KHI NIFs from the cargo console, as they caused about a metric fuckton of salt

PR #237, opened 2018-07-26
Adds alternative titles to the visitor role

NopeMcHaltman:
:cl: Anonymous
tweak: Gives a few alt. titles for the visitor.
/:cl:

Feel free to suggest more titles.


modular_citadel/code/game/jobs/job/assistant_cit.dm
```diff
@@ -0,0 +1,25 @@
+//
+/// Adds some alt titles for the visitor.
+//
+
+/datum/job/assistant
+	..()
```
deathride58: No need for the ..() here, as it'll cause issues with mapdiffbot

PR #244, opened 2018-08-02
Adds a alternate job title for security officers

Wilchenx:
that noone asked for.


Wilchenx:
should probably cap it to one person, cause thats all there is needed to guard the cells

modular_citadel/code/game/jobs/job/security_cit.dm
```diff
@@ -0,0 +1,4 @@
+// Per request
+
+/datum/job/officer
+	alt_titles = list("Brig Officer")
```
deathride58: This will override the "Junior officer" alt-title. Readd the alt-title that was overridden

Wilchenx:
@deathride58 i think making this into a new job is probably a better idea

PR #245, opened 2018-08-02
Allows me to be a clown

Wilchenx:


deathride58:
This PR will do literally nothing if merged, as the server config is completely separate from the config on the repo.

PR #248, opened 2018-08-03
Adds cash registers and the Sec RIG suit.

GrayRachnid:
Cash registers added to Kitchen, Bar, and Cafe desks. 
Security RIG added to Sec EVA. Expanded it by 2 tiles outward so sec now has 2 voidsuits and the RIG.


GrayRachnid:
bleh I did an error setting up mapmerger.

PR #249, opened 2018-08-03
Adds Cash Registers and the Sec RIG

GrayRachnid:
Cash Registers added to Bar, Kitchen, and Cafe.
Sec RIG added to Sec EVA. Expanded by two tiles to have a total of 2 Voidsuits and 1 RIG.


GrayRachnid:
Sec RIG was added in the new update, i'll add the cash registers in a seperate PR.

PR #254, opened 2018-08-11
Protean de-tardization

Cw3040:
-Proteans can no longer ventcrawl


C0rva1r:
We can make a modified more player friendly version of the protean. Keeping the original for testing and while the other can be whitelisted.

PR #291, opened 2018-08-20
[MIRROR] Improves Tesla Engine submap

CitadelStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/4179
--------------------
https://i.gyazo.com/af76ea6d1c5f392d287d0b96246ee25d.png

Removes redundant radiation collectors and wires, makes the force field area more compact, and adds an extra set of tesla coils to the default startup. Generally makes it a lot cleaner and sensible.


deathride58:
Compile failed due to RNG on both checks

PR #300, opened 2018-08-23
Security and Command bowmans now provide protection from flashbangs

jakeramsay007:
Exactly as it says on the tin.


improvedname:
>buffing security
*furiously writes a novel about validhunting in rp*

PR #302, opened 2018-08-25
[S] patches an exploit that allows skids to spawn as any species

deathride58:
Title. Will send to Polaris when i'm not on the verge of passing out. Contains cit edit comments in case of weirdity


king5327:
I take significant offence to your side's branch name.

PR #304, opened 2018-08-25
Translocator corpreg conundrum corrected

jakeramsay007:
Corp Regs and the item description were in conflict, so I fixed it to be 100% more reasonable


deathride58:
The map change will probably get overwritten by upstream map syncs but eh, that's what happens when they're using varedits to handle fluff papers.

PR #306, opened 2018-08-30
Adds a (for now) admin-only consume objective

deathride58:
Title.

When the consume objective is added, it'll make a list of everyone that has their event preference set to "Be Prey" and choose a random person from that list.

The objective will be marked as complete so long as the target is alive and inside of a stomach. Simple stuff really.

This PR is also untested.


Poojawa:
ensure the objectives change if the target cryos out.

nik707:
reopen when we wanna make this work and shit

PR #312, opened 2018-09-04
Virgo Four's water is now in fact wet

jakeramsay007:
Map dif is correct, a shitload of tiles were edited in order to actually make the water wet.


jakeramsay007:
Travis error unrelated to PR

nik707:
Why was this never merged?

PR #351, opened 2018-09-29
Disables salt

GrayRachnid:
You'll have to use something else for seasoning now.


lolman360:
Wow, what a useful pr that removes a useless and abused feature that many agreed is now unneeded and/or overpowered! Good job, Gray.

PR #353, opened 2018-10-02
Fixes an HREF exploit usable to unlock an RnD console without access.

king5327:
While I hate both the HREF fixes' branch naming rules and the modularization, the goal here is to get it over and done with.

Also prevents unlocking the RnD console unless you actually have RnD lab access and not just science front door.


kevinz000:
Topic is not something that should be modularized as this will prevent it from logging on failure.
Please send this upstream to Polaris and virgo (optimally PR on both, it's a 1 line change basically)

kevinz000:
modularized in this way anyways

king5327:
Upstream PR's merged on both Virgo and Polaris. https://github.com/VOREStation/VOREStation/pull/4368

PR #379, opened 2018-10-05
Adds dark long cloak + goliath cloak

NopeMcHaltman:
:cl: Anonymous
add: Added cape-- I mean long cloak. Dark one so don't expect to play hero with this. You can get it from loadout.
add: Ported and adapts Goliath cape into cloak. Also includes "replica" of it for loadout.
/:cl:

[why]: # (Please add a short description [two lines down] of why you think these changes would benefit the game. If you can't justify it in words, it might not be worth adding.)
Decided to make something for Spooktember and upcoming Halloween. It's bad, it's ugly, it's not nicely coded but hey, at least it works.
Ignore hood's blackness, it is acts properly like chaplain's one in-game.
cape_front_preview


NopeMcHaltman:
Okay this looks like spaghetti code. I'll better off learning how to make colorable cloaks for general use. If somebody really wants this instead then, as usually, yell at me.

PR #394, opened 2018-10-09
[DNM] Adds cold protection to the ushanka and beanie, plus replaces ushanka item sprite

NopeMcHaltman:
:cl: Anonymous
tweak: Adds cold protection to the ushanka and beanie.
imageadd: Ported ushanka item icon from tg, modularized code will use it instead of crappy one.
/:cl:

Per request from PantherPusher over Discord. Values i used are quite debatable, though.


nik707:
If you're planning to fix this lemme know, otherwise I'm gonna close it if I don't hear back.

NopeMcHaltman:
@nik707 Can't fix the line-ending inconsistency and thus there's bullshit changes with .dme file, consider merging #457 instead.

PR #407, opened 2018-10-11
Cafe map fixes and additions

avabee:
* Adds a cash register to the surface 1 cafe.
* Removes a duplicate chair.


nik707:

PR #420, opened 2018-10-17
Enables the off-duty CO position

PolymorphBlue:


CameronWoof:
Not super thrilled with the idea of giving an Off-Duty position bridge access. The bridge is a high-security location and staff assistants shouldn't be in there.

kevinz000:
Eh we can remove bridge if there's issues

CameronWoof:
Do you not feel that "greyshirts having access to all the consoles in the bridge" is an issue? That's a lot of power to give someone with zero responsibility.

Useroth:
Gotta side with Cameron here, tbh.
It should be only cosmetic, imo, like all other off-duty roles

PR #423, opened 2018-10-19
*flip and *snap both now have exponentially increasing break chances.…

PolymorphBlue:
…  The break chance decays at 1% per second, and snap has a floor so you can snap twice with no risk.


Andromeda-K22:
why is this required

why not literally just add a cooldown instead of a chance to fuck over limbs

PolymorphBlue:
I guess because there was already the break chance in there from flipping, it still allows you to do something like snap twice in quick succession if you want, and... it's funny.

PR #426, opened 2018-10-20
Reduces rag reagent capacity

CameronWoof:
From 10 > 1.

Turns out instant blood reagent transfer isn't all that balanced who knew. Side effect: curbs medical powergay smothering patients with rags to apply meds and avoid syringes.


LetterJay:
Lol

PR #430, opened 2018-10-23
Adds a new corridor to Surface 2 from the stairs, and fixes Surface 3 airlock name

Alexch2:
Adds a new corridor to Surface 2 past the janitor's closet from the smaller staircase, shown below.
Also fixes the name of the airlock on surface 3's double-wide staircase from just "Glass Airlock" to "Atrium Third Floor," in line with the other airlocks around it.
image


nik707:
Keep a copy of this map change on hand, because the hardsync will break the fuck out of it I bet and you'll need to slap this back over. Will approve at that point.

nik707:

PR #431, opened 2018-10-24
[Ready] Fixes securitron sound circuit 

Trilbyspaceclone:
[What dose this do!]
Makes the Securitron Sound Circuit sec tech and not "Illegal" 
[BUT WHY?]
I think it was never ment to even be illegal, its a sec board for sec robots that sec has. It being against NT and CC, makes no real logic when you even don't think about it, RP wise it says "ROBOTS THAT NT HAS ARE ILLEGAL!" for no reason!

### DIFF
--- a/code/modules/integrated_electronics/subtypes/output.dm
+++ b/code/modules/integrated_electronics/subtypes/output.dm
@@ -197,7 +197,7 @@
 		"secure day"	= 'sound/voice/bsecureday.ogg',
 		)
 	spawn_flags = IC_SPAWN_RESEARCH
-	origin_tech = list(TECH_ENGINEERING = 2, TECH_DATA = 2, TECH_ILLEGAL = 1)
+	origin_tech = list(TECH_ENGINEERING = 2, TECH_DATA = 2, TECH_COMBAT = 1)
 
 /obj/item/integrated_circuit/output/sound/medbot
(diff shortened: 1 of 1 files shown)


avabee:
First, I believe it has illegal tech because of the fact that it has the emagged voice line included by default (which is a long string of profanities). Second, this does sort of fuck over the RnD meta, making maint-crawling a requirement for illegal tech. Even if it is intended to shake up the RnD meta, it doesn't do a good job of it as it only means you have to wait 30 seconds for someone to drop a broken cryptographic sequencer off at the RnD desk.

Trilbyspaceclone:
@avabee 
// So Beepsky talks while beating up simple mobs.
/mob/living/bot/secbot/proc/insult(var/mob/living/L)
	if(can_next_insult > world.time)
		return
	var/threat = check_threat(L)
	**if(threat >= 10)
		playsound(src.loc, 'sound/voice/binsult.ogg', 75)**
		can_next_insult = world.time + 20 SECONDS
	else
		playsound(src.loc, pick(fighting_sounds), 75)
		can_next_insult = world.time + 5 SECONDS

lolman360:
It's intended to be illegal tech, yes, so that science can max RnD without antag items/maintcrawling.

PR #437, opened 2018-10-28
Allows bags to fit on EVA hardsuits.

Hatterhat:
vests was big shit maybe this will be less big shit


lolman360:
a surpising amount of hardsuits don't let this happen actually

Hatterhat:
i mean it's just like this because they're stationside RIGs and you're probably not gonna get the others. god maybe i can just make backpacks fit on every RIG suit storage but that'll be a big fuckin pain in the ass my dude

PR #438, opened 2018-10-29
Increases round-length by one hour

nik707:
Held a vote in the RP democracy channel, and the extend round by one hour vote won by a healthy margin. Assuming I did this right, this should now have the extend vote occur at the 4 hour mark.


Useroth:
How about making it 3 hours + 2 extension votes, rather than hard 4 hours?

nik707:
This is what the vote landed on, if you want to do that you're welcome to but this was what people asked for so I went with it.

Useroth:
Not that I mind, tho, but the vote didn't have a lot of options to choose from. A bit more flexibility would be nice, in the thing, overall.

PR #439, opened 2018-10-30
Improved Quick NIF again, now modularized

Useroth:
Basically adds a prompt that lets you choose the NIF type to implant in the target.

### DIFF
--- a/modular_citadel/code/modules/admin/verbs/debug_vr.dm
+++ b/modular_citadel/code/modules/admin/verbs/debug_vr.dm
@@ -1,0 +1,51 @@
+/datum/admins/proc/quick_nif()
+	set category = "Fun"
+	set name = "Quick NIF"
+	set desc = "Spawns a NIF into someone in quick-implant mode."
+
+	var/input_NIF
+
+	if(!check_rights(R_ADMIN))
+		return
+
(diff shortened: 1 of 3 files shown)


modular_citadel/code/modules/admin/verbs/debug_vr.dm
```diff
@@ -0,0 +1,51 @@
...
+	if(H.species.flags & NO_SCAN)
+		var/obj/item/device/nif/S = new /obj/item/device/nif/bioadap
+		input_NIF = S.name
+		new /obj/item/device/nif/bioadap(H)
+	else
+		var/list/NIF_types = typesof(/obj/item/device/nif)
+		var/list/NIFs = list()
+
+		for(var/NIF_type in NIF_types)
+			var/obj/item/device/nif/S = new NIF_type
```
king5327: You're making one completely new NIF of every type just for this single proc. I've also got my own contestant for this coming in with the other NIF changes.

Useroth: Decent catch, tbh, about the new thing.
I just copied my earlier code without re-reading it, just made sure it compiles and works fine.

PR #441, opened 2018-10-31
Spy Kits Extended

Useroth:
Adds radio functionality to spy bugs and bug monitors.
No additional steps necessary to use it. Each monitor gets assigned a random frequency and each bug automatically transmits at that frequency after being linked.

TODO:
- Make emotes visible through cameras.
- Make the bugs capable of catching whispers and subtle emotes as well.


nik707:

PR #445, opened 2018-11-04
adds sound for awoo, adds nya; ported from main

realbuffalowing:
much needed emotes for the aspiring diet furry


code/modules/mob/living/carbon/human/emote_vr.dm
```diff
@@ -17,8 +17,13 @@
 			message = "mlems [get_visible_gender() == MALE ? "his" : get_visible_gender() == FEMALE ? "her" : "their"] tongue up over [get_visible_gender() == MALE ? "his" : get_visible_gender() == FEMALE ? "her" : "their"] nose. Mlem."
 			m_type = 1
 		if ("awoo")
-			message = "awoos loudly. AwoooOOOOoooo!"
```
nik707: Please don't put us back to the low tier "lets out an X"

realbuffalowing: this is so sad can we get 10 likes alexa play despacito

PR #449, opened 2018-11-08
Removes Wish Granters

CameronWoof:
Why do people make these? 

Removes the syndicate beacon from spawning, which is an item that allows you to become a traitor when used. Because wish granters were too LRP even for Main.


nik707:

PR #453, opened 2018-11-09
Re-adds a singular type of fluff robolimb

jakeramsay007:
Because trying to justify anything is like repeatedly smashing your head against a brick wall


nik707:

PR #301, opened 2018-08-24
[Ready] The Great Lore Renamening

jakeramsay007:
Direct all complaints to the Lore Team

Also happens to remove a few thousand lines worth of virgo fluff items through the power of commenting

PR #454, opened 2018-11-09
[TESTMERGE ONLY] Revert "[Ready] The Great Lore Renamening"

kevinz000:
Reverts Citadel-Station-13/Citadel-Station-13-RP#301

Yeah uh we need to have a talk about this sometime soon when I'm not stressed for time with the whole remove instead of rename/refluff thing.


nik707:

PR #455, opened 2018-11-10
Removes hard mute for ooc during voting

kevinz000:


nik707:

PR #482, opened 2018-11-27
Fixes missing overworld/item icons of HAZMAT engineering suit for sharkbros

NopemanMcHalt:
In not nice way perhaps but... It just works.


nik707:
Can you modularize this just to reduce annoyance?

NopemanMcHalt:
@nik707 there ya go

PR #505, opened 2018-12-10
Allows Vox defibrillation

DerFlammenwerfer:
Makes use of the IS_VOX reagent tag to allow Vox to be defibrillated without removing their NO_SCAN tag. Due to Vox having such a high shock resistance (80% base resistance, according to the Seimens Coefficient of 0.2), this feature is restricted to Defibrillators that can bypass armor, voidsuits, and RIG, such as the Paws of Life and Combat Defibrillator. Anything else will return a message declaring that the voltage rating is insufficient and abort the defib attempt.

### DIFF
--- a/code/game/objects/items/devices/defib.dm
+++ b/code/game/objects/items/devices/defib.dm
@@ -262,10 +262,12 @@
 
 //Checks for various conditions to see if the mob is revivable
 /obj/item/weapon/shockpaddles/proc/can_defib(mob/living/carbon/human/H) //This is checked before doing the defib operation
-	if((H.species.flags & NO_SCAN))
+	if((H.species.flags & NO_SCAN &! H.species.reagent_tag & IS_VOX))
 		return "buzzes, \"Unrecogized physiology. Operation aborted.\""
 	else if(H.isSynthetic() && !use_on_synthetic)
 		return "buzzes, \"Synthetic Body. Operation aborted.\""
+	else if(H.species.reagent_tag & IS_VOX && combat == 0)
+		return "buzzes, \"Voltage rating insufficient. Operation aborted.\""
 	else if(!H.isSynthetic() && use_on_synthetic)
 		return "buzzes, \"Organic Body. Operation aborted.\""
 


kevinz000:
just let all defibs do it

code/game/objects/items/devices/defib.dm
```diff
@@ -262,10 +262,12 @@
 
 //Checks for various conditions to see if the mob is revivable
 /obj/item/weapon/shockpaddles/proc/can_defib(mob/living/carbon/human/H) //This is checked before doing the defib operation
-	if((H.species.flags & NO_SCAN))
+	if((H.species.flags & NO_SCAN &! H.species.reagent_tag & IS_VOX))
```
king5327: This conditional is equivalent to:
`H.species.flags & NO_SCAN & (!h.species.reagent_tag) & IS_VOX`
Since `NO_SCAN` is 4 and `IS_VOX` IS 2, the bitwise and will always return 0. Basically, defibs can't fail this check, period, so they'll always try.
I recommend changing it to `H.species.flags & NO_SCAN && !(H.species.reagent_tag & IS_VOX)` to do what you want it to do. Refer to http://www.byond.com/docs/ref/info.html#/operator for order of operations.

code/game/objects/items/devices/defib.dm
```diff
@@ -262,10 +262,12 @@
...
 		return "buzzes, \"Unrecogized physiology. Operation aborted.\""
 	else if(H.isSynthetic() && !use_on_synthetic)
 		return "buzzes, \"Synthetic Body. Operation aborted.\""
+	else if(H.species.reagent_tag & IS_VOX && combat == 0)
+		return "buzzes, \"Voltage rating insufficient. Operation aborted.\""
```
king5327: The two lines above should be moved **into** the organic check below. Vox IIRC count as organics, so if someone decides to make an FBP vox for whatever reason, you've effectively made it impossible to defib them.

king5327: IIRC, that is, there's no combat mode jumper cables that the RD has in his locker.

code/game/objects/items/devices/defib.dm
DerFlammenwerfer: Ahh okay, thanks for the information. I'll go ahead and make the necessary change

code/game/objects/items/devices/defib.dm
DerFlammenwerfer: This is intended, but can't I just nest an if statement under it to check for isSynthetic() to negate the combat variable requirement, or add a multitool interaction to jumper cables to enable combat mode?

PR #507, opened 2018-12-10
Telescience has been reverted/upgraded back to Aurora's implementation

jakeramsay007:
The bearing/distance system implemented by virgo is fucky as hell. This telescience implementation is tried and tested, although balanced out by the lack of roundstart construction, high power costs and construction costs as well as being logged to prevent abuse by non-antags.

Also adds cargo telepads to make cargo shipping and the likes much easier, although the parts of this system will need mapped in by someone who can actually use the map merge tool


deathride58:
why though

Andromeda-K22:
plsno

with the amount of baddies we get in science someone will use this EVERY. ROUND. to steal the spare ID, or weapons from the armory.

Ragolution:
If you were on main I'd agree, but is the RP server really so bad? Maybe I'm missing the grand espionage.

avabee:
Aurora does not have a telescience calculator publicly available. It doesn't use Baystation's formula.
Are you going to port the bluespace echoes and teleportation conflicts, as well?

avabee:
Aurora's formula change is in `/code/_helpers/game.dm`.
```/proc/projectile_trajectory(var/src_x, var/src_y, var/rotation, var/angle, var/power)

	var/g = 9.81
	var/h = 10
	var/power_x = power * cos(angle)
	var/power_y = power * sin(angle)
	var/time = (power_y + sqrt((power_y*power_y)+(2*g*h)))/g
	var/distance = time * power_x

	var/dest_x = src_x + distance*sin(rotation);
	var/dest_y = src_y + distance*cos(rotation);

	return new /datum/projectile_data(src_x, src_y, time, distance, power_x, power_y, dest_x, dest_y)```
This change needs to be copied over or else the /tg/ calculator will be able to be used. This small change prevents it from being easily calculated.

nik707:
So none of the maintainers are really on board with this. Bring it up again in dev chat if you want another go.

PR #518, opened 2018-12-20
Adds a donator item framework for donator item reskins.

nik707:
Also adds JennaSilver's donator item, a reskin kit for a voidsuit helmet.
Contained is a folder in modular citadel with a file for the items themselves to be created, and another for the reskin kits. This should make it easy for literally anyone to code their own donator item reskin kits in the future. 
Neat.
@LetterJay


nik707:
WHAT THE FUC

PR #573, opened 2019-01-02
Translocators no longer state whether they are contraband or not in their descriptions.

CameronWoof:
Also gives translocators more descriptive examine text, for as long as those files remain not-overridden by some later update.


kevinz000:
After this is merged we're going to talk a bit about translocators and start putting in the "official" stance into the wiki instead of having it on codebase; I don't think administration or anyone else should be forced to codedive to change these things.

PR #575, opened 2019-01-05
Omni-Translators!

Hatterhat:
fuck you, adds translators that understand E V E R Y T H I N G
actually tested, actually works
despacito


Keekenox:
I dig this

PR #577, opened 2019-01-05
[MIRROR] Nerfs cargo pizza

CitadelStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/4625
--------------------
Currently cargo pizza is identical to what the chef can cook. And it has become a huge staple for cargo to order. It's painful for a chef to come later in a shift, only to see everyone eating pizza, or for cargo to order pizza whilst they're on the job. 

This makes pizza from cargo an inferior version to that which the chef can cook. The amount of nutrients is worse, the taste is worse, and it takes more work compared to before, it's barely nutritious, and even dangerous to eat to some before you heat it up. It's still better than what you can get from a vending machine, but not as good as what the chef can make.

Summary of changes:
- supply crates come with frozen pizzas
- frozen pizzas have frost oil, harms prometheans
- heat frozen pizza to get pizza
- pizza looks worse, tastes worse, less nutritious
- new sprites for both frozen pizza, and thawed pizza. Visually different from chef pizza.
- made it cheaper to order pizza, as it is a direct nerf to the value of them.


kevinz000:
Unnecessary.

PR #610, opened 2019-01-14
whats a fuckign balance - SPICY NEW MECH WEAPONS THAT ARE ADMINSPAWN ONLY

Hatterhat:
the Sturmkönig is a machinegun that fires 7.62 (35 brute, +10 compared to the UAC/2's 25 brute) with a 6 round burst.

the Slagger shoots really fast - almost as fast as you can click. still does 40 burn, but it drains more power to assist in cycling the lasers

the Bishop is a marksman's cannon - 2 round burst, but each shell fired is a 14.5mm shell, so...y'know. hit 'em hard, hit 'em fast.


Hatterhat:
i refuse to accept any cries of "BALANCE THIS"

PR #618, opened 2019-01-15
Sickshot buff (last time I'll make a new PR for this)

RMagnum:
Changes brute to burn, makes sense with desc already saying its concentrated sound waves instead of some kinda laser. Raised rng from 20% to 40% because it can only fire 4 times, making it terribly unreliable and a poor substitute to chems with such a low status chance.
(final PR for the same thing, modularized n stuff)


nik707:
fix

PR #620, opened 2019-01-17
taur additions

ParadoxSpace:
my github doesn't have the processing power to find out if this is included in the big-ass polaris sync, so here's a faster port of https://github.com/VOREStation/VOREStation/pull/4574


ParadoxSpace:
daddy

PR #622, opened 2019-01-17
*Peep

RMagnum:
*peep like a bird, sound effect ported from main.


nik707:
So, gonna be real, I'm 100% firmly against more bullshit sound emotes for people to spam. If someone comes along and does an emote cooldown I'm fine with it.

PR #623, opened 2019-01-17
Removes RIGsuit handcuff protection

CameronWoof:
RIGsuits already provide an extreme level of protection for their user--in addition to the resistances provided as a baseline, they also cannot be removed except through surgery by anyone other than the person wearing it. 

In addition to that, they also made you immune to handcuffs. This alters that behavior so that people in RIGsuits can be restrained like an ordinary crewmember, though they still retain their high resistance values and their suit cannot be removed without surgical interference.

### DIFF
--- a/code/game/objects/items/weapons/handcuffs.dm
+++ b/code/game/objects/items/weapons/handcuffs.dm
@@ -70,9 +70,9 @@
 		to_chat(user, "<span class='danger'>\The [H] needs at least two wrists before you can cuff them together!</span>")
 		return 0
 
-	if(istype(H.gloves,/obj/item/clothing/gloves/gauntlets/rig) && !elastic) // Can't cuff someone who's in a deployed hardsuit.
-		to_chat(user, "<span class='danger'>\The [src] won't fit around \the [H.gloves]!</span>")
-		return 0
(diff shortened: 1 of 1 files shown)


ParadoxSpace:
a) have another officer bring you cable/tape
b) neckgrab them and drag them to the brig
c) have a dogborg vore them and spit them out in the brig
d) have you vore them and spit them out in the brig
e) dislocate their legs and drag them to the brig
f) kill them and drag their sleeve to the brig

CameronWoof:
Security officers shouldn't need to carry around rape gear to apprehend people, and people can resist out of neckgrabs and sprint away to their heart's content. Noncon vore is no and if at all possible I'd like to avoid "field execution" as an answer for almost anything security-related.

nik707:
For clarity, Kevinz and I both agreed that this isn't how we want to do this, and instead made cable cuffs no longer work either. RIG suits are mechanically assisted mini-mechs, and they aren't exactly common. They are meant to instill a ridiculous sense of power to the user, and having sec 2 taze/cuff you with no effort completely ruins that as a concept. When someone is using dangerous and powerful work equipment to resist arrest and pose a danger to security and the station, there are ways to handle that using more creative or lethal means. If you're in a position where you've downed someone long enough to handcuff them, you can ask them to remove the gauntlets and if they continue to refuse, you can down them long enough to get them in a critical condition to have the suit removed, and or to have them resleeved. This isn't a common enough problem that I'm ready to start nerfing things over it, but we told Cameron to go ahead and open a PR to see if enough people would care enough to comment make a difference.

PR #627, opened 2019-01-18
Maybe makes medical kits be a slot item maybe

Trilbyspaceclone:
Chang log!
Makes medical kits fit on belts maybe!
Why?
They will not have a belt look on characters and legit hold 7 items like belts do so no powercreep maybe!
They look nice and make logic to me, after all medical kits are normally made in mind to be ease to carry?
Unless im wrong*


Trilbyspaceclone:
God. Damn. It . All

avabee:
Makes sense to me. You could just put it on a belt clip.

PR #633, opened 2019-01-21
Adds the 'Antiseptic Saliva' trait

CameronWoof:
Adds a new neutral trait, Antiseptic Saliva. Characters with this trait can use a verb in the Abilities tab to bandage and salve the wounds of an adjacent character, using nutrition as fuel. The purpose of this trait is to stand up to heal bellies as a way to provide healing to others without needing to use vore.


king5327:
I stand by my request that this verb only heal a single wound instead of an entire limb at once, for a cheaper cost.

CameronWoof:
That would require it to be coded differently and I don't know why you'd want to only heal one of several wounds.

king5327:
Not really. All you have to do is make the loop break once an unsalved and/or unbandaged wound is salved/bandaged. And obviously skip the salved/bandaged ones.

king5327:
High stress scenario where being at max hunger can get you killed is a good example, like on expedition. The last thing you want is a deathclaw charging your ass while you're moving slower than molasses. Being able to choose how many wounds get licked to conserve resources is way better.

PR #648, opened 2019-01-26
Bone clamp surgery kit

RMagnum:
Allows for bone clamps to be put into surgery kits, without actually spawning with one, also adds "singed" to the cautery attack verb cause why not.


RMagnum:
closed and reopened in the hopes travis dosnt screw me over again :<

PR #657, opened 2019-01-28
Nine Tailed Kitsune Sprite

Knouli:
Was planning on donating to keep them private, buuuuuuuut, eh.

Adds a public use, recolorable multi-part nine tailed kitsune sprite as based off of existing sprites from the VORE code ckey restricted content.

This is my first attempt at something like this, I'm certain I've messed something up somewhere.


Knouli:
I did something wrong on the git side, going to correct and reopen with a new one. Sorry.

PR #658, opened 2019-01-28
Nine Tailed Kitsune Sprite

Knouli:
Second time's the charm? Think I figured out where I flubbed up in git. . .

Was originally thinking of donating for this, but decided on a public release. 

Adds a recolorable, multi-layer nine-tailed kistune sprite based off of a pre-existing sprite in the VORE ckey restricted content.


Andromeda-K22:
unit tests runtimed, @kevinz000 rerun

PR #664, opened 2019-01-30
Maybe makes the surv take 6 mins to ready

Trilbyspaceclone:
3 to 6 mins before the game starts
Why?
Game is starting far before its ready


RMagnum:
this was caused by my older PR that increased map init time by 200 seconds as far as I know, which is reverted till I fix it.

kevinz000:
not done the right way

PR #668, opened 2019-02-01
Adds ointment to fire lockers

Trilbyspaceclone:
What do this do?!?
Adds one stack of /obj/item/stack/medical/ointment to each fire locker
Why did ya do dis?!?!
Its a fire locker and in the case of massive fire or hear people get burned. Kinda like how some O2 lockers have O2 kits


LeonDuvall:
would also love to see this on main

PR #679, opened 2019-02-06
Properly-ish prevents an uninitialized game start via ugly sanity test

RobotAlice:
I added a var to the game ticker, block_start. It starts TRUE and is used to conditionally enter an infinite while-sleep loop predicated on if we're still blocking the game from starting or not. It is later set FALSE in master.dm when initializations complete.


RobotAlice:
Known caveat: this breaks the game when you intentionally start the game with the admin verb before initialization completes, and requires either manually changing the game ticker's block_start via view vars, or restarting the server.

PR #698, opened 2019-02-17
Buffs the shit out of tape

Trilbyspaceclone:
Changling
Makes tape last 3x longer
Why
God their still so useless

### DIFF
--- a/code/game/objects/items/devices/taperecorder.dm
+++ b/code/game/objects/items/devices/taperecorder.dm
@@ -363,14 +363,14 @@
 
 /obj/item/device/tape
 	name = "tape"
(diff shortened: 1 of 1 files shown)


CameronWoof:
Ten times three isn't ninety. This would make it record thirty minutes, right?

Trilbyspaceclone:
@CameronWoof 1800 = 30 mins, 
likely was buffed in the past and never updated, as someone that used a tape recoreder form shift start to end with a 1 hour + had only 12 tapes
12 x 10 is 120 hours or 2
When it lasted for 4 hours~
As well as time stamps lasting in 30~ min time staps

CameronWoof:
Gotchu, sorry for the confusion.

PR #707, opened 2019-02-19
Try no. 2 now less memey but still removes dumbass nerfs

Armyguy:
Most of the nerfs were done do to one guy flying out by themself to V2/4 a guy who no longer plays, So lets undo some of that.

It still has its slow down and that you can't see shit, But it drops the - 20 emp protection (he was a robot see - vs emp) and added armor for melee and bullet to get it the same as the sec rig.


granodd:
This puts the stats on par with a combat rig while also being the only rigsuit on station like that. Not to mention removing any form of counterbalance while being Unathi exclusive.

No.

PR #709, opened 2019-02-19
Moves the Unathi Breacher Suit to the Security EVA section

Armyguy:
Yea it just moves that suit to a bigger Sec EVA
https://cdn.discordapp.com/attachments/512864505030639620/547425743639150607/See3.jpg

### DIFF
--- a/maps/tether/tether-06-station2.dmm
+++ b/maps/tether/tether-06-station2.dmm
@@ -33861,9 +33861,9 @@ nV
 oF
 oY
 pE
-qj
+qc
 oY
-rN
+qc
 qc
 sU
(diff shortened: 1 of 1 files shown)


kevinz000:
This is being closed at the 72 hour mark unless you can come up with a reason security needs this in its EVA, you've already made two meme PRs for the same thing with what looks to me like unironically wanting to buff something you tend to abuse, and my patience has worn thin.

kevinz000:
I've determined the supplied reason to be insufficient to move the suit.
It's a gimmick item, would rather have it stay where it is in the current status quo.

PR #713, opened 2019-02-21
Rebalances MediHounds and Standard Medical Cyborgs

CameronWoof:
- Grants all cyborg hyposprays access to bicaridine and kelotane
- Reduces the chemicals available to MediHound sleepers to be in line with standard cyborg hyposprays (removes imidalozine and alkysine)
- Reduces the cost of sleeper injections to be in line with standard cyborg hyposprays
- Removes the water cost for medihound hypospray injections
- Adds advanced brute and burn kits to medihounds

This should bring MediHounds and their standard counterpart closer to parity. I would have done this a long time ago, had I realized the conversation upstream about doing exactly this never went anywhere.


cruxxansata:
This is the kind of thing I've been waiting to see. Dogborgs have always had some kind of special attention paid to their abilities, putting them above standard modules in almost every capacity. It's good to see someone actually biting the bullet and trying to nerf them despite how everyone might react. If I can expect to see more PRs like this in the future, then I'm looking forward to it.

PR #717, opened 2019-02-22
Lowers time you need to be ghost maybe to 5 mins before you can be a drone.

Trilbyspaceclone:
Changlings
Lowers time you need to be ghost maybe to 5 mins before you can be a drone.
Why
Requested by Vail Seria/BV-852i 'Gnut', so people can play the game or such to grater fun


deathride58:
wait why isnt this a config option

PR #722, opened 2019-02-22
Remaps the Bar in a new style

Andromeda-K22:
+ Bar remap
+ Bartender gets a condimaster now.


Andromeda-K22:
https://camo.githubusercontent.com/3afc30bb56bf5de1d60e723a04d704f649b101f9/687474703a2f2f6d617064696666626f742e64657874726173706163652e6e65742f46696c65732f3133323337343330382f36373431323338332f302f61667465722e706e67

Tename:
I think the tile design would be better in the kitchen area along with the booths and the wood design + tables in the bar.

Tename:
And the arcade in the bar instead.

Andromeda-K22:
Aaah shit i forgot to merge.

Andromeda-K22:
@Tename can you be more specific about what you mean with the tile design? That sounds like literal gibberish.

Tename:
@andromeda-K22 The bar floors are tiles...

Andromeda-K22:
image

CameronWoof:
Not super thrilled with this. The black-and-blue tiling looks really out of place, and this clutters up the space so much. The old design allowed for grabbing food from the kitchen and then taking a table while chatting with patrons over at the bar. The dividing wall seems like it's there to make people leave the dining room if they want to socialize.

I don't like how this looks or what it does. I might be more open to it if you removed the dividing wall and made the dining room a contiguous space again, just with one section of tile flooring and one section of wood, but all in all, I think our current setup is significantly better than this suggestion.

Andromeda-K22:
image

Skycave1:
Personally not much a fan of the futuristic style. Maybe I'm just a sack for rustic noir style bars, and not the futuristic metalic styling

king5327:
One barsign needs to go. The two diner tables with four seats need to be replaced with two wall-facing tables with three seats each. Finally, drop the linoleum or drop the futuristic tiles - they don't work well together.

Oh, and remove the doors please.

nik707:
I agree with removing one bar sign, and the dividing wall. The north wall just splits one good social area into two, and I'm not much a fan of it. I think remove the dividing wall and let the floor tiles signify the split of control between chef and bartender. I also agree to decide on a different pair of tile-types, the metal and linoleum clash. The tables however are fine as is, IMO, though remove the double windows. The ones that encroach on the door look out of place.

Andromeda-K22:
The double windows have already been removed.

king5327:
If the dividing walls are removed, then the tables should be extended one tile north each maybe? If the windows get tinted (and moved to the outer tile rather than the inner tile) then it could make for some nice simple privacy stuff.

ForFoxSake:
If you could throw some emergency shutters behind the doors/window separating the canteen and the bar that'd be ideal.
Having actually seen this bar in game over screenshots, I have to say it feels a lot more comfortable to be roleplaying in than the previous design, so I'm looking forward to this being merged proper.

king5327:
Also, the whole top part needs to be moved a tile down, there are tiles in the bar area that the bartender can't see.

Andromeda-K22:
https://cdn.discordapp.com/attachments/323935657821077524/551933075857342474/unknown.png

PR #738, opened 2019-02-24
[DNM] Maybe makes the Rnd console use uppercase letters 

Trilbyspaceclone:
Changlings
Makes Rnd console maybe display Steel as Steel, Osmium as Osmium, and Plastic as Plastic 
Hos is a Ling?
Do gold/diams/green/sliver/glass being all uppercase and the three up their not being upper case
Rather annoying to me


nik707:
oh wait this is DNM 😂

PR #752, opened 2019-02-26
what the hells a CELLRATE

Hatterhat:
removes the CELLRATE multiplier that fucked cell chargers, also fucks the power economy


Hatterhat:
kevinz says no

PR #753, opened 2019-02-26
Makes GPS and real bluespace crystals printable 

Trilbyspaceclone:
Changling
Uncommits out GPSs and adds in blue space crystals
Makes the real bluespace crystals take 3 times more mats then fakes
Makes Adv GPSs Tiny well basic GPS small meaning theirs a reason to upgrade
Hos is a ling
GPSs are useful for Engi or mining, as well as exploitation teams to be smaller for space!
Bluespace real crystals do affect Tele-coms and other types of sci things, making more uses and better affect people safer.


nik707:
but why

kevinz000:
leaving thoughts on this later

PR #773, opened 2019-03-02
Makes Dart guns RND printable 

Trilbyspaceclone:
Adds in dart gun and its carts into Rnd after hand teles


kevinz000:
yeah I'll need to think about this I don't think this is the right way to go for R&D.

PR #791, opened 2019-03-05
Re-add Humvee and Science Rover

DarkWolfInsanity:
Cars for driving in the dangerous hallways and outside world. Expensive so they're not bought every round or at bulk, but cheap enough that a good mining shift could get them.


nik707:
I'm gonna tentatively merge this, but it will hit the FATTEST revert if people are autistic about it.

PR #795, opened 2019-03-06
if this doesnt work let's blame nik

Hatterhat:
@nik707


ForFoxSake:
Is there a reason you've only commented out the warning, and not the entire if block?

PR #801, opened 2019-03-07
[I need help] New tram

Tename:
Shit I can't fix

- Magical holes appearing when the tram moves.
- People get shocked when spawning at the tram station.
- Movement is fucked for some reason.

Stuff it does
- Makes the tram a long boy multi-car monster with a security and tiny medical area.
- Makes CC a bit bigger and fixes some random issues.
- Tram tracks extend outside

https://i.gyazo.com/72d192ba33a99395dc34bfbab05bc5c8.png

https://i.gyazo.com/6eaf91cf23c3301a64c4050f738bb8a6.png

https://i.gyazo.com/3fb630529ea255913e9f29d5252bf7fc.png


CameronWoof:
This looks real damn cool so far!

Tename:
Just remembered back windows are fucked going to have to fix them when I get back.

ForFoxSake:
I was excited to see the prospect of a new tram, as I often find the current tram is a little too crowded to be able to actually get to a seat in a reasonable amount of time.
Though, having counted the number of seats, I'm pretty disappointed that this tram design is twice the size and adds only four more seats. (This design having 34, and the old tram having 30.)

All in all, I like the idea of there being a new tram, but I'm concerned that this particular design may leave people getting trampled during busy rounds, as a lot of people try to rush into the narrow, hard bordered corridors in an attempt to find a seat in time.

Tename:
I can probably move some things around and fit in some more seats, perhaps adding that tableless area the current tram has. The doors however i am not sure if I can do anything about while maintaining the additional areas inside the tram, but I feel with the two doors being so far apart and double in size it should work out.

Tename:
But also consider security staff and any prisoners will probably be in their own area wo that should increase the amount of spare seats for the public

ForFoxSake:
Including the security seats that's still only like 8 seats, half of which aren't even public.
This were my design, I'd likely replace the central doors/walls that separate the central gangway, with single panes of glass, as this will give you room for an additional 8 seats in itself.
tram_section_1
If you then replace the table pairs at the far front and rear of the tram with seats, you'll get an additional 8 seats again.
With these two changes, you get 44 seats on the first carriage alone, and, imo would look more like a real train carriage too.

That's my best suggestion for this design, and I still feel like there would be a few rounds where we'd see every seat on the tram taken.

Tename:
I don't know I just feel like it would look lower quality with a single pane of glass, I am going to keep expirmenting but if it really comes down to it I'll replace it.

Tename:
It's been 15 days I still can't figure out why exactly this isn't functioning properly, any of the fixes I have tried failed. I am going to continue on it more but I am just going to make a new PR.

DM me on discord if you have any ideas.

PR #812, opened 2019-03-10
Makes the respawn timer not make me wanna commit oof()

LiquidFirefly:
Respawn timer is unreasonably long because Vir devs make me want to scream.  This changes it to fifteen seconds, that way it still can't be spammed with any sort of speed, but it's still convenient enough that anyone who's legitimately checking out a round isn't inconvenienced.  Suggested the change to Kevinz, he asked me to make a pull. 

Edit: Upped the time from fifteen seconds to a minute, a time that isn't quite as quick, but still isn't a speedbump in the slightest for people that are using respawn for legitimate reasons.

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -373,8 +373,8 @@
 		var/deathtimeseconds = round((deathtime - deathtimeminutes * 600) / 10,1)
 		to_chat(usr, "You have been dead for[pluralcheck] [deathtimeseconds] seconds.")
 
-		if ((deathtime < (5 * 600)) && (ticker && ticker.current_state > GAME_STATE_PREGAME))
-			to_chat(usr, "You must wait 5 minutes to respawn!")
+		if ((deathtime < (150)) && (ticker && ticker.current_state > GAME_STATE_PREGAME))
+			to_chat(usr, "You must wait 15 seconds to respawn!")
 			return
 		else
 			to_chat(usr, "You can respawn now, enjoy your new life!")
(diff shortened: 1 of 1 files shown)


CameronWoof:
This seems unnecessary and exploitable. I don't want to encourage "round checking" as a common practice.

granodd:
@CameronWoof it already is a thing. 5 minutes is still a fairly short amount of time versus places like Baystation where you have to wait 30 to respawn. This is also a much more casual server than Baystation is which runs regular antag rotations, where as ours is more of a "when the admins feel like mucking with things" deal.

I would much rather have the respawn timer be 1 minute like it used to be, though there isn't much difference I suppose with this sort of thing due to how short it is.

avabee:
Why is "round checking" a bad thing, exactly? Should people not know if something's going on that they don't want to deal with? The alternative is joining, wasting your time when you eventually find out in-character, and then leaving the round having wasted however long it took to accomplish that plus however long the respawn timer is. Would you prefer that?

LiquidFirefly:
It's not a required change by any means, it's far more geared towards QoL for the people, like me, that often have situations that they don't want to jump into, people I don't want to interact with, characters that I want to interact with specific people, and manifests that I don't want to put my name on if certain people are sequestered away.  

Jobhopping happens anyway, checking out a round happens anyway.  It'll continue to happen as long as respawn is enabled, and both of those things are if anything beneficial in a server with large segments of lowpop or high RP focus. 

There's not much space for this to be abused either.  Like, what are you gonna do with it, spam general channels, make bodies, or spawn job items?  All shit you can do with much higher efficiency, and without alerting the whole server if you just grief normally or play the game?   There's absolutely no reason to have the timer as high as five minutes in my eyes, and there's not much space for abuse.  I don't want to have to tab out and wait five minutes every time I check out a round, it's obnoxious.

PR #823, opened 2019-03-13
[Ready] Adds new ID sprites and Secretary Headset gets Service Comms

king5327:
I kinda forgot I was doing the service comms thing when I started the ID stuff, so guess both get bundled into a single PR.

This is testmerge until I'm sure the syndicate fake ID works, but otherwise everything's in order.

Not 100% up to date but here's an image: Cards
Posted because if you look at the actual icon file some of the IDs are split into parts.

Credit goes to @CameronWoof for the sprite design. Which I proceeded to take a hammer to.


king5327:
The PR is entirely functional, and syndicate IDs now properly take the shape of the source cards. The only bug left is if an admin changes the type of sprite_stack to something that's not a list updating the icon makes it invisible but in that case reset_icon works.

Quick guide to bussing in custom IDs: open card_cit.dmi and figure out what pieces you need. Go to the vars of the relevant card and look for sprite_stack. Add, from bottom layer to top, the pieces. In this layout, I recommend base->top->stamp->pips (with pips being optional and stamp in most cases matching top, but in rare cases can match pips instead). Call update_icon and the ID will update to its new appearance. If you fuck up, call reset_icon and everything should return to normal, assuming only sprite_stack was touched.

king5327:
Not major but someone mentioned holographic clothing not working right when set to permits. I forgot to investigate that so I'll get it tonight.

PR #826, opened 2019-03-14
Main SMES changes

nik707:
The main SMES is now a subtype, rather than a mapmaker var-edited normal one. This also means that it now sends a stationwide alert every 3 minutes when between 30 and 20% to alert the station to set up solars, and when below 20% to alert the station that non-Engineering personnel are now permitted to attempt an engine startup. 
It **shouldn't** output the message when the SMES is actively charging, so it doesn't spam the message when you charge it after a discharge for upgrades. The unfortunate thing is, it will paste the message every 3 minutes during controlled discharges for upgrades but, this is hardly spam.
Also makes the main SMES a SMES subtype, so this can be put on any map at any time just by using the /main SMES subtype. Cool. Engine SMES is also now a subtype, but with no new added features. Just ease of mapping.

### DIFF
--- a/code/modules/power/smes.dm
+++ b/code/modules/power/smes.dm
@@ -50,6 +50,7 @@
 	var/lastenginealert = 0
 	var/lastcharge = 2e+007
 	var/lastcheck = 0
+	var/percentfull
 
 /obj/machinery/power/smes/drain_power(var/drain_check, var/surge, var/amount = 0)
 
@@ -379,6 +380,10 @@
 				output_level = (input(usr, "Enter new output level (0-[output_level_max/1000] kW)", "SMES Output Power Control", output_level/1000) as num) * 1000
 		output_level = max(0, min(output_level_max, output_level))	// clamp to range
 
+	else if( href_list["mute"] )
+		lastsolaralert = world.time + 6000
+		lastenginealert = world.time + 6000
+
 	investigate_log("input/output; <font color='[input_level>output_level?"green":"red"][input_level]/[output_level]</font> | Output-mode: [output_attempt?"<font color='green'>on</font>":"<font color='red'>off</font>"] | Input-mode: [input_attempt?"<font color='green'>auto</font>":"<font color='red'>off</font>"] by [usr.key]","singulo")
 	log_game("SMES([x],[y],[z]) [key_name(usr)] changed settings: I:[input_level]([input_attempt]), O:[output_level]([output_attempt])")
 	return 1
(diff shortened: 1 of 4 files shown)


Keekenox:
I personally am against this thing. I think it's okay to have the SMES warnings for low charge and everything, but I am against the thing that tells people it's OK to start the engine regardless of current job or character. I think that's actually specifically why things like maint drones exist. On Virgo and Polaris (directly related codebases) that's pretty much been one of their main uses, you know, starting the engine and stuff like that when there's no engineers around, since you can just ghost into a drone at any time and then leave the round with no expectations to interact with any of the crew.

nik707:
Which would be super cool if drones/ghosts were at all common enough for us. But they aren't.

Keekenox:
Probably because people hardly know they exist because they aren't commonly used. Here's an idea. When SMES charge gets too low, send out a pop up for ghosts asking for a drone shell with a prompt saying that they should set up the engine or something. Alternatively just rely on people to send out faxes about the situation and have admins try and get people from deadchat. ANOTHER alternative is doing the discord ping thing that virgo did (that worked surprisingly well). 

Virgo had a role opt in thing on the discord where you could opt in to departmental roles. There is an IC channel, I can show you an example. This system was actually generally regarded on virgo as a fantastic idea (especially since it's opt in) and worked really well.

[https://i.imgur.com/99Ldy7Z.png](url)

nik707:
ITT person complaining about too many channels wants another one. This idea is neat and all, but the setup is needed, and an admin has to be around to deal with it. This is a 100% hands-off solution that no admin needs to deal with. It solves a large number of problems in a much easier way than having admins need to babysit and ping roles for things that, lets be honest, most people will probably ignore. Sure, we can do it too sometime, but unless someone wants to take the time to set it all up again, this is my solution.

king5327:
As long as there's a button on the SMES to disable the speaker.

Because otherwise people are just going to start deconstructing and reconstructing the SMES when they start clearing it for upgrades so that it doesn't shriek every 5th tick when coil or turbine power drops to 0 for like one cycle.

nik707:
> As long as there's a button on the SMES to disable the speaker.
> 
> Because otherwise people are just going to start deconstructing and reconstructing the SMES when they start clearing it for upgrades so that it doesn't shriek every 5th tick when coil or turbine power drops to 0 for like one cycle.

It's on a 3 minute timer, and turning it off completely ruins the purpose. Someone joins, turns it off, leaves, power goes fucky and drains, no one knows until it's too late. So, probably not going to happen.

PR #827, opened 2019-03-14
Click pens (oh my god)

peeesh:
little clicky pens ok if people spam it i'll uhhhh remove the sound and add a cooldown if i can figure it out


LiquidFirefly:
if people spam it, People spam it IRL

PR #835, opened 2019-03-16
Voidsuit QoL fixes/additions

peeesh:
"well, fucking hell peesh, whats new?" i hear you saying

- voidsuits can now be modified (helmet/boot/tank/cooler attachment, removal) in hand.
no, it doesnt work when actually worn on the suit slot. i checked that dont worry

- voidsuit helmet deploy sound
ported from tg, that FUCKING NOISE is the stuff of my wet dreams, so its here too

- action button to toggle helmet
spawned from the noise, i realised deploying helmets was a chore. instead of it being a verb in the object tab/context menu, its now a button. i commented out the verb code, in case theres a suddenly better way to do it.

wait, peesh. not even kev really knows those buttons?
well yeah

when used inhand, the suit now deploys the helmet. if its not equipped into the suit slot, it does nothing. i tested this, with full + empty hands. its fine.

then why did you do that?

the button calls to use the suit inhand. so when it does, the helmet deploys

i fucking despise this code and this is the easiest method i saw, so sorry if it fucks up somewhere else


peeesh:
one day we're gonna go full circle and just have straight up /tg/ hardsuits

ForFoxSake:
Why'd you comment out the verb altogether? Is there a problem with having both?
I'm probably the only person to do this but I type verbs out for almost everything.
I know that XKCD comic is gunna get linked, but, it's like muscle memory for me to type toggle-helm at this point, and I think the option would be kinda nice.

peeesh:
oh right people dont use hotkeys sometimes, sorry. ill uncomment it

PR #842, opened 2019-03-17
Makes localhost a thing.

GrayRachnid:


king5327:
I know this is already a feature in TG. I'm reasonably sure vorecode also has it.

GrayRachnid:
Vore code does not have it before this pr. I know because I just suffered from it.

ForFoxSake:
Looks like a great feature that will make my life a lot easier in testing.

Though, would it be possible to make it a config option that defaults to on?
I've seen localhost IP spoofing used as an attack vector on programs before, and this seems like it could be a big back door on a live server.
Probably not something the average script kiddy could get their hands on, but I feel it's worth keeping it in mind.

PR #853, opened 2019-03-20
Adds Junior Scientist, Junior Engineer, Medical Resident

rpilcrow:
these should provide a good intermediary between interning and working the full role, while still broadcasting 'i'm learning, teach me!'

todo: potentially better title than 'junior scientist'? just waiting on nicc to ask a friend about it.
update: only 'good' alternative was 'associate scientist', which sounds a bit too ambiguous. junior scientist will do.

## Changelog
:cl:
add: 'Junior' alt titles for Science, Engineering, and Medical; Junior Scientist, Junior Engineer, and Medical Resident.
/:cl:


avabee:
Honestly, I could see changing "scientist" to something like Senior Researcher and then the intermediary role to just Researcher?

nik707:
I'd rather have a role to say you don't know things, than one that says you do know things.

PR #857, opened 2019-03-22
Adds more radio channels to IAA headsets

DigitalSquirrel95:
Felt like IAA should have more access to comms in order to do their job more effectively, because as it is now, they need someone else to tell them if there's an issue in a department that their headset doesn't have access to.

### DIFF
--- a/code/game/objects/items/devices/radio/headset.dm
+++ b/code/game/objects/items/devices/radio/headset.dm
@@ -307,7 +307,7 @@
 	name = "internal affair's headset"
 	desc = "The headset of your worst enemy."
 	icon_state = "com_headset"
-	ks2type = /obj/item/device/encryptionkey/heads/hos
+	ks2type = /obj/item/device/encryptionkey/heads/captain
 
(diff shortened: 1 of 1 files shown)


Andromeda-K22:
-1
IAA has no real power and should have no access on captain channels. 
They're arbitrators to company policy; they shouldn't worry over everything. 
This is also a pretty massive buff - so unless IAA becomes on the level of heads in terms of QC and power, this shouldn't happen.

CameronWoof:
I could see IAA becoming kinda supercoppy with a buff like this. Listening in on every department, pointing out breaches of SOP, etc. Not sure they really need this.

nik707:
IAA already fit under the same requirements of play as Heads. This was detailed in the announcement about the upping of standards. I'm for it, but since people are unsure @kevinz000 can weigh in.

kevinz000:
fenced tbh
i'm not actually a fan of too much "law and order" so there's that :v)

kevinz000:
sorry but i don't think this is necessary/going to be ultimately a good move, maybe later

PR #871, opened 2019-03-26
Security oties no longer chase crew set to arrest

CameronWoof:
This should make them significantly less dangerous to have around.

### DIFF
--- a/code/modules/mob/living/simple_animal/vore/otie.dm
+++ b/code/modules/mob/living/simple_animal/vore/otie.dm
@@ -172,7 +172,7 @@
 	vore_pounce_chance = 60 // Good boys don't do too much police brutality.
 
 	var/check_records = 0 // If true, arrests people without a record.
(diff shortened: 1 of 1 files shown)


Tename:
It kind of defeats the point of buying a security version over the standard.

CameronWoof:
If someone wants to overhaul them to not maul people to near-death as punishment for not having vore enabled when set to arrest, that'd be better. I don't know how to do that and this solves that problem without removing them entirely, so people can still buy them as stylish pets to drag around.

Tename:
> If someone wants to overhaul them to not maul people to near-death as punishment for not having vore enabled when set to arrest, that'd be better. I don't know how to do that and this solves that problem without removing them entirely, so people can still buy them as stylish pets to drag around.

I mean, it isn't really supposed to be safe, it's like many other aspects of security and their tools, it comes with risks. And I mean is it really that big of an issue, they are easy to outrun, not too smart when it comes to z-levels, often has a handler nearby, and security rarely even sets status.

PR #957, opened 2019-03-28
[MIRROR] [MIRROR] [READY]Ports /tg/station pixel projectiles, processing subsystems, timer subsystems, and some misc stuff to make it all work

CitadelStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/4719
--------------------
Original PR: https://github.com/PolarisSS13/Polaris/pull/5636
--------------------
What it says on the tin.
For those of you unfamilier: Pixel projectiles on /tg/ gives projectiles true angular motion, and while you may say that's already a thing with the current system it doesn't take into account where you clicked on a tile, and only takes into account which tile you clicked so I'm porting the tg one. Also, the new projectiles system is processed on a projectile subsystem, and is 100% accurate and more realistic! Yeah.
Anyways does this place have an auto changelogger or do I have to manually put changelog files? :/


nik707:
From before hardsync began

PR #962, opened 2019-03-28
[MIRROR] [MIRROR] Adds Alien Reality pods, aka AR pods, for use in events or PoIs.

CitadelStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/4727
--------------------
Original PR: https://github.com/PolarisSS13/Polaris/pull/5798
--------------------
  **Tweaks:**
Organs can now add or remove verbs.
Organs now can have special effects on processing.
Xenomorph plasma vessels now regenerate phoron passively, at an incredibly slow rate. Can be accelerated by consuming liquid phoron.
Virtual Reality pods cleaned up.

 **Additions:**
Xenomorph organs now give their respective abilities when implanted.
Alien Realiy pod added for PoIs or admins to use. Functions similarly to a VR pod, except it will produce a living body on the pod's turf.
Three replicant species added for event and PoI use. They are shape-shifters each with different abilities due to their organs.
Replicant organs added, versions of many organs that do not reject in their host. Two currently exist that give special passives.


nik707:
From before hardsync began

PR #969, opened 2019-03-28
[MIRROR] [MIRROR] Introduces caseless ammo, and a caseless rifle

CitadelStationBot:
Original PR: https://github.com/VOREStation/VOREStation/pull/4750
--------------------
Original PR: https://github.com/PolarisSS13/Polaris/pull/5886
--------------------
Originally a part of https://github.com/PolarisSS13/Polaris/pull/5837

So let me be very clear about one thing.
## This gun isn't balanced yet.
I need feedback to decide how it will be balanced. Right now, this is just a proof of concept.

The goal here is to make it possible for certain ammunition types to be caseless. I had done this previously in a very old version of Baycode like five years ago, but it was tied to the gun rather than the ammo. In this current version, you could theoretically change _any_ ammo to have a caseless version, and it will work in any gun that chambers it.

The caseless ammo rifle here is intended to be something about as powerful as a pistol caliber, but with a large 30 round magazine, and maybe options for specialized ammo such as stun ammo or EMP ammo in the future. R&D will be able to build this gun.

But like I said, balancing it is an issue, so I need feedback on what materials should be required to make the gun (phoron obviously), and what sort of damage output this gun should have.

### Unrelated but needed changes
- Deleted several unused icons out of gun.dmi


nik707:
From before hardsync began

PR #1040, opened 2019-03-29
Fixes S&R loadouts to allow them to use medical stuff

RMagnum:
I dunno why this was never done but its done now. Tell me if I missed something on accident. or if I fucked something. I did this in like an hour while not paying attention and falling asleep, but it worked fine in testing.


kevinz000:
merging at 25 hours

ParadoxSpace:
Can you give them literally any NIFsoft access?

RMagnum:
> Can you give them literally any NIFsoft access?

There is already 9 NIFsoft disks in the room below chemistry which only require access to the door, which they have. I'll see about giving them NIF access, but it's never something I've touched before.

PR #1046, opened 2019-03-30
Revert "[MIRROR] [MIRROR] Prevents people from joining with the same character"

avabee:
Reverts Citadel-Station-13/Citadel-Station-13-RP#1002

This should not have been merged without discussion.

### DIFF
--- a/code/modules/mob/dead/observer/free_vr.dm
+++ b/code/modules/mob/dead/observer/free_vr.dm
@@ -11,7 +11,7 @@ var/global/list/prevent_respawns = list()
 	set category = "OOC"
 	set desc = "Free your job slot, remove yourself from the manifest, and prevent respawning as this character for this round."
 
-	var/confirm = alert("This will free up your job slot, remove you from the manifest, and allow you to respawn as this character. You can rejoin as another \
+	var/confirm = alert("This will free up your job slot, remove you from the manifest, and prevent you from respawning with this character for the round. You can rejoin as another \
 	character if you like. Do this now?","Quit This Round","Quit Round","Cancel")
 	if(confirm != "Quit Round")
 		return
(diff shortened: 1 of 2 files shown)


code/modules/mob/dead/observer/free_vr.dm
```diff
@@ -11,7 +11,7 @@ var/global/list/prevent_respawns = list()
 	set category = "OOC"
 	set desc = "Free your job slot, remove yourself from the manifest, and prevent respawning as this character for this round."
 
-	var/confirm = alert("This will free up your job slot, remove you from the manifest, and allow you to respawn as this character. You can rejoin as another \
+	var/confirm = alert("This will free up your job slot, remove you from the manifest, and prevent you from respawning with this character for the round. You can rejoin as another \
```
deathride58: this is very inaccurate

code/modules/mob/new_player/new_player.dm
```diff
@@ -169,13 +169,6 @@
 		ViewManifest()
 
 	if(href_list["SelectedJob"])
-	
-		//Prevents people rejoining as same character.
-		for (var/mob/living/carbon/human/C in mob_list)
-			var/char_name = client.prefs.real_name
-			if(char_name == C.real_name)
-				usr << "<span class='notice'>There is a character that already exists with the same name - <b>[C.real_name]</b>, please join with a different one, or use Quit the Round with the previous character.</span>" //VOREStation Edit
-				return
```
deathride58: comment this out instead and add //CITADEL CHANGE - allows people to respawn as the same character

kevinz000:
make the pr again if you still think this should be reverted but the previous pr doesn't prevent respawning if you cryo/tram and it's just named in a really shitty way and described in a shitty way that doesn't actually say what it did

PR #1056, opened 2019-03-30
Adds an option to fix conveyor movement direction

Auris456852:
## About The Pull Request

For one reason or another, conveyors, when constructed by players, have south and north set as their forwards and backwards directions, despite the conveyor's actual direction. I couldn't find a way to just make it fix on spawning, but, I did the next best thing.
This only adds a wrench option on conveyors. If you use a wrench on a conveyor that isn't operating and has its panel open, it runs the direction-setting part again, fixing the forwards and backwards directions.

## Why It's Good For The Game

Simply put, actually being able to create conveyors. I do hope, however, that people don't actually proceed to use this for the old banana peel conveyor circle. I also hope, that it isn't shortly used to crash the server. You people are better than that, aren't you?

## Changelog
:cl:
add: You can now use a wrench on a conveyor with an open panel to fix the direction it's supposed to move objects in.
/:cl:

### DIFF
--- a/code/modules/recycling/conveyor2.dm
+++ b/code/modules/recycling/conveyor2.dm
@@ -108,6 +108,21 @@
 					C.conveyors |= src
 			return
 
+	if(istype(I, /obj/item/weapon/tool/wrench))
+		if(panel_open)
+			if(operating != 0)
+				to_chat(user, "<span class='notice'>Turn the conveyor off first!</span>")
+				return
+			else if(dir & (dir-1)) // Diagonal. Forwards is *away* from dir, curving to the right.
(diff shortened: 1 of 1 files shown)


kevinz000:
don't you need to update_icon/make sure it's visually right?

Auris456852:
It doesn't change the dir of the actual conveyor any. The direction of the conveyor is set already and so is the icon on construction of the conveyor. Just that the forwards and backwards variables, those being, the direction it's supposed to move things, don't set properly. So you get horizontal conveyors that move things vertically.
This possibly slightly crude fix should rectify that.

PR #1065, opened 2019-03-31
Project Hardsync

kevinz000:
Now with a new PR! And renewed effort!

All that's left is ballistics/gun rewrite ~~that I've spent way too long on~~ but we'll make it!


Hatterhat:
i believe

PR #1073, opened 2019-04-01
Storage Rebalancing

kevinz000:
## About The Pull Request

Normalizes boxes at 7 small items/14 tiny items rather than 4 small items/8 tiny items
This means chameleon boxes are nerfed.
Increases storage reach by 1.
Quick equipping via E now attempts to put it into any storage you have if it can't equip to a slot because why not?

Decreases minimum game click detection to 0.1 from 0.2 seconds, as a bugfix. Comments said that's how it should be, let's give it a try!

## Why It's Good For The Game

Makes zero sense that boxes have different storage amount because they're different colors and even if someone goes "but they have better quality material" - No. R&D shouldn't be able to print huge boxes that can somehow cram more mass into the same space, and normies shouldn't be punished just because they didn't roll a job with colorful rather than grey cardboard.

Storage reach is powercreep QoL because I think people should be able to reach further into their backpacks and don't think this is too much of an issue.
Quick equip change is mostly to make equipping stuff easier but is less powercreepy-but-still-powercreepy QoL that I think is good.

## Changelog
:cl:
balance: All boxes now have 7 slots regardless of what it is
balance: Storage reach increased by 1
experimental: Quick equip now seeks out storage to quick equip into rather than only looking for clothing slots.
/:cl:


Zna12:
fucking speedmerge this shiiiiiiiiiiiit.

kevinz000:
@nik707 done with some small issues that i can't correct but shouldn't be an issue atm
@blackmajor i'll do the stuff later for clicking behind stuff when i refactor and overhaul storage interface, which is going to be stuffed in after hardsync now that hardsync is almost done

PR #1082, opened 2019-04-02
Revert "Revert "Snow planet addition"", now we're getting somewhere

RMagnum:
Reverts Citadel-Station-13/Citadel-Station-13-RP#1080

Adds a new Snow planet, temporarily named V6/Snowfields.

-Features cold atmosphere that requires a snowsuit, boots, and gloves doesn't require internals to survive, but it'll stop the .5 burn damage.

-Voidsuits are nerfed here, because of the slowness of walking on snow with them stacks, meaning snowsuits or explorer suits are more efficient. (I plan on adding new snowsuits made specifically for this at some point.)

-Re-balances diyaabs and other sif specific animals to better fit the nature of v6

-Ports some unused snow POIs, along with a few new ones, all of which were edited to contain loot. (I plan on adding more, since there are only 6 active, finished ones.)

-Changes the shuttle refuel time from 45 to 20, Kev cleared this one to be in the same PR, bully him not me...

-Reverts: #861 to assist in init times.

-Re-balances the cost and duplicates of v4 AND v2's POIs, making them cost based on size, instead of loot, so that it dosnt run out of room and spawn cheap things that are massive and empty, but were cheap because they didnt have loot, then spam it over and over till it runs out of tries, increasing init times by a ton. (Hopefully this works as well as it did on my private tests)

-Adds a pretty animation for snow during transit that makes it look like flying, like how space and v4 sand does it.

-Makes snowsuit protect the head even without a hood item, sprite, or toggle, it just does now. You'll still need gloves, a mask, and boots, but it's got the head open for like.. a hat or helmet, at least until I sprite a hood for it at some point.


realbuffalowing:
https://www.youtube.com/watch?v=6SxjBWJPf78

PR #1084, opened 2019-04-03
Solarmoth Port

EtheoBoxxman:
//Details copied from original PR//

Adds the solarmoth to the game finally.
Stats
Aggression: Retaliatory
Health: 500
Melee damage: 1~7
Speed = 3 (negative is faster, higher is slower)
Armor: armor = list(
"melee" = 50,
"bullet" = 90,
"laser" = 90,
"energy" = 90,
"bomb" = 100,
"bio" = 100,
"rad" = 100)

SPECIAL FEATURE: This thing is like "TASTE THE SUN, MOTHER FUCKER!" It heats up a room to 10,000C. Sounds crazy, right? Don't panic. There is an easy way to melee this thing to death while in the room and the room is a total pressure cooker. The heat and pressure will not spread to other rooms unless you open the doors. Spoilers at the bottom. BUT! Also consider that maybe this can be used as a possible thermal engine? It's as hot as a white dwarf star under the right conditions, so maybe if you can find the right coolant? Also, trying to get it to the engine fast enough?

Adds code for the Solargrub to also track an how much power it has arbitrarily drained. Can possibly be used for other things as well. A global variable has been added in the Solarmoth code that the Solargrub keeps track of to see how many Solarmoths there has been in the world. Only up to two moths can ever be spawned in the world before the rest of the grubs decide to no longer evolve.

Also if you want to spoil yourself on how to deal with it, just use a fire suit and emergency air supply. It's impervious to the temperature and pressure, allowing you to safely deal with it.


peeesh:
what in the goddamn. we're citadel everyones an idiot. this just seems a bit nuts

CameronWoof:
Got these today. Three solarmoths spawned over the course of the round (to my knowledge, not more than two at once). They seem really cool, but the EMP effect at the end is really aggressive. It's super long-range and can permanently damage consoles, too. Consider toning that one back a smidge.

nik707:
Nerfs applied as follows.
>First stage heating increased to 350, from 200, as well as decreasing the rate at which it increases.
>Second stage heating power reduced from 9999999999999 to 9999999999, so the second stage is less punishing, and also reduced it's max temp to 5k down from 10k
>Reduced all EMP ranged by half

king5327:
Can we not merge this until the line endings are fixed. For some reason the files here use the wrong ones.

Sharkmare:
There's also a lunar moth variant in the [original code](https://github.com/CHOMPstation/CHOMPstation/blob/d4f981040556c30b44b661bf3a9852a823ffb4c7/code/modules/mob/living/simple_animal/vore/solarmoth.dm) it's nothing finished that they turn into but it is a lil fun variation i wrote that has some more spaghetti code in it

nik707:
Closing to inactivity

PR #1087, opened 2019-04-03
Fixes conveyors, for real this time

Auris456852:
## About The Pull Request

Now, this might not be that much better, I'll await input from those more experienced. But, after noting that objects that are already in the map set movement properly, I decided to just try delaying the movement setting slightly. And, well... It works. That's the most important thing, right?
With the wrench option now redundant, I replaced it for turning the conveyor. The icon does change together with the dir without calling an update icon function. I tested.

## Why It's Good For The Game

Being able to use conveyors as they properly should be used is nice.

## Changelog
:cl:
fix: fixed conveyors
/:cl:

### DIFF
--- a/code/modules/recycling/conveyor2.dm
+++ b/code/modules/recycling/conveyor2.dm
@@ -25,15 +25,18 @@
 	// create a conveyor
 /obj/machinery/conveyor/initialize(mapload, newdir, on = 0)
 	. = ..()
+
+
 	if(newdir)
 		set_dir(newdir)
 
-	if(dir & (dir-1)) // Diagonal. Forwards is *away* from dir, curving to the right.
-		forwards = turn(dir, 135)
-		backwards = turn(dir, 45)
-	else
-		forwards = dir
-		backwards = turn(dir, 180)
+	spawn(5)
+		if(src.dir & (src.dir-1)) // Diagonal. Forwards is *away* from dir, curving to the right.
+			forwards = turn(src.dir, 135)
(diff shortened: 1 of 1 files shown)


code/modules/recycling/conveyor2.dm
```diff
@@ -25,15 +25,18 @@
...
 	if(newdir)
 		set_dir(newdir)
 
-	if(dir & (dir-1)) // Diagonal. Forwards is *away* from dir, curving to the right.
-		forwards = turn(dir, 135)
-		backwards = turn(dir, 45)
-	else
-		forwards = dir
-		backwards = turn(dir, 180)
+	spawn(5)
```
kevinz000: spawn is banned use addtimer if you really can't use lateinitialize and i don't know why you'd need this anyways just separate syncing directions to another proc and then override setdir

Auris456852: Dude I don't even entirely know where the fuck set_dir is or if it's supposed to do literally anything but set the dir. Makes me a little cautious about doing things with it. Nor do I know all the new functions we're supposed to work with in this codebase, though the ones you mentioned sound rather self-explanatory. But I'll see about modifying it then, sometime when I can, and also bothering you about the former. It's not a hard fix, methinks.

nik707:
This hasn't been touched in months so I'm gonna close it in the interest of cleaning up PRs, feel free to reopen it when you've got the time to touch  it again

PR #1089, opened 2019-04-03
Lewdlaws - Requested by Ro *H-76*

Trilbyspaceclone:
Changlogs - 
Adds in two more AI Lewd laws!
Forceful Lust -Dom evil laws that are illegal much more
Submission - Sub laws to convert a borg into a love toy thing
Why?
Was requested by H-67 and such
This can lead to ERP and RP and such. I highly think they will never be used


Andromeda-K22:
Yeah, no.
Research can print these - they will be used to prefbreak people with 'no lewd' in their OOC notes as AI. Which, normally you can't see!

Also, Trilby, I hate to actually have to say this, but _your grammar here is absolutely atrocious. Have someone spellcheck and correct it, please._

peeesh:
this just seems... absolutely pointless, and even if people DID change laws, 99% of the time the ai and borgs would get fucking disassembled out of panic. solid no

nik707:
Yeah I'm gonna close this, hardcoded fetish stuff is a risk with our playerbase, and this is uh... questionable at best, because any random idiot will just law borgs to do bullshit. People can RP stuff like this.

PR #1091, opened 2019-04-04
Medical Access tweaks -WIP

Snownbloodyflowers:
## To-do before merge
**Do not merge, the door access fix causes runtimes in the script in code/game/jobs/access.dm line 40. I am unsure why it returns an error. ** 

Fix error here so the door actually works: 	
if(LAZYLEN(req_access - accesses)) // we don't have all the accesses we need
		return FALSE

## About The Pull Request
The Search and rescue looses virology, psychiatrist office access. The Psychiatrist looses morgue access. The chemist looses virology, morgue, and surgery access. In addition, the map was edited to reflect the bug that heads did not retain access to security foyer: See changes here graphically: https://imgur.com/a/IGrH2vV

I'm not sure why the psychiatrist-access tag is given to Search and Rescue. 

 In addition, first time using Git so try not to kill me.

## Why It's Good For The Game
The intent is to better differentiate the medical jobs and their required skill levels and general knowledge. There is also a map fix to allow for CMO/CE/RD access to the maps in the link shown below.

In addition, I would have restricted chemistry access to the CMO/chemist/Search and Rescue only but there was too much push back so that's for later.

## Changelog
:cl:
code deleted: Search and Rescue no longer has "access_virology, access_psychiatrist  "  on line 120 in maps/southern_cross/southern_cross_jobs.dm
code deleted: Psychiatrist no longer has access "access_morgue" on line 102 in code/game/jobs/job/medical.dm
code deleted: Chemist no longer has "access_morgue, access_surgery, access_virology"  on line 63 in code/game/jobs/job/medical.dm
code deleted: Paramedic no longer has "access_virology" on line 120 in code/game/jobs/job/medical.dm
Map tweak: Security airlocks at (44,82,1), (45,82,1), (57,82,1), (58,82,1), (57,76,1), (58,76,1)  have had list variable of access changed from (1,38) to (1,63)
/:cl:


CameronWoof:
Chemist - This job is already Really Bad Medical Doctor. Removing accesses they may actually need in an emergency is not a good idea.

Security Foyer - No. There is no reason for heads of staff to be wandering around inside Security. Their jobs do _not_ require that access.

Snownbloodyflowers:
The idea behind the chemist change is to give a medical job where people can go do something, and then go off to do what they want without a headset instead of being yelled at for not showing up when someone is dying like a doctor/paramed.

Security Foyer- I am going off of the fact that the heads _do_ have "security" access which was originally meant to be the front security doors. I can think of justification for the HOP/CMO, RD I dunno, CE is a maybe (greytide???). They shouldn't be there unless sec asks them to be. 

Regardless, I am however more concerned with the runtime that pops up when attempting to use the variable for the "Security" access with the doors.

CameronWoof:
I disagree with removing _everyone's_ ability to help in an emergency to cater to a minority who signs up for a job in Medical and wants to avoid occasionally being asked to help out.

Snownbloodyflowers:
The chemist probably shouldn't know how to fix people, or fix bones.  Here's what I was thinking of for level of mechanics understanding:
Medicine wise: 
CMO>Med Doc>SnR>Paramedic>Nurse>Intern
Chemistry wise:
Chemist>CMO>SnR>Para/Medoc/Nurse>Intern
Overall:
CMO>Med Doc>SnR> Paramed>Chemist>Nurse>Intern

The goal behind this PR (aside from a bugfix) are,
1)To allow for a clear delineation of skill between medical jobs
2)To allow chemist to be a somewhat entry level medical position

 But in order for people to play chemist, you've either got to:
1) Remove other people's access to chemistry (Which, when I asked about SnR everyone said no)
2) Make it so that there's an advantage to chemist to where they are now.

By removing the informal requirement of chemist knowing how to do medical related stuff, you can allow for Chemists to leave and be generally free to do what they want. Granted, your argument also applies, but I have never had a full medical staff on most of my shifts. 

Perhaps we could split nurse off into it's own category, and Nurses would not be expected to know anything but basic 4 chems, stryinges, IV, and "whatever the doc tells them to grab?". Medical Intern learning medical is completely dependent on *IF* the CMO/Doctors take time to teach them.

As for virology, we shall wait and see until when viral outbreaks happen again what to do, there's not much we can do aside from that.

CameronWoof:
You're trying to make Chemist more desirable by making it worse. I don't think that's the right direction. The reason people don't play Chemist is because Medical Doctor is Chemist++, and the reason Medical Doctor has access to Chemistry is because they don't have a job if no one is making chems.

If a Chemist doesn't want/doesn't know how to do surgery, they can say as much in-character. It's not their job, it makes sense that they wouldn't know how. 

However, removing the _option_ for Chemists to do surgery _if they need to_ is a straight-up nerf to an already very overshadowed job.

Snownbloodyflowers:
Please do not merge. I obviously have no idea how to fork correctly as I accidentally pushed my Protean tweaks onto this branch.

Cameron, the idea to remove the surgical possibility is to allow for players to NOT have to be requested to do surgery. The goal is for the chemist to be a job where the requirement of knowing surgery oocly is removed and that the players are not EXPECTED to do that at all. If you want to know chems, medicine, surgery, go play _CMO, Search and Rescue, or a doctor_. The change is to let Chemists have slightly more rp and not have to remain on alert throughout the entire shift like the a doctor.

Snownbloodyflowers:
Going to close this since I need to split them up. Sorry!

PR #1171, opened 2019-04-09
Fixes the observer prompt to actually match the respawn timer.

LiquidFirefly:
What it says on the tin.


RMagnum:
there are different respawn times depending on where you exit, the tram is instant, the teleporter is 5 minutes, and cryo tubes are 2, last I checked, but I dont use cryo anymore so *shrug.

Issue #1168, opened 2019-04-09
Proteans and large embedded objects


Snownbloodyflowers:
#### Brief description of the issue
Proteans when hit by large objects (like a mining drill) that embed themselves, and have the object removed, may heal but suffer reduced move speed like organic species.

#### What you expected to happen
-yank out object
-blobform
-Protean should be back to fine

#### What actually happened
-Yank out object
-Start gasping for breath
-paincrit in blob form
-Reform but be only able to move extremely slowly.

#### Steps to reproduce
- Throw a large object at someone Protean
- Yank out object once it gets stuck inside them.
- Profit!

#### Code Revision
- 2019-04-03
9094314542b0dac909bb3d94124a8db45cf1822c 
 
#### Anything else you may wish to add:
- May also be related, Protean always displays low pressure issue even when they are immune.

PR #1176, opened 2019-04-09
Fixes issue #1168

realbuffalowing:
Fixes issue #1168 and also defines prometheans better


realbuffalowing:
int too hard call me intalowing

nik707:
Thanks babe

PR #1302, opened 2019-04-21
Protean balance tweaks

Snownbloodyflowers:
Protean balancing. Current heal does not allow for proteans to heal large toxins without a long wait time (~5 minutes and with multiple times required to absorb steel) 100 units of steel heals 0.1 toxins. So total heal of 10,000 steel is... 10 toxins.

Suggestion, revert the latter change if balance proves to be good for the health regeneration tweak.

## Changelog
:cl:
Tweaked: 	O.damage = max(0,O.damage-0.1) to 	O.damage = max(0,O.damage-1). Allows for proteans to heal 100 toxins without waiting forever and running out of metal. Current rate is 0.1 for 100 steel, new rate will be 1 for 100 steel.

Tweaked:	if(health < maxHealth*0.5) to 	if(health < maxHealth*0.35) to allow for protean to still be somewhat useful in combat (current forces proteans to auto-blob at half health)
The max health change should be reverted if the heal rate is not as slow as currently.
:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
+++ b/code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
@@ -234,7 +234,7 @@
 
 	//Blob form
 	if(temporary_form)
-		if(health < maxHealth*0.5)
+		if(health < maxHealth*0.35)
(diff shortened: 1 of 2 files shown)


Snownbloodyflowers:
Please note that it is also impossible to refactor while taking any damage that automatically blobforms you AND the fact that refactors don't heal instability (AKA: Toxins).

PR #1309, opened 2019-04-21
Fixes NIF Surgery

realbuffalowing:
Nearly always caused brainstem injury when operating on the neck, which could result in death soon after without immediate treatment. Now, this step will not have deadly failure results. The surgery is troublesome as is and didn't really need this extra layer of difficulty.


realbuffalowing:
stop it travis

PR #1315, opened 2019-04-22
Fixes the panic bunker

kevinz000:
the issue is player_age is 0 until the next day after they connect.

WARNING: THIS IS NOT TESTED. I DO NOT HAVE THE MEANS TO SET UP A TESTING DATABASE ~~or I'm just too lazy to~~ so uh, testmerge.


Useroth:
The main problem was that it wouldn't let in literally _anyone_ at all, when enabled by default.

kevinz000:
working

PR #1365, opened 2019-05-02
replace ambiance with ambience

nik707:
This replaces all mentions of ambiance with ambience to make it more consistent to work with


realbuffalowing:
would you nerds merge this already

PR #1366, opened 2019-05-03
Removes voice changers from trash piles.

DigitalSquirrel95:
Because something that people will only use to be dicks to each other really shouldn't be this easily accessible.

### DIFF
--- a/code/game/objects/structures/trash_pile.dm
+++ b/code/game/objects/structures/trash_pile.dm
@@ -206,7 +206,6 @@
 					prob(1);/obj/item/device/flashlight/pen,
 					prob(1);/obj/item/device/paicard,
 					prob(1);/obj/item/weapon/card/emag,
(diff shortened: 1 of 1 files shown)


Keekenox:
Normally I'd say REEE DON'T REMOVE FUN STUFF, but yeah, I've seen it get abused so much and it's fucking annoying, especially if you are security.

Tename:
I don't know, I don't see it as a really big issue.

DigitalSquirrel95:
> 
> 
> I don't know, I don't see it as a really big issue.

Given I've had to delete quite a few at this point because people were using them to shit on another player, I'd have to disagree.

Keekenox:
Well one of the issues is, it usually ends up with one person impersonating another person's voice, and it usually causes quite a bit of turmoil. Another one of the issues is that you can't easily track down the person doing it.

PR #1374, opened 2019-05-13
thin fax machine

Hatterhat:
## About The Pull Request
makes fax machines density = 0
## Why It's Good For The Game
a fax machine is not dummy thicc enough to stop people from climbing around it
note: may make spare stealing easier and less easy to spot instantly due to "why is the fax machine on the fucking floor"
## Changelog
:cl:
tweak: fax machines are no longer dense
/:cl:


ForFoxSake:
I really want "build fax machines on tables"... But at the same time, when you're saying you're only doing this to be able to steal the spare... HRMM.

PR #1379, opened 2019-05-18
Removes Compliance Disks from Trash Pile Pool, Take 2

VarkaSelyn:
## About The Pull Request

It's #1378 but it actually compiles.

By request, after the results of an rp-emojiocracy poll. NIFSoft disks will remain in-code, but will only accessible via adminspawn, certain other RNG-heavy routes, or preplaced spawns, not randomly going through the garbage.

## Changelog
:cl:
del: Removed Compliance NIFSoft Disks from Trash Pile Loot
/:cl:

### DIFF
--- a/code/game/objects/structures/trash_pile.dm
+++ b/code/game/objects/structures/trash_pile.dm
@@ -216,7 +216,7 @@
 	return I
 
 /obj/structure/trash_pile/proc/produce_beta_item()
-	var/path = 			prob(6);/obj/item/weapon/storage/pill_bottle/tramadol,
+	var/path = pick(		prob(6);/obj/item/weapon/storage/pill_bottle/tramadol,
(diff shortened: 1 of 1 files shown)


VarkaSelyn:
whoops, guess I accidentally closed this whilst cleaning up branches yesterday

Tename:
I still think 51 to 36 is a bit iffy to remove something. Also the way the poll was made confused me the first time and I voted for the wrong thing as it says "Would you like compliance discs to keep existing." and thumbs up is for them to not keep existing.

VarkaSelyn:
As far as the poll goes, I'd rather have seen there be a "lower the drop chance but keep them as non-admin-only" similar to the change made in #1383 tbh. I expect that PR to be the one that gets merged if either are.

PR #1385, opened 2019-05-21
Head of Security sidearm gunbox

AnalWerewolf:
## About The Pull Request

Adds a gunbox to the HoS' gear locker. It is similar to the gunboxes available in the armoury, however this unique box contains five additional choices of sidearms. 

image

## Why It's Good For The Game

Allows a greater selection of sidearms for the Head of Security. While intended mainly to add a bit of flair for those playing the role, it also allows players the option of having a far less overkill lethal firearm than those available in the armoury, that being shotguns, laser rifles, or machine pistols.

More firearm options for the gunbox are planned from not only myself but other contributors too using this sidearm box, potentially not even limited to just sidearms if Kevinz gets around to implementing his ideas. 

An issue I do see with the sidearm box is that a HoS player may choose a sidearm using it, then cryo out later, disallowing new arrivals the ability to choose their own. However, adding the sidearm box to standard HoS kit may see multiple sidearms piling up in the HoS' locker depending on how many players end up playing the role in one round. Because of this, the placement of the sidearm box spawn may change in the future.

## Changelog
:cl:
add: Added the Head of Security sidearm box, featuring additional sidearm options.
tweak: Added the HoS sidearm box to the HoS gear locker. 
/:cl:

### DIFF
--- a/code/game/objects/items/gunbox_vr.dm
+++ b/code/game/objects/items/gunbox_vr.dm
@@ -16,3 +16,26 @@
 			if(istype(AM, /obj/item/weapon/gun))
 				to_chat(user, "You have chosen \the [AM]. Say hello to your new friend.")
 		qdel(src)
+
+/obj/item/gunbox/hos
+	name = "Head of Security sidearm box"
+	desc = "A secure box containing a security sidearm."
+
+/obj/item/gunbox/attack_self(mob/living/user)
+	var/list/options = list()
+	options["M1911 Dynamic (.45)"] = list(/obj/item/weapon/gun/projectile/colt/detective, /obj/item/ammo_magazine/m45/rubber, /obj/item/ammo_magazine/m45/rubber)
+	options["'Consul' Revolver (.44)"] = list(/obj/item/weapon/gun/projectile/revolver/consul, /obj/item/ammo_magazine/s44/rubber, /obj/item/ammo_magazine/s44/rubber)
+	options["SW 625 Revolver (.45)"] = list(/obj/item/weapon/gun/projectile/revolver/detective45, /obj/item/ammo_magazine/s45/rubber, /obj/item/ammo_magazine/s45/rubber)
+	options["Webley Service Revolver (.44)"] = list(/obj/item/weapon/gun/projectile/revolver/webley, /obj/item/ammo_magazine/s44/rubber, /obj/item/ammo_magazine/s44/rubber)
+	options["Lumoco Arms HE Colt (.357 LETHAL)"] = list(/obj/item/weapon/gun/projectile/revolver, /obj/item/ammo_magazine/s357, /obj/item/ammo_magazine/s357) // Yikes, lethals only for .357s. I'll have to fix that later.
+	options["'Mateba' Revolver (.357 LETHAL)"] = list(/obj/item/weapon/gun/projectile/revolver/mateba, /obj/item/ammo_magazine/s357, /obj/item/ammo_magazine/s357)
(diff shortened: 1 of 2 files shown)


AnalWerewolf:
@kevinz000

ForFoxSake:
More guns for the gunsman.

I'm personally not keen on this pull, I'd prefer to see two things change with it:
1. This should be exclusive with the prototype energy gun. I'm all for the choice between energy or ballistic, but this is just going to remove that choice, as Sec Heads will always choose both now.
2. Don't give antag calibres to a station role. Just, c'mon dude.

AnalWerewolf:
> 
> 
> More guns for the gunsman.
> 
> I'm personally not keen on this pull, I'd prefer to see two things change with it:
> 
>     1. This should be exclusive with the prototype energy gun. I'm all for the choice between energy or ballistic, but this is just going to remove that choice, as Sec Heads will always choose both now.
> 
>     2. Don't give antag calibres to a station role. Just, c'mon dude.

I was tempted to include the head of security's prototype energy gun in the box buuuuuut, the prototype energy gun is an antagonist objective item, not having it spawn consistently on station would obviously cause issues there. Sure, we don't have antagonists on RP right now, however there is talk about having regular secret rounds someday. 

It's unfortunate that the .357 is the antagonist calibre, it's currently horrendously overpowered compared to similar calibres to the point it makes no sense whatsoever. .357 is also unfortunately only available with lethal ammo.

I have a PR in the works that standardizes bullet, casing, projectile, magazine, and gun code across the board and balances .357 and .44 in line with the other pistol calibre counterparts while also adding hollow-point, armour piercing, and rubber variants to all pistol calibres. Perhaps this PR would be far more appropriate with the rebalanced bullets.

Thanks for the feedback.

VarkaSelyn:
Is the Security Chief's MPEG actually a traitor objective? Are you sure you're not thinking of the _Captain's_ antique, with its display case and all?

Also hardmode suggestion; lock the powerful calibers to single-action only revolver for non-antags. Yes, you have a stupid powerful gun option, but one that has low capacity and is much more fiddly to operate, compared to the all-in-one, window-bypassing power of an energy weapon like the X-01 MPEG.

Thus, for the _practical_ Chief, the MPEG or a semi-auto are the sane choices; they do everything better, faster.

Antags still get easy-to-use double-action handcannons, obvs.

AnalWerewolf:
> 
> 
> Is the Security Chief's MPEG actually a traitor objective? Are you sure you're not thinking of the _Captain's_ antique, with its display case and all?
> 
> Also hardmode suggestion; lock the powerful calibers to single-action only revolver for non-antags. Yes, you have a stupid powerful gun option, but one that has low capacity and is much more fiddly to operate, compared to the all-in-one, window-bypassing power of an energy weapon like the X-01 MPEG.
> 
> Thus, for the _practical_ Chief, the MPEG or a semi-auto are the sane choices; they do everything better, faster.
> 
> Antags still get easy-to-use double-action handcannons, obvs.

Ah! Yeah turns out the MPEG isn't a traitor objective. Gotcha. I'll probably add the MPEG into the box at a later date, currently the firearms available in this box are a bit much, true, though the Head of Security is allowed to use, and is encouraged, to have any type of non-explosive lethal firearm so long as it's hidden from sight, a 60 damage hand cannon is a bit much and would cause issues when players inevitably would abuse this privilege. 

That is not something I intend to do with this PR.

Closing this PR until my gun code rebalance and standardization PR is ready. 
Thank you both for the feedback.

PR #1392, opened 2019-05-27
lol

LetterJay:
lol


modular_citadel/code/modules/clothing/under/accessories/donatoraccessory.dm
```diff
@@ -19,7 +19,6 @@
 /obj/item/clothing/accessory/collar/holotags
```
deathride58: be sure to make this a subtype of `/collar/donator` to make sure it inherits the `icon` and `icon_override`

PR #1404, opened 2019-06-03
Custom Ringtone Preferences

avabee:
## About The Pull Request

Adds the ability to specify a custom ringtone in preferences.

## Why It's Good For The Game

It was requested and money was offered. It's also a good feature.

## Changelog
:cl:
add: Adds the ability to set a default ringtone for a character. If left blank, the default will be used.
/:cl:


Tename:
Looks nice, but could we not make boom the default tone for science, it seems rather unfitting for RP science as toxins is rarely used.

Perhaps a new option with the PAI startup tone could be used?

PR #1419, opened 2019-06-17
Penumbra station

BlackMajor:
Obviously, still a work in progress.
This is just a PR for people to come and look in at it when I request them to or at their own leisure.

## Welcome to PenumbraStation.
Or its officially known station name, [No name yet] on planet Nerada 8.

PenumbraStation is a little project of mine consisting of a 512x512 single z-level planet station with far more hostile wildlife.
Just stay inside the perimeter, and you'll be fine.
Dont worry about the turrets firing at something unseen in the dark.

Nerada 8 is a chilling frigid planet with a breathable atmosphere, albeit cold enough to need a coat.
The station in question here was settled by NanoTrasen to research the long since dormant infestation on the planet simply known as "The Aberrant". A strange sprawling mass of organic matter, rather akin to coral under the sea, spewing out various monstrosities. Albeit in the last hundred years or so.. the progress of such matter that spans the planet seems to have halted.. gone dormant or extinct even.

That's what your station is here to find out. What happened here, and what can be gained from it?

Nanotrasen tells you that this planet is "Totally safe to work at!" and that it's just "Dormant, like an extinct volcano." You believe them, don't you?

We're in frontier space, expect a little danger.

More to come as it gets developed.

- [ ] Initial room pass through (Equipment, tables, structure etc.)
- [ ] APC pass through
- [ ] Atmospherics/Disposals pass through
- [ ] Misc pass through (Newscasters, fire extinguishers, timeclocks etc)
- [ ] Decals


KillianKirilenko:
Coming along nicely, but it seems like the whole thing is a bit too big? Between the number of triple-wide corridors and things like that barfront with something like fourteen seats, it feels like the station is being built with a very large playerbase in mind. Can't really comment on specific areas given most of them are still very WIP.

If you're planning to trim away the excess once the main details are squared away though, it'll probably be fine.

buffyuwu:
can't wait to see this!

PR #1423, opened 2019-06-21
Changes up compliance disk OOC mechanics

kevinz000:
Well for one they now log.

Compliance disks have been changed to only be effective as a scene tool, rather than a general gameplay-capable mechanic.

tell me if the wording is shit or something i typed this up while half conscious

### DIFF
--- a/code/modules/nifsoft/software/15_misc.dm
+++ b/code/modules/nifsoft/software/15_misc.dm
@@ -107,14 +107,14 @@
 		if((. = ..()))
 			to_chat(nif.human,"<span class='danger'>You feel a strong compulsion towards these directives: </span><br><span class='notify'>[laws]</span>\
 			<br><span class='danger'>While the disk has a considerable hold on your mind, you feel like you would be able to resist the control if you were pushed to do something you would consider utterly unacceptable.\
-			<br>\[OOC NOTE: Compliance laws are only a scene tool, and not something that is effective in actual gameplay, hence the above.\]</span>")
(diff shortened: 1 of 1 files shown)


Tename:
I still feel like they should have a purpose more than lewding. There are plenty of interesting scenarios and events blocked off by this and it would be a bummer for it to be an OOC item. I also don't really see the purpose for doing this.

Keekenox:
I want the drama from these to stop as somebody that plays security. Any security player would agree that these things are a fucking headache.

Tename:
This doesn't stop any drama, in fact I would argue more drama were to come from this by putting compliance disc's in some limbo of icly existing but not existing.

Keekenox:
Honestly you're right. To that I would say just remove them then.

Tename:
RP is always going to be filled with drama and it's less mechanics than it is the player base. People hold grudges, pull shit on for too long, and make crap up, removing compliance discs or making them this weird limbo of IC and OOC isn't going to fix that. This will piss people off and just generally be confusing and frustrating for security. Removing them is going to cause half of the community to be angry. If you want this to be merged or disks to be removed all together with the purpose reducing drama, it's not going to work. 

I would also like to add that security comes with the expectation of dealing with drama and if it is frustrating you should take a break from that role.

Snownbloodyflowers:
Leave the item in as a spawnable admin item. Otherwise just remove it from the trash pile spawn chance (or set it to like 1% or something).

Tename:
:/

PR #1432, opened 2019-07-01
Removes the gutter whitelist

AnalWerewolf:
## About The Pull Request

Surprisingly not even close to the most snowflaky language we have.

## Why It's Good For The Game

More languages, just don't be stupid with it and claim you're an epic syndicate. The language can hint at a poor upbringing or life event, remember that your character is working for a megacorp.

## Changelog
:cl:
tweak: Removed the Gutter language whitelist
/:cl:

### DIFF
--- a/code/modules/mob/language/station_vr.dm
+++ b/code/modules/mob/language/station_vr.dm
@@ -92,5 +92,5 @@
 /datum/language/seromi
 	flags = 0
 /datum/language/gutter
-	flags = WHITELISTED
+	flags = 0
 	machine_understands = FALSE


ForFoxSake:
I realised a potential issue with this PR the other day, but didn't see it was open, so apologies on the tardyness of this comment.
All in all, I'm largely eager to see more languages become available.

I have caught a small snag in specifically removing this from the whitelist, however.
Gutter, as I understand it, is an underground language often used by criminals. It's intention is to allow shadier characters to organize together to do shady shit.
But if the whitelist is removed, there's nothing from letting a HoS main from taking the language to powergame their way past any reasonable RP, just to get some valids in the brig.

Again, not against the whitelist being removed, just concerned there might be issues with it.
If the whitelist gets dropped, how will issues like these be dealt with?

AnalWerewolf:
> 
> 
> I realised a potential issue with this PR the other day, but didn't see it was open, so apologies on the tardyness of this comment.
> All in all, I'm largely eager to see more languages become available.
> 
> I have caught a small snag in specifically removing this from the whitelist, however.
> Gutter, as I understand it, is an underground language often used by criminals. It's intention is to allow shadier characters to organize together to do shady shit.
> But if the whitelist is removed, there's nothing from letting a HoS main from taking the language to powergame their way past any reasonable RP, just to get some valids in the brig.
> 
> Again, not against the whitelist being removed, just concerned there might be issues with it.
> If the whitelist gets dropped, how will issues like these be dealt with?

Indeed, a Head of Security may take the gutter language without penalty, but as it currently stands many players already take languages their characters wouldn't realistically know, this presents an issue in that if we are to enforce such restrictions on gutter, we should be doing it for other languages as well, like a Vulpkanin knowing Sagaru, or a human speaking fluent Daemon and Canilunzt.

If this is to be approved, I'd see it being handled ICly rather than any rigid enforcement, the same way most admins currently handle characters. We do have rules stating that a character must be believable, but you can simply claim that your character took the time and effort to learn three languages and have it be within the realm of possibility.

PR #1440, opened 2019-07-10
Unwhitelists Gutter

CameronWoof:
Whitelists are stupid. In the case of this language in particular, whitelisting makes it completely pointless as no one can even speak it.

Also makes it understandable by translators, because no widely-spoken language should be untranslateable. That's also stupid.


kevinz000:
duplicate + fixed

PR #1456, opened 2019-07-30
Hmm yes today I will pr 'working' code

RMagnum:
Just checking what travis is upset about and considering if it's worth finding to fix and ask for merge, since everything client side works. If you know what it is, let me know.


RMagnum:
hmm yes today I will scrap my project because of hidden issues that travis wont explain

PR #1459, opened 2019-08-03
Changes to allow explorers to use any webbing or holster

AnalWerewolf:
## About The Pull Request

Title.

## Why It's Good For The Game

Hatter had a PR up doing this shit but the changes weren't applied to code. Weird because as a SaR all the webbings are accessible in game. The holsters are so that Explorers/SaRs/PFs can take an exploration phaser or a backup pistol in lieu of a machete.


AnalWerewolf:
Oh I've just realized why Hatter's change worked but the changes aren't appearing. He actually used modular_citadel. I can't believe it.

PR #1475, opened 2019-08-15
Toricrates: Remastered: Complete Edition 2019

Michiyamenotehifunana:
:cl: Toriate
imageadd: Updated the sprites of all the regular crates
/:cl:

This PR will resprite crates (again).

Quick comparisons of a few crates which have been given much needed makeovers:
3Is3ehesam
kBsDGuYgnY
rfU8orhkGk
qrB6US66Vh
FG3vOr6Pny

As can be inferred from the comparisons, the red "THIS WAY UP" arrows have been removed from all of the crates.

In-game screenshot:
dreamseeker_dJvtncaoF8


CameronWoof:
These are so much less colorful and fun than the originals. I don't really like them as much as the originals we already use.

AnalWerewolf:
Cameron has a bit of a point. Imo a brighter palette would be much more in line with the very loosely defined artstyle we currently have.

PR #1482, opened 2019-08-21
Fixes wheelchair movement.

DigitalSquirrel95:
Direct port of this https://github.com/VOREStation/VOREStation/pull/5087
Fixes wheelchairs so they aren't stupid fast and we have to keep bwoinking people over it instead of just fixing the problem.


TheFurryFeline:
This would be one of the few times where I approve of removing an untended feature that's potentially abusable for powergaming.

PR #1484, opened 2019-08-21
Delay before unbuckling someone from a wheelchair.

JGlitch:
Adds the usual do_mob() delay and a message warning when doing so.

This is important because it's kind of LRP that you can throw someone off a wheelchair in a single click, insta-stunning them.

### DIFF
--- a/code/game/objects/structures/stool_bed_chair_nest/wheelchair.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/wheelchair.dm
@@ -130,8 +130,12 @@
 		MouseDrop(usr)
 	else
 		if(has_buckled_mobs())
-			for(var/A in buckled_mobs)
-				user_unbuckle_mob(A, user)
+			user.visible_message("<span class='warning'>[user] starts undoing the straps of the wheelchair</span>")
+			if(do_mob(user,src, 20))
+				for(var/A in buckled_mobs)
+					user_unbuckle_mob(A, user)
+			else
+				user.visible_message("<span class='warning'>[user] has stopped undoing the straps of the wheelchair</span>")
 	return
 
 /obj/structure/bed/chair/wheelchair/CtrlClick(var/mob/user)


AnalWerewolf:
Needless delay.

JGlitch:
> Needless delay.

Because being able to instantly stun someone only using your hands with a single click in a HRP server is better than having a small delay to prevent it.
Try pushing someone off a wheelchair, it aint that easy, you know? Even if the person cant even move their body, you gotta put some effort into doing so.

lolman360:
>known shitter buffing tool commonly associated with shitters

wew

DigitalSquirrel95:
Not the place for that, lolman, seriously.

JGlitch:
> > known shitter buffing tool commonly associated with shitters
> 
> wew

This is an objectively good change and a serious topic, not a popularity contest.

kevinz000:
Why should we make it so we can't force unbuckle someone from a vehicle which otherwise protects them from grabs and makes them immune to most forms of slowdowns even if it's about to get nerfed to be slower?
What good does this do for the server because I'm not seeing the point here at all.

kevinz000:
Also the real meme is no amount of popularity can force a PR through based on that alone so lmao

JGlitch:
> Why should we make it so we can't force unbuckle someone from a vehicle which otherwise protects them from grabs and makes them immune to most forms of slowdowns even if it's about to get nerfed to be slower?
> What good does this do for the server because I'm not seeing the point here at all.

Now that you put it that way, that's a good point, to be honest. But I still think some extra changes are needed honestly, maybe shorten the delay even more so its not INSTANT.

JGlitch:
I thought it was a good idea, and I was proud of it as it was the first time I coded anything, at all.
But thinking about it even when the wheelchair gets its speed reduced it will be pretty overpowered if you dont get any slowdowns from tasers and cant get grappled, closing it down until I figure how to fix those issues too. Wheelchair revamp soon..!

PR #1504, opened 2019-09-03
TEG Additions

Andromeda-K22:
## About The Pull Request

+ TEGs can now be lubricated with spacelube to give a decent boost in thermal efficency and prevent power loss from sparking. TEGs use 0.01 units of lubrication per tick, so it will last a while - they hold 120 units. you'll probably get 2 hours out of one fill, probably more.
+ At ABSURD power levels, +10 MW from a TEG, they will now unleash a tesla zap. Overcharge engines are your own risk. Just use a grounding rod to cancel out the zap, make sure it's anchored.

+ Yes, they are deadly.

## Why It's Good For The Game

It adds some risk to making absurd thermal engines and gives engineering a reason to talk to chemistry.

## Changelog
:cl:
+ TEGs can now be lubricated
+ TEGs now teslazap at absurd power levels.
/:cl:

### DIFF
--- a/code/modules/power/generator.dm
+++ b/code/modules/power/generator.dm
@@ -4,7 +4,7 @@
 	icon_state = "teg"
 	density = 1
 	anchored = 0
-
+	flags = OPENCONTAINER
 	use_power = 1
 	idle_power_usage = 100 //Watts, I hope.  Just enough to do the computer and display things.
 
@@ -22,12 +22,20 @@
 	var/lastgen2 = 0
 	var/effective_gen = 0
 	var/lastgenlev = 0
+	var/lubricated = 0
+
 
 /obj/machinery/power/generator/New()
 	..()
+	create_reagents(120)
 	desc = initial(desc) + " Rated for [round(max_power/1000)] kW."
 	spawn(1)
 		reconnect()
+
+/obj/machinery/power/generator/examine(mob/user)
+	..()
+	to_chat(user, "It appears to have [reagents.total_volume] units of lubricating fluid left.")
+
 
 //generators connect in dir and reverse_dir(dir) directions
 //mnemonic to determine circulator/generator directions: the cirulators orbit clockwise around the generator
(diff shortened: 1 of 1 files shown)


Andromeda-K22:
Diffs are fucked and i'm not sure why.

nik707:
Problems I have.
The lubrication is basically just pointless because you just go talk to chem for 3 seconds every 2 hours, not to mention if a chemist isn't around it just means nothing. I feel like giving an over 50% boost to possible output for a fairly simple chemical is a BIT much. On top of that, the tesla zaps make it more likely for uneducated paramedics to get killed when trying to save someone who fucked up during creation. I don't *really* see the gain in this. I like the IDEA, but I don't like this execution. I don't dislike it enough to say a flat no but I'd like to see Kevinz' thoughts before I make any call.

Andromeda-K22:
It's not 50%. It is *15*%. In any case, the tesla zaps are still on a five percent chance every tick - either you hear them, or get incredibly unlucky while darting in and out to grab the corpse of the person who screwed it up. 

I'm not entirely sure how to make this more meaningful in the respect you wish. I have other things planned, such as making them up-gradable with research. A simple solution is unfortunately the best I have; multi-reagent lubrication is ... well, kind of *a pain*. I tried it on another codebase and decided that it was too much work for not enough benefit.

nik707:
image
23%. Anyway, upgrade-able through science seems better, hell even a cargo order for better shit. Also, 5% per tick?
Math works out to about 1 per 20 seconds if Kevinz' is right on this subsystems tickrate being 1/s, which seems a little high. Also, if you forget a rod or someone fucks it up or something you can say g'bye to the engineering control room. This is a spicy one

Tename:
I really don't like the overcharged zap function, engineers shouldn't be punished for doing their jobs well.

Andromeda-K22:
It's not a punishment, it's *danger*. You can always build a rod to deal with it.

Tename:
It's a negative effect for creating a good engine. Either way it makes no sense that basically the equivalent of a motor should release a bolt of electricity. And the fact that these tegs are usually in the engine room just increases the chance of it hitting something important and causing a delem, which I don't think many people are really looking forward too.

Andromeda-K22:
I don't think anyone on the regular is going to be making +10MW per TEG engines, aside from about three people who understand the physics behind it.

nik707:
Stale, the player in question is no longer active, and it was already not a positive reception

PR #1509, opened 2019-09-05
Ports multi directional dogborg rest sprites from virgo.

BlackMajor:
Also adds animations to some others, and adds sit sprites for cit exclusive dogborgs.

Example sits.

Doggos

Edit: Oh right I forgot to mention, also moved the modular_citadel widerobot stuff to widerobot_vr. We're not downstreamed like that anymore so there's no need for them to be in seperate files.


BlackMajor:
Diff bot ded

KillianKirilenko:
Has the alt. sechound's flank badge changing colors been fixed? You can even see it here where the badge is white on the standard rest pose but cyan on the sitting pose (and, I think, regular standing pose). Might want to square that away before final merge.

nik707:
can we rename it from widerobot who's idea was that wtf

BlackMajor:
> Has the alt. sechound's flank badge changing colors been fixed? You can even see it here where the badge is white on the standard rest pose but cyan on the sitting pose (and, I think, regular standing pose). Might want to square that away before final merge.

I did yeah

PR #1514, opened 2019-09-07
Attempts to fix travis unit tests failing due to unsimulated turfs getting added to ZAS zones

deathride58:
My suspicion is that this is due to an off-by-one error since it only happens on map edges, but let's go ahead and see if any of these darts I'm throwing at the wall are gonna stick

## Changelog
:cl:
fix: Added a sanity check to the add() proc in zone datums, which should hopefully fix the issue where unsimulated turfs get added to ZAS zones and cause runtimes that make travis fail
/:cl:


code/ZAS/Zone.dm
```diff
@@ -171,7 +173,7 @@ Class Procs:
 	M << name
 	for(var/g in air.gas)
 		M << "[gas_data.name[g]]: [air.gas[g]]"
-	M << "P: [air.return_pressure()] kPa V: [air.volume]L T: [air.temperature]�K ([air.temperature - T0C]�C)"
+	M << "P: [air.return_pressure()] kPa V: [air.volume]L T: [air.temperature]�K ([air.temperature - T0C]�C)"
```
nik707: Okay maybe this is just the PC I'm on right now but this is showing as question mark error characters and this worries me

PR #1529, opened 2019-09-15
Ports /tg/ input system, adds diagonal movement and multiz movement keybinds, changes some admin bindings around. Defaults scaling to nearest neighbor.

kevinz000:
## About The Pull Request

Ports the entirety of SSinput and most /tg/ macros
Function key binds are now /tg/'s instead of ours so say/ooc/etc are all rebound.
Multiz movement now has keybinds ` and ~ and , and . for up/down, second set is for those using arrow keys.
Diagonal movement is added.
Defaults scaling mode to NN

## Why It's Good For The Game

Diagonal movement's sweet.
Nearest Neighbor makes stuff sharper. Added a pref for it too.

## Changelog
:cl:
experimental: Ported /tg/ SSinput system, diagonal movement added, etc etc.
/:cl:


Skycave1:
Let's not and say we did? I enjoy the current system of movement, and having is change it to a different system would only promote poor experiences. We don't need Diagonal movement or changed hotkeys. It's only a mess.

ForFoxSake:
I was about to type up some points about one or two things that've changed here, and then I noticed the fucking size of this pr.
Movement, input, keybinds, macros _and_ scaling.
I'm no maintainer here, so I probably shouldn't be the one to call this stuff out. At any rate, you'd make everyone's lives easier if you could split this into three, maybe four seperate PRs for each system so that they can be critiqued and bugtested individually without holding certain features that people seem to want (diagonal movement) behind other features people seem to strongly not want (tg hotkeys).
It's difficult to post a comment on any one particular change in amongst 3,000 changes, especially when half of them are regarding features that're completely irrelevant to the thing you're trying inspect.

kevinz000:
hotkeys are staying as rp default.

diagonal movement is hand in hand with smooth movement and similar things as decoupling it is more effort than it's worth, and there's no reason to not move to a more lagless input system.

scaling is bundled in because the interface files are being changed and I figured it's better to change it all at once because the systems that need also happen to tie into new input.

PR #1534, opened 2019-09-20
map fix that probably doesn't work (temporarily switches back tgm to dmm on tether runtime loaded levels)

kevinz000:
if this actually fixes the "map is sliced in half" issue i'm going to go ape


nik707:
image

PR #1539, opened 2019-09-22
Makes Phoron scrubbers stronger

Mewchild:
## About The Pull Request
Port of https://github.com/VOREStation/VOREStation/pull/5470
Makes the Phoron Scubbers work almost twice as fast, while taking slightly more power. They now heat the air while working, preventing atmos alarms from the cold. They now filter down to a lower phoron amount, preventing leftover phoron in the connecting area.

## Why It's Good For The Game
Makes getting into and out of the station easier, faster, and safer. Mainly a QoL change, as waiting for an airlock to cycle is booty.

## Changelog
:cl:
tweak: Makes Phoron Scrubbers work much faster, adds air heating to them as well
/:cl:

### DIFF
--- a/maps/tether/tether_phoronlock.dm
+++ b/maps/tether/tether_phoronlock.dm
@@ -91,6 +91,28 @@ obj/machinery/airlock_sensor/phoron/airlock_exterior
 	radio_connection.post_signal(src, signal, radio_filter = RADIO_AIRLOCK)
 	return 1
 
+/obj/machinery/portable_atmospherics/powered/scrubber/huge/stationary/phoronlock		//Special scrubber with bonus inbuilt heater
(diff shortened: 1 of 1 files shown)


Mewchild:
I saw someone complaining about it in the discord, and realized "Wait, they fixed that on Vorestation"

kevinz000:
i'm going to be honest the only reason i'm passing this pr is because i'm not going to require anyone to refactor atmos machinery code to not be bad but the whole "manually call scrub air for 3-4 times manually rather than calling once and just adjusting how much it scrubs during that one call" and the whole "override everything instead of put everything on base types" drives me fucking insane but hey, atleast it'll be fixed soon tm.

PR #1544, opened 2019-09-24
Revert "adv ore scanner fixes but also not fixes"

nik707:
hatter is dumb and fucked this up so trav now fails. thanks hatter


nik707:
## UNIT_TEST: IDs - The following entries have the same value - rigmod_orescanner: /datum/design/item/rig_module/orescanner and /datum/design/item/rig_module/orescanneradv
## UNIT_TEST: !! FAILURE !! [RESEARCH: Designs Shall Be Unique]: 1 issues with research designs found.

PR #1557, opened 2019-10-16
[READY] Ore Scanner Range Toggle (Altclick) + Ore Boxes No Longer Explode

Hatterhat:
## About The Pull Request
Ore scanners can now have their range fiddled with when alt-clicked. Ore boxes no longer get destroyed on explosions.
## Why It's Good For The Game
Ore scanners - More precise usage of scanners, primarily the advanced scanner variant. Only available for handhelds, todo is adding the modifying functionality to the hardsuit module itself.
Ore boxes - I crashed the server once by blowing up an ore box that was stuffed with ores. This is a stopgap measure. (Why doesn't it just have variables or datums or whatever for ores? I don't know, but I hate it. ORMs when, kevinz?)
## Changelog
:cl:
add: Ore scanners can now have their range switched around with an alt-click!
tweak: Ore boxes no longer explode and drop all their contents on ex_act(), meaning that they probably shouldn't crash servers when they get broken. I hope.
/:cl:

### DIFF
--- a/code/game/objects/structures/transit_tubes.dm
+++ b/code/game/objects/structures/transit_tubes.dm
@@ -56,7 +56,7 @@
 
 
 // When destroyed by explosions, properly handle contents.
-obj/structure/ex_act(severity)
+/obj/structure/transit_tube_pod/ex_act(severity)
 	switch(severity)
 		if(1.0)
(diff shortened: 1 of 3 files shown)


Keekenox:
Doing gods work

PR #1563, opened 2019-10-20
Shortening xenochimera reconstitute form timers and deleting cooldown + extra

DiscordWizard:
## About The Pull Request

So I found out that the reconstitute form has a literal one hour long cooldown even if you use it when not dead, and an obscene ammount of wait time for what is basically a healing protean change-form. This pull request shortens the ammount of wait time and completely removes the cooldown to reconstitute again, as well as adding the trashcan ability to the species since it was suggested that something that's basically a living consuming virus amalgamation should be able to eat trash.

Tested it in a private server, shouldn't break anything, shouldn't give a free out of death card, if it does you know who to blame. Race's whitelisted either way.

## Why It's Good For The Game

Honestly it doesn't give xenochimera that big of an advantage, they still need help when they die to revive and this is going to be a way to either heal themselves when out of danger like the living virus they are and give them the ability to properly change shape and appearance without having to stare at their monitor for 10 minutes and then wait an hour to change again if they need to. What's the point in giving these species said abilities, if they can use it only hourly with already a cost added to it.

## Changelog
:cl:
add: Added trashcan ability to xenochimera
balance: Lowered xenochimera timer for reconstituting related to available nutrition
code: Deleted cooldown code
/:cl:


nik707:
Has this been properly tested?

DiscordWizard:
Did test it on a private server and so far it worked, trashcan ate trash, took less time to be horizontal and revive and could do it again right after with more time as a result.

Didn't change or break anything else, would still die if killed mid reconstituting and still needs outside help to revive

nik707:
aa conflict

DiscordWizard:
What be an aa conflict and how do I solve this, cause in-game it seemed to have no issues.

PR #1597, opened 2019-11-06
Small tweaks/fixes for the RPD.

ForFoxSake:
## About The Pull Request
This PR makes three tiny changes to the RPD that should make it more consistent to the end user.
1: The Wrench Mode button no longer turns grey when it's turned off. (One person reported not seeing it at all.)
2: Pipes that aren't paintable, no longer print pre-painted. While I enjoyed abusing this issue during development, it turns out some pipes disappear entirely if printed pre-painted. Whoopsie.
3: Fast layering has been entirely removed. When I ported the device I wasn't really sure what this feature did, and I think it's fairly clear no one else does either.
Fast layering was a feature that meant you could click a pipe, and the RPD would attempt to automatically attach the next pipe to that one. Obviously, this is confusing, as you might not notice you clicked a pipe, and begin to wonder why the pipes are now printing at your feet instead of where you're clicking.

## Why It's Good For The Game
This should improve user interactions with the RPD, making using it faster, and learning to use it faster.

## Changelog
:cl:
tweak: Tweaked the RPD to be more readily usable.
/:cl:

### DIFF
--- a/code/game/machinery/pipe/pipe_recipes.dm
+++ b/code/game/machinery/pipe/pipe_recipes.dm
@@ -9,17 +9,17 @@ var/global/list/all_pipe_recipes = null
 /hook/startup/proc/init_pipe_recipes()
 	global.atmos_pipe_recipes = list(
 		"Pipes" = list(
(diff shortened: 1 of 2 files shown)


Tename:
May I ask whats fast layering?

PR #1604, opened 2019-11-12
Fixes the default Mapmerge2 config

Tename:
## About The Pull Request
Fixes the maps file path changing it from _maps to maps
Disables convert to TGM

## Changelog
:cl:
fix: Changes the map path on the Mapmerge2.bat files from /_map/ to /map/
fix: Disables convert to TGM format by default.
/:cl:

### DIFF
--- a/tools/mapmerge2/Prepare Maps.bat
+++ b/tools/mapmerge2/Prepare Maps.bat
@@ -1,5 +1,5 @@
 @echo off
-cd ../../_maps/
+cd ../../maps/
 
 for /R %%f in (*.dmm) do copy "%%f" "%%f.backup"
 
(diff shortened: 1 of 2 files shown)


deathride58:
Why disable TGM conversion?

Tename:
> Why disable TGM conversion?

I'm pretty sure we don't use it on RP considering Virgo doesn't even use map merge and the original map files from them, I just wasnt sure if it would cause issues or not.

KillianKirilenko:
Uhhh no. #1519 updated mapmerge - it should already have fixed this, and TGM is now **MANDATORY** or else MapDiffBot will reject all of your changes.

Tename:
Alright

PR #1603, opened 2019-11-12
makes space bikes faster and reduces cargo cost

Tename:
## About The Pull Request

Increases the speed of space bikes on land, and reduces cargo cost.

## Why It's Good For The Game

They have a dumbly high power usage and the batteries last about 10 minutes, they are also slow. Instead of changing the power usage I made them significantly faster

Cargo cost has also been reduced because I feel it is a bit much, while doing this I noticed that for some gay reason ATV TRAILERS(Not the ATV itself) were priced at some dumb shit and reduced it.

## Changelog
:cl:
balance: Increase bike speed
tweak: Reduces bike cost from cargo from 350 to 200
tweak: Reduces ATV trailer cost from 250 to 50
/:cl:

PR #1610, opened 2019-11-21
Spacebike buff

Tename:
## About The Pull Request
Increases the speed of space bikes on land, and reduces cargo cost.

## Why It's Good For The Game

They have a dumbly high power usage and the batteries last about 10 minutes, they are also slow. Instead of changing the power usage I made them significantly faster

Cargo cost has also been reduced because I feel it is a bit much, while doing this I noticed that for some gay reason ATV TRAILERS(Not the ATV itself) were priced at some dumb shit and reduced it.

## Changelog
:cl:
tweak: Reduces ATV trailer cost from 250 to 50
tweak: Reduces bike cost from cargo from 350 to 200
balance:  Increase bike speed
/:cl:

### DIFF
--- a/code/datums/supplypacks/robotics.dm
+++ b/code/datums/supplypacks/robotics.dm
@@ -189,7 +189,7 @@
 /datum/supply_pack/robotics/bike
 	name = "Spacebike Crate"
 	contains = list()
-	cost = 350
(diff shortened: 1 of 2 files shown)


Tename:
#1603

PR #1621, opened 2019-12-26
ports virgos new midpoint

nik707:
## About The Pull Request

port of https://github.com/VOREStation/VOREStation/pull/5890

see there for more info


KillianKirilenko:
isn't kev already porting this as part of hardsync's limited map changes?

either way, make it happen

PR #1631, opened 2020-01-05
ADds an APC and relay to transit

nik707:
## About The Pull Request
transit has radio now

## Why It's Good For The Game
comms are nice

## Changelog
:cl:
add: relay and APC to transit level

/:cl:


Mewchild:
If you're touching this area, the disposals sorter in the maint behind the elevator also needs to be flipped. Arrow needs to be pointing north.

PR #1643, opened 2020-01-08
Changes Firing range doors to have same access as research door

### LINKED ISSUE #1642: Researchers can't use Research Firing Range
So science staff can't actually open the windoors on the research firing range. Not even the RD.

Mewchild:
Closes #1642 
Changes doors in the firing range to have reqaccess(47), which is the same as the research front desk.


TheFurryFeline:
Try re-opening the PR. BYOND was having issues earlier in the day.

PR #1645, opened 2020-01-10
Adds stuff for QM head, makes QM a head

yourdoom9898:
## About The Pull Request

It gives QM a reason to exist by making them a head

## Why It's Good For The Game

As it is now, there is no reason for QM to exist, as they're effectively a glorified CT with a office and no actual authority. By making them a head, it fixes that, and causes them to be held accountable slightly more than regular command by being a part of command (therefore limiting Cargonia memes)

## Changelog
:cl:
add: Uncommented out the QM encryption key
add: Added QM headset and QM bowman headset, both of which replace the normal supply headsets in the QM locker
tweak: Gave QM basic bridge access
/:cl:


yourdoom9898:
I fucked up, let me try this again

PR #1646, opened 2020-01-10
Adds stuff for QM head, makes QM a head: Take 2, now with less dumbfuckery

yourdoom9898:
## About The Pull Request

It gives QM a reason to exist by making them a head

## Why It's Good For The Game

As it is now, there is no reason for QM to exist, as they're effectively a glorified CT with a office and no actual authority. By making them a head, it fixes that, and causes them to be held accountable slightly more than regular command by being a part of command (therefore limiting Cargonia memes)

## Changelog
:cl:
add: Uncommented out the QM encryption key
add: Added QM headset and QM bowman headset, both of which replace the normal supply headsets in the QM locker
tweak: Gave QM basic bridge access
/:cl:


CameronWoof:
Honestly, I don't like this. Cargo is too powerful a department to exist without _outside_ regulation and oversight. I like that the current department head isn't part of the office itself. It's separate from and above the personalities and politics of the department, and it requires that the QM answer to someone else when they want to order ridiculous things or expensive trash for themselves.

yourdoom9898:
@CameronWoof And yet the Research Director and Chief Engineer are perfectly fine with only having oversight by the Captain? Also, the Head of Personnel before this is the only head with oversight on 2 departments (Supply and Service), and because of that, it ends up with the QM existing in this weird spot where they are supposed to be a boss of the Supply department, but have exactly 0 actual authority besides what autonomy the Head of Personnel gives them.

BlackMajor:
I'm down for this.

KillianKirilenko:
As a neutral party who doesn't care one way or the other? The way I see it, this really won't change much.

QM plays stupid games? Either way they're gonna win stupid prizes, be it brigtime for contraband/smuggling or a jobban for misuse of the role.

The HoP's "oversight" amounts to writing sternly-worded letters to Central, which is what you should be doing when the QM fucks around and orders dumb shit *anyway.* Their officially a Head doesn't suddenly grant them diplomatic immunity.

nik707:
After discussion with the staff team, this is, for now, a no go.

PR #1647, opened 2020-01-10
Adds a few events into rotation, takes PDA spam out

yourdoom9898:
## About The Pull Request

Adds the cargo bounties back into rotation, removes PDA spam from rotation

## Why It's Good For The Game

The bounties give a otherwise bored cargo something to do, and PDA spam is just obnoxious and spawns _every_ shift

## Changelog
:cl:
tweak: Changed event rotations
/:cl:


CameronWoof:
No on appendicitis. One of those scene-crasher events that ends up being way too disruptive. Probably why it was removed.

PR #1649, opened 2020-01-12
Ports fully realized timeclock system

Mewchild:
This is a port of the timeclock system from vorestation, that allows players to go on and off duty via the timeclock. Full details can be read here https://github.com/VOREStation/VOREStation/pull/5101

This will need to be testmerged to make sure it works. There was no way for me to verify that PTO was being built, on a local server, as I don't have a SQL database set up.


Crystal9156:
100% for this, it is very useful.

Mewchild:
This is currently testmerged, and everything seems to be working as intended. This is ready to fully merge.

PR #1654, opened 2020-01-14
Completely remakes the maint bar near the construction dorms

Mewchild:
This PR ports over a change I made in July of last year, on Vorestation, in which the maint bar is greatly renovated to make it easy to repair/use.


Mewchild:
image

PR #1660, opened 2020-01-16
Revert "More Misc Map fixes."

nik707:
Reverts Citadel-Station-13/Citadel-Station-13-RP#1652


Putnam3145:
thanks technically-git-friendly-but-not-really formats

PR #1663, opened 2020-01-16
Shelter-B-GON, Hydroponics Moved

KillianKirilenko:
## About The Pull Request

They say if you want something done, do it yourself. So I did. This PR replaces the useless-ass fire/phoron shelter with a slightly-expanded hydroponics bay. The old hydroponics area is now the 'vacant shop' area.

The surrounding area has been reworked a little too;
image

I consider this 'ready to merge' in its current state, but am open to feedback or suggestions, especially from players with more hydro experience.

Also adds a strip of a new 'bar maintenance catwalk' area and an APC to the exterior of surface 3 just outside the bar, so the lights outside the bar will be powered.

## Why It's Good For The Game

The phoron/fire shelter was pretty much never used at all ever by anyone? I mean I guess you could go fuck there but really, who would choose it over a dorm room?

Hydroponics is now bigger so there's lots more room to grow stuff.

You can now go straight between hydroponics and the kitchen via the freezer/garden too. So that's neat?

## Changelog
:cl:
add: New exterior catwalk area by the bar, to power the external lighting.
tweak: Removed the old fire shelter, replaced it with a bigger hydroponics bay.
tweak: Old hydroponics is an ugly empty area for now.
/:cl:

### DIFF
--- a/maps/tether/tether-02-surface2.dmm
+++ b/maps/tether/tether-02-surface2.dmm
@@ -2972,12 +2972,6 @@
 	},
 /turf/simulated/floor/tiled/techfloor,
 /area/maintenance/lower/bar)
-"gC" = (
-/obj/structure/catwalk,
-/obj/structure/ladder/up,
-/obj/effect/floor_decal/industrial/outline/yellow,
-/turf/simulated/floor/tiled/techfloor,
-/area/maintenance/lower/bar)
 "gD" = (
 /obj/structure/catwalk,
 /obj/structure/cable{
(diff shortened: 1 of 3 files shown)


VailTheWolf:
The phoron shelter is anything but useless. General reminder that phoron floods all 3 Z levels during serious breaches. Some events are being switched around which means these will become frequent possibly. There's already a hydroponics. Improve it first.

KillianKirilenko:
I'm not against the *concept* of the shelter, but I think it would be safer (and smarter) to have designated 'shelter zones' within given departments, rather than a singular shelter in a far corner, due to the size of the map.

Let's say we have a major phoron/fire incident in science. The shelter is far enough away that there's very little risk of the incident reaching it, but at the same time the shelter is useless to the people who might need it in that particular moment, the scientists.

As for events, the most probable cause would be a meteorite and those are extremely rare. I don't think any 'new' dangerous events have been added, or will be.

PR #1670, opened 2020-01-26
Readds a removed proc from my fluff item

Mewchild:
Re-adds a proc to this item, that stops it from losing it's custom name/description when the name is set


nik707:
@LetterJay

PR #1734, opened 2020-03-21
Revamped food changes part 1, also known as: Burgers only contained 4 useful food units while shakers had 65 omegalul.

Enzo-Leon:
## About The Pull Request

This will be handling the first big re-balance of food changes for the server. We're hoping to have this set in a way that chefs are no longer overshadowed by one singular food item they couldn't even make.

This is only a singular part of this adjustment. I want to eventually do at least three more waves of balancing as players report differences in their play experience, and if chefs need further buffs or a bit of dialing back to ensure the flow of the round is not affected in a negative manner.

## Why It's Good For The Game

QOL adjustments, that will also help add extra balance for players that were forced to rely on one single food item to not die.

## Changelog
:cl:
tweak: Lots o' number crunching and adjustment.
balance: Balanced Shaker to contain 30 protein and 10 nutriment vs 35/30. The protein value it to not immediately screw over species that depend on this, and can slowly figure out food from the chef can keep them alive a bit easier now.
spellcheck: One food contained a "/" in the name. Removed it and added the proper capitalization to its name.
code: Adjusted a large amount of food items to contain more protein, and nutriment where needed. Most meats now will contain more than 1 or 2 units of protein to help with crew that rely on this.
/:cl:

### DIFF
--- a/code/modules/food/food/snacks.dm
+++ b/code/modules/food/food/snacks.dm
@@ -295,31 +295,32 @@
 	reagents.add_reagent("tricordrazine", 8)
 	bitesize = 3
 
-/obj/item/weapon/reagent_containers/food/snacks/candy
+/obj/item/weapon/reagent_containers/food/snacks/candy // Buff 4 >> 8
 	name = "candy"
 	desc = "Nougat, love it or hate it."
 	icon_state = "candy"
 	trash = /obj/item/trash/candy
 	filling_color = "#7D5F46"
-	nutriment_amt = 1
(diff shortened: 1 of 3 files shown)


code/modules/food/food/snacks.dm
```diff
@@ -1579,13 +1584,13 @@
 	reagents.add_reagent("water", 10)
```
Mewchild: Why does this not give protein, if it's made using meatballs?

PR #1748, opened 2020-03-28
Changes Compliance disk description, removes it from trash piles, adds to l…

yourdoom9898:
…oadout

## About The Pull Request

Changes the Compliance disk's description to make it not illegal, removes it from trash spawns, and adds it to the Loadout for 2 points

## Why It's Good For The Game

Right now, these are a inconsistent scene tool, and treated as illegal IC for a reputation it can't manage due to IC and OOC rules. This allows them to be much easier to get for scenes, and not a giant headache for everyone if caught with one by sec.

## Changelog
:cl:
add: Added compliance disks to loadout for 2 points
del: Removed discs spawning in the trash piles
tweak: Compliance disk description
/:cl:

### DIFF
--- a/code/game/objects/structures/trash_pile.dm
+++ b/code/game/objects/structures/trash_pile.dm
@@ -215,7 +215,7 @@
 	return I
 
 /obj/structure/trash_pile/proc/produce_beta_item()
-	var/path = pick(prob(10);/obj/item/weapon/disk/nifsoft/compliance, //Citadel Override probability, 3.6%
+	var/path = pick( //prob(10);/obj/item/weapon/disk/nifsoft/compliance, Citadel Override probability, disabled
 					prob(6);/obj/item/weapon/storage/pill_bottle/tramadol,
 					prob(4);/obj/item/weapon/storage/pill_bottle/happy,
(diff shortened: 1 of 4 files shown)


Keekenox:
It makes very very little sense for something like this to be legal.

kevinz000:
yeahhhh no
they are fine as is, it makes no sense for this to be legal IC even if it's only a scene tool
the scene tool clause was added to reduce the ridiculous amount of NIF compliance disk drama from back when the rp server had high player count when certain people kept abusing this and didn't get banned for it so hey here we go, hotfix patch time. 
if you want a disk for scene purposes go look for one and don't get caught by security with one. this is a game, circumstances change per round, which means i feel it is actually more interesting if things like these, which are mostly only useful for erp and similar private-ish rp things, are not just given the "you can grab one from loadout" treatment. we also have body snatchers. would those be too if people liked it? no. dms exist for these things if you don't want to deal with the dynamic nature of a *video game*.

yourdoom9898:
Except body snatchers can actually fulfill the reason they're considered illegal, while the disks can only _act_ like they do but really can't due to OOC rules, so they now just exist in the weird limbo of being far less powerful than they're portrayed in character

PR #1766, opened 2020-04-04
Virology crates now stocked, plus new BENEFICIAL symptoms. Otherwise known as, "lol lynch viro anyways."

Enzo-Leon:
## About The Pull Request

Virology has been a quiet/empty spot for the server since before I got here that's for sure, and even if it was accessible it had almost no benefit for players except antags. This is set to be a series of changes to bring Virology into a spot that is fitting for the server and community, and provide exploration and others some bonuses in their rounds.

## Why It's Good For The Game

Bringing in more things for medical to do in their shifts, and provide some possibly fun RP situations. Eventually we will have a less RNG heavy method of virology, and a much larger swath of positive, and neutral, symptoms for the medical team to find.

## Changelog
:cl:
add: Virology crate now comes stocked with 4 random virus samples
add: new possible symptoms to get.
balance: nerfed the stage 4 viruses to not be as round-ruining as they were currently.
code: New symptoms added, basic stuff for now. Will be adding much more better symptoms if we proceed with this.
/:cl:

### DIFF
--- a/code/datums/supplypacks/medical_vr.dm
+++ b/code/datums/supplypacks/medical_vr.dm
@@ -28,3 +28,11 @@
 	containertype = "/obj/structure/closet/crate/secure"
 	containername = "Virology biohazard equipment"
 	access = access_medical_equip
+
+/datum/supply_pack/med/virus
+    name = "Virus sample crate"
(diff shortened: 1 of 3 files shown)


Enzo-Leon:
Closing as this did not process properly and add tags or notify the server.

PR #1797, opened 2020-04-12
protean fixes

lolman360:
im best c0d3r


code/modules/mob/living/carbon/human/species/station/protean_vr/protean_powers.dm
```diff
@@ -235,9 +235,11 @@
 		to_chat(src,"<span class='warning'>You don't have a working refactory module!</span>")
 		return
 	if(refactory.processingbuffs == TRUE)
```
Akrilla, line 237: probably worth removing the == TRUE here even though you didn't touch this

PR #1895, opened 2020-04-18
MORE MORE MORE, I REQUIRE MORE CLOTHING

thestubborn:
from boh and bay, mostly
also some eyehud from main
whirrrrrr

oh yeah and i removed some references to >KHI
donate to my onlyfans


code/modules/clothing/glasses/hud_vr.dm
```diff
@@ -1,6 +1,6 @@
 /obj/item/clothing/glasses/omnihud
 	name = "\improper AR glasses"
-	desc = "The KHI-62 AR Glasses are a design from Kitsuhana Heavy Industries. These are a cheap export version \
+	desc = "The KHI-62 AR Glasses are a design from Vey Med. These are a cheap export version \
```
Nethaufer, line 3: I would suggest changing KHI to another abbreviation like 'VM'. KHI as an acronym also means nothing in our lore.

Nethaufer: Same goes for the KHI in the other glasses you modified.

thestubborn: oh never even thought of that, will do

thestubborn: done btw

thestubborn:
#1901

PR #1902, opened 2020-04-19
Vox pressure suits.

SpessInquisitor:
my first code please no bully.

For real, though, this should add the Vox pressure suits to the game's inventory, and add the generic assistant pressure suit to test if the code works properly. If it works, then we can code in the rest of the suits, for a pressure suit per department. Who said phoron boys should have all the pressurized fun?


Nethaufer:
We in boys.

PR #1907, opened 2020-04-22
Let Shadekin die.

BlackMajor:
Being canonically unable to die in-game is a bit too handwavey for my taste.

If they get blasted and die, then let them actually die.

Webedit PR man good.


lolman360:
misleading title, thought it was removing shadekin

PR #1929, opened 2020-05-02
THE PR ABOUT SPEED: sprinter module, hyperzine

Hatterhat:
## About The Pull Request
- hyperzine metabolism 4x slower (desc says long lasting)
- sprinter now actually makes your suit go fast
liquidprotein sprite tweaked to be closer to the other two ration packs, liquidvitamin sprite given proper icon name
## Why It's Good For The Game
gofast
## Changelog
:cl:
balance: Hyperzine now lasts much longer in the bloodstream.
tweak: The sprinter module now should be more effective in hardsuits.
/:cl:


lolman360:
might want to drop the od to 10 then down from 15 if you're making it last 4x

PR #1941, opened 2020-05-06
Parts Are Important

Nethaufer:
## About The Pull Request
Thermal Regulators use New when they're called to be made which was no longer being called for mapped in Thermal Regulators which also meant they spawned with no component parts. This fixes that by swapping New with Initialize.

## Why It's Good For The Game
Parts are important

## Changelog
:cl:
fix: Thermal Regulators get their parts should they exist through inits.
/:cl:


Hatterhat:
hey can you expand this to the explosive simulator

Nethaufer:
Yeah I can do it tomorrow morning.

PR #1942, opened 2020-05-06
Removes the teleporter's prob(5) to oof you.

BlackMajor:
It's supposed to be called when the teleporter is uncalibrated.
But guess what, teleporters on this code don't do calibration, so that prob(5) is always called.


BlackMajor:
Wait I'm retarded there's a right click verb called "test fire" that sets the accuracy.

That's pretty unintuitive, I'll address that later.

PR #1948, opened 2020-05-07
Prevents SMES overload from Breaking Charging on other SMESes

Nethaufer:
## About The Pull Request
Copies PolarisSS13/Polaris/pull/7089


Nethaufer:
Closing because apparently clamp causes problems for 512 users.

PR #1953, opened 2020-05-08
Prevents Cargo Ordering Terminal Shenanigans

Nethaufer:
## About The Pull Request
Removed If/Else logic that caused the supply ordering terminal to allow approving, denying, and sending the shuttle instead of just being limited to the control terminal. Control terminal requires cargo access to open again.

## Why It's Good For The Game
Locking the control terminal via an if/else does very little good if you can just approve orders on the non control terminal anyway.

### DIFF
--- a/code/game/machinery/computer/supply.dm
+++ b/code/game/machinery/computer/supply.dm
@@ -14,6 +14,7 @@
 	var/active_category = null
 	var/menu_tab = 0
 	var/list/expanded_packs = list()
+	authorization = 0
 
 // Supply control console
 /obj/machinery/computer/supplycomp/control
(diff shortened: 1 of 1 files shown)


code/game/machinery/computer/supply.dm
```diff
@@ -14,6 +14,7 @@
 	var/active_category = null
 	var/menu_tab = 0
 	var/list/expanded_packs = list()
+	authorization = 0
```
yourdoom9898, line 17: Isn't this already declared on Line 10?

Nethaufer: So it is, I completely overlooked it because I saw the var/ and just erased it from my mind. Fixed now.

PR #1961, opened 2020-05-12
Fixes Voidsuit Injection Ports Not Functioning

Knouli:
## About The Pull Request
After our hardsync with VORE we've been unable to utilize the 'injection ports' of voidsuits in order to administer chemicals through syringes. I initially thought this a balance change, a nerf, to heighten the importance of hyposprays, but after exploring the code for syringes and seeing that the 'injection port' code is intact, I figured his was an error.

This PR adds two checks in the human.dm where 'injectability' was checked - the lines in question checked if the material covering a given bodypart was thick and rejected efforts to inject into bodyparts covered by 'thick' clothing pieces. This originally worked in conjunction with the syringe code to allow for voidsuits to be injectable, despite being classified as 'thick', but since the hardsync has been in error. This may be due to the way we're not handling THICKMATERIAL definitions, migrating from items_clothing.dm to flags.dm, or it could be something dealing with priority of human.dm 'checking' before syring.dm is checking to see if they're wearing a voidsuit and failing the attempt without checking if they are wearing a voidsuit.

This is a bit of a hacky fix, and would be better if it could point to a list of 'permitted' thick clothing that could be injected through, but in our code we had two exceptions - voidsuits, which was in VORE code originally, and a Citadel edit to permit plate carriers being injected into. I elected to not implement our edit for plate carriers as all other armors comparable, such as standard security armor, also blocks syringe injections.

This change doesn't adjust the jank that is the lazy syringe code check for voidsuit helmets and suits - that is to say, attempting to inject into someone's bare head while they are wearing a voidsuit will still have you attempting to 'look for an injection port' despite their head being bare. Similarly, injecting into someone's head while wearing a voidsuit helmet but not body will inject without any special text. This was how the code was working originally, and the scope of this bug fix was simply to allow voidsuits to have their injection ports once more.

For sanity's sake, the use of an ignore_thickness injector, the CMO's hypospray Mk2 to be specific, was tested and maintained proper functionality. 

Finally, this preserves injection blocking for all non space-suit suits, suits such as the rad suit or bio suit, which are THICKMATERIAL flagged suits and helmets still block syringe injection attempts.

image

## Why It's Good For The Game
As far as I am aware, this is a bug rather than a balance issue. If it is intentional for voidsuits to no longer have injection ports, then I am in error. But the fact that the injection port code still exists leads me to believe that the removal of functional voidsuit injection ports is a bug.

This will permit the standard interaction of voidsuits and syringes as expected - ie, affording an 'injector' to search for an injection port in another character's voidsuit.

## Changelog
:cl:
fix: Voidsuit injection ports working once more
/:cl:


lolman360:
to summarize this multi-para pr: voidsuit inject port broke, knouli did hacky fix.

kevinz000:
this is amazing

PR #1963, opened 2020-05-12
No More Magic Power

Nethaufer:
## About The Pull Request
Newly made machines respect the area power instead of just being made.
Newly made machines default to stat = 0 which means they have power which is fine in most cases but if made in an area with no power (no APC) or an area where the machine power is disabled on the APC, they pull power anyway. Now they detect power properly.

### DIFF
--- a/code/game/machinery/_machinery.dm
+++ b/code/game/machinery/_machinery.dm
@@ -120,16 +120,17 @@ Class Procs:
 	var/speed_process = FALSE			//If false, SSmachines. If true, SSfastprocess.
 
 /obj/machinery/Initialize(mapload, newdir)
+	.=..()
 	if(newdir)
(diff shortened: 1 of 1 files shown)


code/game/machinery/_machinery.dm
```diff
@@ -120,16 +120,17 @@ Class Procs:
 	var/speed_process = FALSE			//If false, SSmachines. If true, SSfastprocess.
 
 /obj/machinery/Initialize(mapload, newdir)
+	.=..()
```
kevinz000, line 123: are you sure this is a good idea?
sometimes, stuff depends on initialization proc call order being in a certain way.

Nethaufer: Made the change back, just in case

PR #1985, opened 2020-05-16
Gives Vox an outfit for the you that believes in the me that believes in you.

SpessInquisitor:


BlackMajor:
Gurren Lagaan was my favourite anime.

PR #1994, opened 2020-05-20
massively nerfs security/buffs carp/simples

lolman360:
simplemobs no longer take burn damage from halloss

### DIFF
--- a/code/modules/mob/living/simple_mob/defense.dm
+++ b/code/modules/mob/living/simple_mob/defense.dm
@@ -144,7 +144,8 @@
 
 /mob/living/simple_mob/get_shock_protection()
 	return shock_resist
-
+// tasers no longer damage simples. Break out the lasers.
+/*
 // Shot with taser/stunvolver
 /mob/living/simple_mob/stun_effect_act(var/stun_amount, var/agony_amount, var/def_zone, var/used_weapon=null)
 	if(taser_kill)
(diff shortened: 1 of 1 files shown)


VailTheWolf:
How about no? Warden, HOS, CD are never 100% on shift all the time. Lethal distribution should be limited to threats that require lethals, like the blob or event characters. Security would have absolutely no ability to respond to incapacitated miners in the underdark unless they powergamed into the armory.

lolman360:
security shouldn't be responding to incap miners in underdark, paramedics should. Why shouldn't carp/grubs/etc require lethals? If they happen on lowpop one secoff probably won't be able to handle it all even with a taser.

VailTheWolf:
I do not have to bear the burden of proof in this, it's not my PR. All you've done is make the change in the code and asked "Why not?" You haven't followed the format, and you sure as hell haven't given a proper explanation to why this needs to change. Nonetheless I'll humor you for the moment. 

Security is a proper department to help secure miners in a hostile environment. Carps and grubs do not warrant a code change worthy of lethal deployment, thus letting security respond with their normal gear is fine. These are two minor level threats. I'm going to bring up an old argument here as well. "Expedition team shouldn't get ballistic lethals because they Blue on blue too often." If that argument was allowed to stand, then security responding to simple mob threats with "non-lethal to crew" weapons should too. If we gave out lasers for every runtime event, then tasers, stun revolvers, etc, would become mute except for roleplaying. 

"Lasers don't get enough use as it is." Then change that by other means. Get events going. Maybe make it so that certain sub-types of simple mobs require lethals instead of blanketing the entire simple mob stats. (I.e. fix and then PR xeno events.)

lolman360:
actually yeah hold on i can just override the proc on protean blob level

PR #2016, opened 2020-05-24
Protoblobs can use radio and IDs now.

lolman360:
Protean blobs should be able to use radio now and can use ID.
For some reason radio messages don't always seem to go through but I have no clue what's up with that can someone help aaaaa


lolman360:
saycode a shit, don't merge until i can figure out why their radios only work 10% of the time

lolman360:
okay now it works all the time BUT they eat the radio's encryption keys

PR #2032, opened 2020-05-28
preference tweaks

Detective-Google:


Detective-Google:
also fixes an oopsie from kev where he created an entire tab for one command that didn't work
image

PR #2043, opened 2020-06-02
aaaaAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA (belt mining take 2)

lolman360:
it works now, what the fuck

port of VOREStation/VOREStation#6492

tldr: roguelite mining area. populated with mobs and loot spawns.
procgenned.


Mewchild:
https://files.catbox.moe/fql16p.png
Hallway is one tile wide to and from the area. No belts from the unloader. No control console for the smelter. Shuttle has missing turfs around it
Miners don't have access to the exterior door, meaning they can't come back in
Airlock is only one tile, can't move equipment inside. Same with the shuttle halls being 1 tile wide, which means you can trap yourself with your equipment

yourdoom9898:
- Control Console doesn't link to smelter, which eats all mats
- Airlock is impossible to use with ore boxes without very creative door exploiting
- Void jetpack just doesn't work
- the shuttle is missing the corners on the windows
- shuttle could *really* use a tcomms relay to not leave miners stranded

lolman360:
@yourdoom9898 jetpacks don't work in suit slot, only back slot. This is intended behavior as far as I'm aware.
The "missing" corners on the windows never existed in the first place.
I'll see about adding a relay but it's impossible to leave miners stranded as the only way to move shuttle is the console on the shuttle itself.
Airlock access was fixed in a previous commit, testmerge needs to be updated.

You don't have to be creative to get the airlocks to work, just force shuttle exterior/all of main airlock, push the orebox through to the last tile inside the shuttle airlock (by the controller) and it's done however I'll look at making it a 2x2 airlock as that's the most common complaint.

console/smelter i have no idea what's up but i'll see if i can fix that

yourdoom9898:
@lolman360 The airlock doesn't work because if you turn it to cycle, it errors out because it doesn't have a pump and scrubber, and therefore freaks out.

The corners on the shuttle should really be there so the window isn't held on with space magic

I meant stranded from station contact, so if they kick the bucket, no one knows until the backup bitches, if they have one.

lolman360:
this needs map changes i'm waiting on someone else from

PR #2048, opened 2020-06-04
Fixes Portable Shield Diffusers

Nethaufer:
## About The Pull Request
Portable shield diffusers were added to the locker contents for engineering so that they could remove shields in certain areas, however they are completely unable to be recharged, bar the recently added inducers. This allows cells to be swapped with the portable devices as well as be inserted as a whole into device rechargers.

## Why It's Good For The Game
Shield diffusers should be able to be recharged and not just tossed away when used up.

## Changelog
:cl:
fix: Allows shield diffusers to have their cells swapped as well as recharged in rechargers.
/:cl:


lolman360:
is it possible to make this a small item if it's running off a device cell? just a suggestion

PR #2049, opened 2020-06-04
Buffing Medigun and reducing costs

Enzo-Leon:
## About The Pull Request

Current Medigun cells are REALLY weak for what they costed and what you got from them. This buffs the core cells to be a lot more useful.

## Why It's Good For The Game

QOL and makes combat situations in the future easier for crew to overcome.

## Changelog
:cl:
balance: reduced cost of cells by 10% over all materials
balance: buffed core medigun cells by 2x
/:cl:


lolman360:
being able to recharge the entire magazine at once would probably be a better buff than just doubling the power of the shots

PR #2050, opened 2020-06-05
Adds the Ice Pick

lolman360:
Tin. Adds an ice pick. Small item, can fit in pockets, somewhat robust (1 less brute than a hatchet) but is able to be used as a mining tool and can also be used to scale cliffs. 2400 mining points in the vendor.


yourdoom9898:
To be perfectly honest, this does too much for too cheap. It's a pocket golden pick that can scale cliffs, and with minimal effort can be turned into switchblade spam. At the very least, 15 dig speed is *way* too fast for a one handed ice pick.

lolman360:
@yourdoom9898 up to 25 digspeed now

PR #2051, opened 2020-06-05
Generic Persistence Framework - Requested by Drof

kevinz000:
## About The Pull Request

<3 drof

Adds a generic persistence framework that persists objects per map
Unfortunately, only supports loading onto whatever turf it's on (or its closest turf) for now.

Just call AddElement(/datum/element/persistence) on any /obj to persist it. It will save at round end (triggered by ticker) and load on bootup to whatever place it was when it last saved.

To have object data persist, override serialize_list() and deserialize_list() and have an object save/load data that way. The element will call it when saving/loading respectively and create objects as necessary.

Also adds bunker persistence for admins, and parts of the new map system.

## Why It's Good For The Game

Persistence framework for gimmicks

## Changelog
:cl:
add: A generic persistence system has been added.
/:cl:


lolman360:
we Persistence Station now

PR #2054, opened 2020-06-08
Hyperzine nerf/buff

BlackMajor:
@kevinz000 decided to be a powergaming catgirl again and merge OP things
https://github.com/Citadel-Station-13/Citadel-Station-13-RP/pull/1929
This PR made hyperzine last for a meme amount of time.

It's too good right now.

So this PR removes Hyperzine's speed buff, and instead makes it remove any pain slowdown and item related slowdown without increasing your speed.

PS. Y'know I just realised this removes the speed up from changeling epiniphrene, nuka cola and the electric spider toxin since those relied on the chem speedboost.
I'm gonna call that intended since [meme speed] infuriates me in regards to balancing anything.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_movement.dm
+++ b/code/modules/mob/living/carbon/human/human_movement.dm
@@ -28,8 +28,9 @@
 		tally += (health_deficiency / 25)
 
 	if(can_feel_pain())
-		if(halloss >= 10)
-			tally += (halloss / 10) //halloss shouldn't slow you down if you can't even feel it
+		if(!(CE_SPEEDBOOST in chem_effects)) //Hyperzine stops pain slowdown
+			if(halloss >= 10)
+				tally += (halloss / 10) //halloss shouldn't slow you down if you can't even feel it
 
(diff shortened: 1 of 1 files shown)


Hatterhat:
- every other bay-fork codebase has hyperzine lasting longer than most other chems
- pain/item slowdown nullification seems really powerful for "i'm going to press three buttons on this chem dispenser"
- funny enhancement vests probably won't play nice with the ignoring item tally slowdowns, maybe move the CE_SPEEDBOOST check on that one

nik707:
I'm... not a fan of this. Get kevinz to look

AvaricePleonexia:
Watch me destroy my whole argument by prefacing it with saying, "How do balance, good?" (In other words, I'm no balancing expert). _However_, I do like the removal of meme speed; secondly, I appreciate the removal of slowdown effects caused by carrying items when hyperzine is active.

Gives it a different role as a temporary hardy trait. Me likey.

yourdoom9898:
This just makes it a discount painkiller without you know, being a painkiller, and during it's main use case will end up just having you go into paincrit way before the chem fades and you legs give out

PR #2055, opened 2020-06-08
Sledgehammers

BlackMajor:
Ports, fixes and adjusts sledgehammers from ChompStation.

Sledgehammers can be created by using any material with a hardness > 50 (Steel, plasteel, durasteel etc) and creating a hammer head.

The hammer head then requires rods, welder, rods, welder, steel sheets, welder, wrench to finish constructing.

A held sledgehammer is a two handed weapon that functions similarly to a fire axe (can door pry, smash windows, cleave), but deals slightly less damage than a fire axe.
Sledgehammers are too large to fit in backpacks or backs, and have slowdown 1.5.
Sledgehammers also have 50% armour penetration since, y'know.. it's a hammer.

Also includes an admin spawn only Mjolnir. Does what you'd expect.


nik707:
Okay a few things here.
Remove door prying, how the fuck would a sledgehammer pry doors?
Also, what if you could use wire on both a sledge hammer and an axe to fashion a sling to put in in the backpack slot

PR #2080, opened 2020-06-14
Removes Long Distance Transmission from Inducers

Nethaufer:
## About The Pull Request
There was no check to keep someone from hopping onto a camera with an inducer in hand and remotely recharging things. That's a bit... yeah no. This fixes that.

It also fixes inducers locking up if you try to recharge a gun with its 'recharge_weapons' var set to 0.

## Why It's Good For The Game
Prevents... interesting scenarios with cameras and inducers.


Nethaufer:
Hell of a time for the Gitbot to start working again.

PR #2092, opened 2020-06-16
Make Cells Appear to be Empty on Fabrication

Nethaufer:
## About The Pull Request
Tin

## Why It's Good For The Game
So people don't think their cell is full when printing a new one.

## Changelog
:cl:
fix: Cells appear empty when fabricated with no energy
/:cl:


BlackMajor:
Should probably have specified in your PR description that this PR also makes items that have powercells in them start not charged as well.

Whatever, it's a change I'm in favour of anyway.

Nethaufer:
They... aren't supposed to be empty on new item spawn

Zeldazackman:
New cells are empty too, most things start depowered and a good number of things such as the drills and rigs are worthless without emptying out the cell first and then babysitting them at a recharger for the next while.

PR #2094, opened 2020-06-16
Adds breath heating masks for nicc

silicons:
this code is not tested

tuned to last at -80C for just a bit over 10 minutes.

0.0125 power units per kelvin heated per breath heated/second/whatever (timing units are weird when it's not processed on a single subsystem)


AvaricePleonexia:
I really want to see this work; this kind of stuff would be super cool (haha, temperature pun) with the map that's currently in development.

PR #2097, opened 2020-06-16
It's Its, not It's

Nethaufer:
## About The Pull Request
Its is the possessive form of it, meaning 'of it'. ... Its, without an apostrophe, is a possessive form, where an apostrophe is usually required. It is similar to words like his and hers, neither of which needs an apostrophe.

## Why It's Good For The Game
Its is the possessive form of it, meaning 'of it'. ... Its, without an apostrophe, is a possessive form, where an apostrophe is usually required. It is similar to words like his and hers, neither of which needs an apostrophe.


AvaricePleonexia:
10/10.
These are the changes that we need to have, thank you Neth for doing your part. :heart:

PR #2123, opened 2020-06-22
Mining Love, otherwise known as "Hey you get guns now"

Enzo-Leon:
## About The Pull Request

Adds a mining weapons locker for station locked Phaser weapons. This will work the same as exploration weapons.

## Why It's Good For The Game

QOL and needed mining buff

## Changelog
:cl:
add: Weapons Cab for Mining
/:cl:

### DIFF
--- a/_maps/map_files/Tether/tether-01-surface1.dmm
+++ b/_maps/map_files/Tether/tether-01-surface1.dmm
@@ -1015,7 +1015,10 @@
 /obj/effect/floor_decal/corner/brown/border{
 	dir = 1
 	},
-/obj/structure/closet/secure_closet/miner,
+/obj/structure/closet/secure_closet/guncabinet/excursion{
+	name = "mining weaponry cabinet";
+	req_one_access = list(48)
+	},
 /turf/simulated/floor/tiled,
 /area/tether/surfacebase/mining_main/storage)
 "aca" = (
@@ -1113,14 +1116,13 @@
 /turf/simulated/floor/tiled,
 /area/tether/surfacebase/mining_main/eva)
 "ach" = (
-/obj/structure/table/rack,
-/obj/item/mining_scanner,
 /obj/effect/floor_decal/borderfloor{
 	dir = 5
 	},
 /obj/effect/floor_decal/corner/brown/border{
 	dir = 5
 	},
+/obj/structure/closet/secure_closet/miner,
 /turf/simulated/floor/tiled,
 /area/tether/surfacebase/mining_main/eva)
 "aci" = (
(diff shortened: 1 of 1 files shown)


lolman360:
can they get phase pistols and not the infinite recharge explo guns? There's nothing that phase weapons cant hurt in UD and they're useless for combat vs anything more complex than a simplemob (which is a good thing).

Enzo-Leon:
Good point lolman. Adjusting to a phase cab.

lolman360:
Plus, it's better that they have non-rechargeable weapons - Mining is fairly close to the station, and Explo having to run all the way back to the shutt;e every time they fight enemies is pretty detrimental. 

2. If you're giving them phase carbines right off the bat, maybe make the vendor give phase rifles or even like, phase grenades or something

Enzo-Leon:
I might depending on how the other changes that are going into place effect it. If it's too hard, then we can add those for them.

yourdoom9898:
This is, honestly, worthless to mining, when a PKA with triple damage is only 3900 mining good boy points, does a extra 26 damage, and has infinite ammo

Enzo-Leon:
KPA has limited ammo and a slow recharge rate on-top of a low fire-rate. Most of the miner players I've reached out to have expressed wish if we could adjust the KPA to what main TG has which performs a lot better.

Initial testing with the phase carbine last night showed the the guns operate on an equal part with the KPA but with a better fire-rate. The ammo limitation with batteries is to give the carbines more of a "back-up" when mining more than a hunting weapon. 

Mining is also getting buffed in terms of difficulty and providing loot that specifically benefits the station and not the single miner specifically (Increased ore spawns and outcrops around hostile areas, as well as upgraded gear science/engineering/medical could benefit from getting)

This is still in the very early testing phase and we'll easily pull back from the weapon portion since the mapping section is different from this.

yourdoom9898:
The KPA has quite literally limitless ammo, though it does have a ammo maximum, and as far as it is concerned, the PKA on main and RP are statistically similar, with the Main PKA doing 40 damage base vs RP's 32, but RP's KA has one more tile range. (I also am unsure of Main's PKA recharging on it's own)

Enzo-Leon:
The recharging on mains KPA is that you don't have an ammo reserve. You can fire the KPA as soon as it finishes recharging. Where as here you have the recharge delay, but... limited shot until it recharges another tick for you, and you can not replace the cell or have a quicker way of charging it currently. (I think a weapon re-charger might work on it though)

silicons:
phase weapons are shit but w/e

PR #2130, opened 2020-06-23
Re-allows Airtight Plastic Flap Production

Nethaufer:
## About The Pull Request
Re-adds the ability to make airtight plastic flaps from the plastic sheet menu.


lolman360:
these are really op actually

Nethaufer:
If they're abused we'll take care of them. Either that, or make a way to make functioning airlocks. The only reason why I noticed this was because I was trying to make a voxbox and saw that I couldn't make them like I normally did.
If you want I can boost up the plastic requirement more or something to keep people from spamming them all over the place, for some reason.

PR #2138, opened 2020-06-24
Don't Explode with Invalid Gastype

Nethaufer:


TheFurryFeline:
Active edge caused by a particular POI on V2, it seems. Else, I don't see a problem with this fix.

PR #2139, opened 2020-06-25
pteric flight bad

lolman360:
non-wing flight is good and possible. flying already isn't amazing (HUGE nutrition drain, slam into airlocks and walls) so this isn't that significant of a change


Nethaufer:
Why though

yourdoom9898:
Slight note, it is called Winged Flight in species creation over at code/modules/mob/living/carbon/human/species/station/traits_vr/positive.dm . Might want to fix that

lolman360:
@Nethaufer there's no invisible wings sprite so here we are

it opens up flight to characters who don't have wings. flight isn't really that good of an ability because it drains a shitload of nutrition so *shrug

nik707:
Honestly I'm okay with this because it allows for fun roleplay like synths with fucking rocket boots xd

PR #2143, opened 2020-06-26
changes endnotify to notify for consistency with main

silicons:
why did i make it endnotify instead of notify this is just annoying to have to memorize both


Zandario:
Won't this cause conflicts with main?

silicons:
nope

PR #2149, opened 2020-06-28
Adds firing pins and their associated behaviour. No more gunsci please.

BlackMajor:
This PR does the following.
-Adds firing pins to guns which control how and when they can fire.
-Defaults all guns to a standard firing pin, leaves a commented out block in designs.dm for removing the firing pin from protolathe printed guns (In my opinion we really should uncomment it, but the other maintainers disagree with me.)
-Adds Test range pin, explorer pin, dna pin, laser tag and loyalty pins with associated functions.
-Defaults explorer guns to exploration pins, same behaviour applies as before, sans emagging the guns. Gun emagging to unlock feature has been moved to all guns.
-Adds one box of 8 pins to RD locker, HoS locker. And two boxes to the armoury.
-Adds a 10credit supply pack of 3 standard firing pin boxes under armoury access.
-Fixes some misc bugs with guncode that I found while making this PR, namely examine().
-Ran the tether maps through sdmm to sanitize variables.
-Since we don't have table crafting like main, pins can be removed by inserting a new pin into the gun, destroying the old pin in the process. No getting free standard pins from a secvend.

Most of these changes won't really matter much unless we were to add more instances of guns without pins, which this PR would serve the basis for in future if we decide we want to start implementing gun control.


lolman360:
can you still swipe armory IDs on explo guns to allow them to work onstation

BlackMajor:
> can you still swipe armory IDs on explo guns to allow them to work onstation

The functionality has been moved to explorer firing pins, which explorer guns start with.
So yes, you can.

lolman360:
also literally just do
```
gun/attackby()
    if istype(obj/item/id && pin) //emags are ids
         pin.attackby(id)
    
```

BlackMajor:
Fuck you're right.

VailTheWolf:
My big concern about this is that right now Phaser and Phase weapons are basically hardcoded to be station bound and giving them a reason to potentially not be will put Security back into a "you can't have your guns even though you're going to UD right this moment" mentality. Let's not add Phasers to the mix on purpose, maybe.

BlackMajor:
> My big concern about this is that right now Phaser and Phase weapons are basically hardcoded to be station bound and giving them a reason to potentially not be will put Security back into a "you can't have your guns even though you're going to UD right this moment" mentality. Let's not add Phasers to the mix on purpose, maybe.

>Giving them a reason to potentially not be
Bro they already have the "scan armoury ID or emag to unlock on station" functionality, I just moved that to the explorer pins instead.
They're functionally identical, major difference is now you can explorer pin /other/ guns in order to station lock them.

VailTheWolf:
I didn't know they did that, never mind me. Had a "back in my day" moment.

PR #2155, opened 2020-07-01
Let there be Light

Nethaufer:
## About The Pull Request
Why does custodial get light replacers but not engineering.

## Why It's Good For The Game
Let me see

## Changelog
:cl:
add: Light replacer to electrical closets
/:cl:


BlackMajor:
Change light replacer count to 1 instead of 2 per electrical locker.
You don't need that many light replacers, considering the number of electrical lockers that exist.

Nethaufer:
You mean... three? There's not a whole lot of electrical lockers. 

But yeah I can drop it down to one regardless.

PR #2165, opened 2020-07-04
Adds ateopine and associated injectors + map changes

AnalWerewolf:
## About The Pull Request

Blind port of IS12 Warfare's Ateopine autoinjectors, redesigned and rebalanced for Citadel use.

## Why It's Good For The Game

A common complaint is that SaRs, and lesser so, parameds have a habit of 'powergaming' and 'carrying around medbay on their backs'. These complaints aren't necessarily true, a lot of this supposed powergaming is simply the necessary setup for SaRs prior to an expedition, of course one doesn't always have to prepare for an expedition but it is good practice in case a team capable arrives. 

So the problem here is the setup and amount of equipment needed, right. So how do we fix that?
We fix it by cutting down on the amount of unnecessary prepwork and running around.

To achieve this I will be gradually implementing new QoL changes and equipment for both medbay and SaR that hopefully push the meta away from chem grinding in chemistry for half an hour, then running around grabbing things for another half hour. First and foremost in this plan is to remove the need to carry a defib around on their backs or in their bags to be effective in long range recovery of wounded or deceased. Defibs aren't too common, they're massive and annoying to lug around, so I've introduced ateopine autoinjectors.

Ateopine autoinjectors are used to:
- Revive patients similarly to a defib shock about eight seconds after injection.
- Restore brain damage as a result of oxyloss or blood loss .
- Restore a SMALL amount of heart and lung damage, up to a maximum damage of 12, if damage is much higher than that, all they restore is 5 organ damage to each.
- Stabilize patients immediately for a short time.
- Enable SaRs and paramedics to quickly revive patients even through hardsuits or away from medbay.

However, being autoinjectors and not full-blown defibrillation machines, they do not advise you of who is able to be revived, when something prevents revivification, or prevent you from accidentally wasting an injector. It's a skill based reward sort of thing, if you can tell when someone is able to be revived, know when it is safe to do so, and keep them alive after injection, you will get a lot out of these.

image
You'll receive a message, as will the admins, on a fail, success, or in this case if they're revived but have no client (he's a spawned human).

## Changelog
:cl:
add: Ateopine and ateopine autoinjectors from IS12 Warfare
/:cl:

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
@@ -1279,7 +1279,6 @@
 
 /datum/reagent/ateopine/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	var/chem_effective = 1
-	var/mob/living/carbon/human/H = M
 	if(alien != IS_DIONA)
 		M.add_chemical_effect(CE_STABLE, 5)
 		M.adjustBrainLoss(-5 * removed * chem_effective)
@@ -1292,10 +1291,11 @@
 		if(prob(33))
 			H.Confuse(10)
 	if(ishuman(M))
+		var/mob/living/carbon/human/H = M
 		for(var/obj/item/organ/I in H.internal_organs)
 			if(I.robotic >= ORGAN_ROBOT || !(I.organ_tag in list(O_LUNGS, O_HEART)))
 				continue
-			if(I.damage > 15) // Fixes their heart and lungs about 5 points or until it hits 15.
+			if(I.damage > 10) // Fixes their heart and lungs about 5 points or until it hits 10.
 				I.damage = max(I.damage - 1 * removed * chem_effective, 0)
 				H.Confuse(5)
 
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
@@ -1266,6 +1266,39 @@
 		M.make_dizzy(5)
 		M.make_jittery(5)
 
+/datum/reagent/ateopine
+	name = "Ateopine"
+	id = "ateopine"
+	description = "Ateopine is a medicine used to treat bradycardia and asystole patients when injected into heart muscle, it is also used to treat nerve gas and insecticide poisoning. Commonly used by emergency teams to resuscitate patients."
+	taste_description = "astringency"
+	reagent_state = LIQUID
+	color = "#DA00FF"
+	metabolism = REM * 1.5
+	overdose = 10
+	scannable = 1
+
+/datum/reagent/ateopine/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
+	var/chem_effective = 1
+	var/mob/living/carbon/human/H = M
+	if(alien != IS_DIONA)
+		M.add_chemical_effect(CE_STABLE, 5)
+	if(alien == IS_SLIME)
+		chem_effective = 0.25
+		if(M.brainloss >= 10)
+			M.Weaken(5)
+		if(dose >= 5 && M.paralysis < 40)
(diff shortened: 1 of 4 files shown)


code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
```diff
@@ -1266,6 +1266,39 @@
...
+	taste_description = "astringency"
+	reagent_state = LIQUID
+	color = "#DA00FF"
+	metabolism = REM * 1.5
+	overdose = 10
+	scannable = 1
+
+/datum/reagent/ateopine/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
+	var/chem_effective = 1
+	var/mob/living/carbon/human/H = M
```
mattroks101, line 1282: This should go down below the `ishuman(M)`

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -182,6 +182,83 @@
...
+/obj/item/reagent_containers/hypospray/autoinjector/ateopine
+	name = "autoinjector (ateopine)"
+	desc = "A sealed autoinjector system designed for resuscitation of patients whose hearts have stopped."
+	icon_state = "purple"
+	flags = null
+	filled_reagents = list("ateopine" = 5)
+
+/obj/item/reagent_containers/hypospray/autoinjector/ateopine/do_injection(mob/living/carbon/human/H, mob/living/user)
+	. = ..()
+	var/error = can_resus(H)
```
mattroks101, line 194: If you can inject things other than humans then please make sure to explicitly check that what you injected was indeed a human.

```suggestion
    if(!istype(H)
        return
	var/error = can_resus(H)
	
```

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -182,6 +182,83 @@
...
+			O = H.internal_organs_by_name[organ_tag]
+			if(!O)
+				return TRUE
+			if(O.damage > O.max_damage)
+				return TRUE
+
+	return null
+
+/obj/item/reagent_containers/hypospray/autoinjector/ateopine/proc/make_alive(mob/living/carbon/human/M, mob/living/user)
+//	var/deadtime = world.time - M.timeofdeath
```
mattroks101, line 234: You should probably remove this.

code/modules/reagents/reagent_containers/hypospray.dm
```diff
@@ -182,6 +182,83 @@
...
+
+	M.timeofdeath = 0
+	M.stat = UNCONSCIOUS //Life() can bring them back to consciousness if it needs to.
+	M.failed_last_breath = 0 //So mobs that died of oxyloss don't revive and have perpetual out of breath.
+	M.reload_fullscreen()
+
+	M.emote("gasp")
+	M.Weaken(rand(10,25))
+	M.updatehealth()
+//	apply_brain_damage(M, deadtime) // Ateopine heals brain damage anyway.
```
mattroks101, line 249: Same here.

AnalWerewolf:
Thanks a bunch for the review @mattroks101, I appreciate the help!

VailTheWolf:
This is literally just a get out of jail free card that negates the need for surgery, the medigun, and stasis bags all in one trip. We don't need these. Field Medic's equipment room already has a chem machine to let aaaalll the SARs make the chems they want. (I have my issues with it, and as I've shared before you don't need much to keep the pace going.)

The Medigun has corspemend cartridges that we've already buffed, thanks Enzo. So we already have an easy fix for severe damage and easy revival.

silicons:
i suppose i'll say publicly that i'm against ateopine as it seems like main's strange reagent only without the limitations of it 

makes a death a bit too easy to fix, even with the given limitations.

porting main revival + reconstruction surgeries would be much better if we want to make resleeving less needed rather than making death itself less impactful via quick fixes.

nik707:
I'm also not a fan, with that given I think I'll close this for now and if more alterations are made to bring it more in line then we can handle it then

PR #2170, opened 2020-07-06
Fixes to Misc failing to load.

GLDoS:
"Runtime in code/modules/maps/tg/reader.dm, line 205: Undefined model key in DMM: tether_misc.dmm, mN"


BlackMajor:
Turns out missing variables aren't the cause of that runtime but hey, variable sanitzation is nice so this gets merged even if it didn't fix the listed runtime.

I should probably run all the maps through sdmm variable sanitzation sometime..

PR #2177, opened 2020-07-10
Readds cargo carp/spider

yourdoom9898:
## About The Pull Request

Readds the DIY Carpotoxin and DIY Spidertoxin crates lost in hardsync. Also buffs the meat output of base spider/carps.

Also thanks to @Hatterhat for the idea before hardsync, I'm just readding it

## Why It's Good For The Game

Hey, we have these advanced medchems that we can't really use. This lets us use them

## Changelog
:cl:
tweak: Readded Carp/Spider crates
balance: buffed base Carp/Spider output
/:cl:


lolman360:
the advanced med chems are meant to be locked behind an event so they aren't obtainable every round. guaranteed spider spawns in UT make spider crate not as much of an issue but free rezadone/osteodaxon from carp is a bit 😕

yourdoom9898:
There were already a thing for months, they just got wiped by hardsync. This also still requires people order them and kill the live carp, so it's not like it's for free.

PR #2178, opened 2020-07-11
subtler. anti. ghost.

silicons:
fine ill do it myself

adds a verb

invisible to ghosts

logged.


lolman360:
make it same tile only

PR #2183, opened 2020-07-12
adds some black magic that all of 4 coders on citadel total understand so i opt out of typing indicators to not lag off a cliff again i don't know how hrp players deal with this shit 

silicons:
this will allow for rebinding T and M without collisions happening with byond input handling so the user can hardbind T and M to say/me respectively without conflicts.


silicons:
oh epic rp handles it even more differently goddamnit i hate this place sometimes why is this so painful

silicons:
ok

PR #2189, opened 2020-07-15
New Sci Outpost

ThatOneTeshari:
Changed up the Toxins Outpost, giving it a new layout as well as a room for Telesci and a room for Particle Smashing so Science doesn't have to rip up Toxins every shift if they wanted to do stuff with it. This also comes with a break room, dorms, a place to wash up, as well as a first aid station. 

As well added areas for the new layout so things shouldn't go jank with the fire alarms. 

PR Photo


lolman360:
this just gives sci a PA/focus roundstart? there needs to be at least a little bit of interdepartmental cooperation to get it online, it's really quite powerful (i rebalanced morphium tech values based on the fact that it's not a guaranteed every round thing to get the PA set up).

i don't personally have a problem with telesci as we dont have antags but telesci is obscenely powerful still and probably should start in at least a partially completed state

ThatOneTeshari:
Heya, took a read at what you said and I'll see what I can do! Just realized giving Sci the PA right away sorta removes the need to talk with Cargo at all. Which is a good point to make about interdepartmental cooperation. "

Following this I will get those changes applied since it would be simple enough for people to assemble it all and test it as needed. So going to apply those changes! Also if you spot any issues, drop them in here. Already fixing a few. Like power and APC oddness.

ThatOneTeshari:
Following photo are the changes to the two areas:

Update

PR #2204, opened 2020-07-18
Sci post adjust/fix

ThatOneTeshari:
Got a report something wasn't working with Toxins - found out that the burn chamber lacked two pipes. 

Other changes: Slightly smaller toxin disposal room, plants in the middle area have been changed to fit better, (No grass/flowery stuff and more what is normally spawned on the surface.). Added a minor adjustment to the western hall.  As well adds a console screen so people don't need to run back and forth too much.

Hopefully the last but will update and fix as issues come to my knowledge.


ThatOneTeshari:
I'm extremely hopeful this is the last thing I need to correct. x.x'

PR #2235, opened 2020-07-29
travis test runs now force a specific rand_seed so we can maybe not have random travis test failures

silicons:
*eyeroll* no this doesn't fix the issue but it does get around it.


Zandario:
Stale and conflicts

PR #2238, opened 2020-07-30
[READY] Xenolove: Last Hotfix, Secmed Returns, Civilian and Mining Bodysuits too

JennaSilverShark:
Amend to list of suits, as simple a fix as it gets


JennaSilverShark:
Added EVA, Mining and Secmed bodysuits, also added the 'Security Physician' role

dreammaker_m02UWPBrwW
dreammaker_EXpOTeONGG
dreammaker_mhcvSj5gOx

PR #2248, opened 2020-08-06
what if rp players could actually vote on the headmin vote

silicons:


buffyuwu:
poggy woggy!!

VailTheWolf:
What if the headmin vote was over?

yourdoom9898:
Still waiting to cast my vote!

PR #2252, opened 2020-08-09
Colorable Vox

Captain277:
Help, help. I dropped my Vox in this vat of dye and now they're all colored.

1. _Converts Vox body sprites to grayscale and allows skin color customization._
**Why:** Vox with different color patterns could be neat.

2. _Adds previously two-toned areas on the Vox body to the Markings list as colorable options._
**Why:** Making the Vox body grayscale naturally removes two toned areas like the claws and beak. These can now be located under the markings list and equally colorized, if the Vox player chooses to have two color tones instead of one flat body color.

This doesn't mean Hot Pink Vox are canon. Use common sense.


lolman360:
finally, red vox

JennaSilverShark:
Ah sweet glad I'm not the only one who had this idea!

PR #2258, opened 2020-08-11
[READY] Xenolove: Zaddat Polish

JennaSilverShark:
dreammaker_QjTgROAQ9i

dreammaker_0K0V5n9hJQ

The Freefarer Shroud actually works properly now and the ancient Shroud helm no longer looks like a pill


code/modules/clothing/spacesuits/void/zaddat.dm
```diff
@@ -6,7 +6,7 @@
 	heat_protection = HEAD
 	body_parts_covered = HEAD|FACE|EYES
 	slowdown = 0.5
-	armor = list(melee = 0, bullet = 0, laser = 0, energy = 0, bomb = 0, bio = 100, rad = 70) //realistically would have some armor but balance.
+	armor = list(melee = 0, bullet = 0, laser = 0, energy = 0, bomb = 0, bio = 100, rad = 100) //hey look! some rad protection!
```
JennaSilverShark, line 9: If it were up to me it'd have a touch of armor too but it isn't

PR #2261, opened 2020-08-14
Ports Main Statues, and Sandbags(?)

Captain277:
1. _Adds Main's/tg's statues._
**Why:** I was shocked to find these weren't craftable last week. The code is questionable, but I think I successfully converted it.

2. _Adds Sandbags(?)._
**Why:** Implementing them now as part of a future potential Exploration rework. I spent all day on this, and I have literally no idea if it works. It probably won't. I'd like input on it either before or after it's merged. These are my first ports from Main's code.


Captain277:
I really do need another pair of eyes on this one.

PR #2285, opened 2020-08-23
hey what if we liked ore not taking 40 minutes to process

lolman360:
what #2265 should've been

altclick processor to make it go fast


BlackMajor:
Turning speed_process on/off doesn't actually affect its speed. The var is only checked during the machines Initialize on whether or not to add it to the speedprocess subsystem.

If anything I'd rather just have it turned back on, since turning it off was a mistake.

lolman360:
oh wack

PR #2289, opened 2020-08-27
TRIUMPH TEST MERGE. EXPLORATION READY. CARP DONT CRASH US ANYMORE. ALL DOCKERS CONTROLLERS FIXED

Enzo-Leon:
## About The Pull Request

E

## Why It's Good For The Game

eeeeeeeeeeeeeeeeee

## Changelog
:cl:
add: haha ship go vroom
/:cl:


Enzo-Leon:
Final notes
- Belt mining reverted back to level 1, but will instantly jump to level 5 and should not cause further issues when another scan is done. (Old set-up starting at level 3 might of been the issue)
- Overmap event tile spawning is still acting weird, but did one more adjusting to count numbers. Unsure if this will fix it. We'll need to keep tabs
- Mining planet is still lacking ores, and the error keeps popping up, I adjusted the rare value drop AGAIN to see if this resolves it and the error goes away. At the very least it should still keep populating with ores in the end.
- Cult event spawn is set to MAJOR only with a weight of 10 that'll ONLY increase with sec players, this is the same weight as the other major events currently on the roster. The actual spawn numbers are greatly reduced, and will no longer spawn in "public hallway" locations, or the kitchen. They've been given new areas to spawn that fit their theme, or are more prepared to handle a hostile attack as such.

Zandario:
Can't wait to get this out of my hair, holy fuck

PR #2314, opened 2020-09-10
Loadout Alphabetization, File Merging, and Cleanup

drofoljaelisglis:
## About The Pull Request

Compared to my last attempt at this, this has a much lower chance for anything to go wrong. This merges all the various loadout files that I could find from an immediate, though not deep or thorough, search for files that handled loadout for various items into one file. This may not have been necessary to achieve alphabetized sorting, but I think it cuts down on code bloat. There's much work to be done with getting the loadouts in a better state - but this is a good first step.

## Why It's Good For The Game

The current loadout display was nonsensical due to an oversight. It's been corrected, and now there is a significantly higher quality of life experience that comes with a structured order present in the way the loadouts are displayed.

## Changelog
:cl:
fix: Fixed the poorly sorted loadouts to sort properly.
/:cl:


code/modules/client/preference_setup/loadout/loadout_accessories.dm
```diff
@@ -287,8 +287,91 @@
...
+/datum/gear/accessory/webbing
+	cost = 1
+
+/datum/gear/accessory/stethoscope
+	allowed_roles = list("Chief Medical Officer","Medical Doctor","Chemist","Psychiatrist","Paramedic", "Field Medic")
+
+/datum/gear/accessory/khcrystal
+	display_name = "KH Life Crystal"
+	path = /obj/item/storage/box/khcrystal
+	description = "A small necklace device that will notify an offsite cloning facility should you expire after activating it."
```
Zandario, line 357: While you're editing these files can you please scratch out this instance of KHI <3
Don't forget to touch the box's define as well so it doesn't bork
```suggestion
/datum/gear/accessory/vmcrystal
	display_name = "VeyMed Life Crystal"
	path = /obj/item/storage/box/vmcrystal
	description = "A small necklace device that will notify an offsite cloning facility should you expire after activating it."
```

Zandario: Also scratch out the [/datum/gear/accessory/lifecrystal] as it's a bugged copy of this item

Zandario:
Once you do the other things we discussed, this PR is officially Poggers

PR #2320, opened 2020-09-14
Cyber Horror 

ThatOneTeshari:
Sprites from VGstation 13

Adds in a handful of new simple mobs that uses some sprites from it, they're mostly melee focused.

Mobs have varying amounts of armor and health, mostly to promote different weapon combinations and tactics, mobs also have different attacks. Mostly things being from mob types we already have, in terms of special attacks.

Will be keeping an eye out for suggestions/issues. 

Only sprite in the DMI that wasn't use was mouse_brown.

Horror PR


Zandario:
ooga booga gimmie da horror

PR #2322, opened 2020-09-14
19 lines of absolute shitcode to make telebatons only harm on harm intent without making them hardstun

silicons:
*groan

refactor when

30 pain force on nonlethal that should be good without being op

weaken is a hard no due to it being op.


code/game/objects/items/weapons/swords_axes_etc.dm
```diff
@@ -87,12 +89,22 @@
 			user.Weaken(3 * force)
 			if(ishuman(user))
 				var/mob/living/carbon/human/H = user
-				H.Weaken(5*force)
```
silicons, line 90: this is a 150 second stun

PR #2324, opened 2020-09-14
[READY] Xenolove: RIGs for Everyone

JennaSilverShark:
dreammaker_2MSxyLmeEA
dreammaker_WATlhaPqxP
Finally RIGs are available for all species across the universe! Special developments in habitation simulation for RIG suits have been made so that they can be suited on Zaddat and Phoronoids now, with all new looks! Just another step in making this a better working environment for all our lovely xenos!


JennaSilverShark:
Could probably use some extra code to have phoronoid RIGs deploy over their containment suits, but I have no idea how to do that.

PR #2332, opened 2020-09-17
reverts brig physician alt title

silicons:
## Why It's Good For The Game

i like multi jobbing as much as the next guy but it's pointless to point to rp and say fulfill your role not others' roles when you have a title that is just "security officer but they can point to themselves and call themselves a doctor". it makes zero sense, if we wanted it we could have made another role for it, a special slot-limited role rather than just officer but they're called a doctor. no point in this, adds confusion and makes no ooc sense from a design perspective.

## Changelog
:cl:
del: alt title removed: brig physician
/:cl:


VailTheWolf:
Duplicate of #2333

Zandario:
Duplicate of merged #2333 
The suits still being in could still encourage the behavior

PR #2348, opened 2020-09-19
We're not that horny. Drastically reduce compliance disk spawn rate.

BlackMajor:
Webedit PR.

Bro there's like 20 of these things in maint in every round you don't need that many for the one or two people that have a fetish for it.

Also cleaned up the loot list. Reeks of downstream, which we're not anymore.


yourdoom9898:
Aww, the bodysnatcher is kinda cool

BlackMajor:
> Aww, the bodysnatcher is kinda cool

My bad, deleting those lines was accidental when I was just removing the comments.

PR #2352, opened 2020-09-19
Firearm Recoil (but funny)

Hatterhat:
## About The Pull Request
Recoil - (of an action) have an adverse reactive effect on (the originator).
Adds a "shootback" var to guns, which, when enabled, makes every shot fired shoot the user in either the chest or the head.
Mercenaries drop weapons which have a 60% chance to have shootback enabled, except the Grenadier, whose shotgun is always shootback.
## Why It's Good For The Game
Heho funny.
## Changelog
:cl:
add: Reports of mercenary outfits across the sector outfitting their arms with advanced recoil systems have increased.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -106,6 +106,7 @@
 	var/obj/item/firing_pin/pin = /obj/item/firing_pin
 	var/no_pin_required = 0
 	var/shootback = FALSE
+	var/fix_attempts = 0
 
 /obj/item/gun/CtrlClick(mob/user)
 	if(can_flashlight && ishuman(user) && src.loc == usr && !user.incapacitated(INCAPACITATION_ALL))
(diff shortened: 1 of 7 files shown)


lolman360:
finally, protean only weapons

CharlesWedge:
I like this actually Good fix for merc weapon powercreep

code/modules/projectiles/gun.dm
```diff
@@ -203,6 +205,20 @@
...
 		return 0
+	if(shootback)
+		var/obj/P = consume_next_projectile()
+		var/targetedlimb = pick(BP_HEAD, BP_TORSO)
+		if(P)
+			if(process_projectile(P, user, user, targetedlimb))
+				to_chat(user, "<span class='warning'>UNAUTHORIZED USER DETECTED.</span>")
+				user.visible_message(
+					"<span class='danger'>\The [src] buzzes, firing by itself into [user]'s [targetedlimb]!</span>",
+					"<span class='danger'>\The [src] buzzes, and as your grip loosens on it, it fires point-blank into your [targetedlimb]!</span>"
```
chaoko99, line 216: `you lose your grip as the [src] buzzes, firing straight into your [targetedlimb]!` flows a bit better.

chaoko99 (changes requested):
Grammar.

VailTheWolf (commented):
I'm against this entirely. We don't need a sub variety of weapons that inexplicably make you 180 your point of aim and make you shoot yourself in the face because of "recoil". All of these weapons have/are in semi automatic firing mode by default, and with the exception of the shotgun don't have meaningful recoil in the first place. 

Completely unnecessary and fruitless PR. Exploration already carries Multi-tools, and usually full tool belts or switch tools way more than they need to. This will only encourage that behavior.

code/modules/projectiles/gun.dm
Hatterhat: thank you for your input

CharlesWedge:
Well Perhaps we can reflavor it as Merc Guns being ID locked so when a none-merc tries to fire them it blows up in their face. I imagine it would be a common feature to keep the guns out of the hands of pirates that may get lucky and kill a merc or three.

Hatterhat:
it's my comedy PR and i decide what the lore is
and the lore is "advanced recoil system" because funny

Zandario:
Meme PRs are a nono.

PR #2366, opened 2020-09-23
makes r&d weapons start firingpin-less

lolman360:
## About The Pull Request

makes all protolathe weapons start without a pin

## Why It's Good For The Game

gunsci is becoming an issue

paging @BlackMajor @Enzo-Leon @VailTheWolf  for input

## Changelog
:cl:
balance: R&D guns no longer start with firing pins.
/:cl:


VailTheWolf:
Petitioning for this PR to be merged with an SOP change. 
R&D needs to issue out weapons to Exploration with the appropriate Station-lock pin installed. 
Such a standard should only be overruled by the HoS or Facility Director in times of great emergency i.e. a Confirmed Code Red threat on-board the Station/Ship.

We have the population. We have the security players. We don't need constant Exploration Militia for every boo boo. (And this is coming from a Field Medic main.)

lolman360:
personally agree with the above point, it was fine back on tether with the mediocre frontier phasers but more often than not you'll see explo with h90ks or such now

BlackMajor:
I said it when I made the PR to the other maintainers that it'd be a problem, but uhh. Majority vote was to leave it in since "It's not a problem yet."

Since I wanted it in the first place, for obvious reasons I'm already in favour.

Zandario:
I mean I don't see a problem with this either.

silicons (commented):
ensure armory has 2 boxes (or maybe even more depending) and the rd has a box, of electronic pins, make sure cargo can order electronic pins at a cheap cost
make sure pins and guns with pins can be emagged to fully break the pins to be electronic.
also tempted to say remove dna pins but it's whatever atm, dna pins are pretty bullshit imo but i don't particularly care until antagonism is a thing.

BlackMajor:
> ensure armory has 2 boxes (or maybe even more depending) and the rd has a box, of electronic pins, make sure cargo can order electronic pins at a cheap cost
> make sure pins and guns with pins can be emagged to fully break the pins to be electronic.
> also tempted to say remove dna pins but it's whatever atm, dna pins are pretty bullshit imo but i don't particularly care until antagonism is a thing.

Cargo can already order pins.
Pins and pinned guns can be emagged to ignore the pin restrictions already.
DNA pins were left printable in science because it didn't matter seeing as every gun printed started with a pin, now that they're printing without pins I'd appreciate if the dna pin design was removed from RnD.

PR #2385, opened 2020-09-28
Suit cycler improvements

BlackMajor:
Couple of bugs left to squash in this but it's safe to test merge this right now.

This PR just updates the suit cyclers code to be a little more reliable, as well as adding an option to all suit cyclers to just... not change the style of a void suit. Because sometimes you need to refit that sweet flaming suit you found but can't without also changing the aesthetic.

I'd appreciate having this testmerged incase I scuffed anything relating to the numerous species and suit styles.


BlackMajor:
I made this PR while drinking alcohol in the snowy mountains on vacation using my phone. I just want to remember that when I wake up tomorrow.

code/game/machinery/suit_storage_unit.dm
```diff
@@ -616,37 +616,38 @@
...
 	can_repair = 1
 
 /obj/machinery/suit_cycler/exploration
 	name = "Explorer suit cycler"
 	model_text = "Exploration"
-	departments = list("Exploration","Old Exploration")
+	req_access = list(access_explorer) //Old Exploration needs fixing up
+	departments = list("Exploration", "No Change")
 
 /obj/machinery/suit_cycler/exploration/Initialize()
```
Zandario, line 652: Exploration Voidsuits exist for Teshari, seems this was the culprit preventing them from getting them. Verified this ingame via varediting the cycler. Only problem being the obj sprites aren't showing, but that should be simple enough. Though could be outside the scope of what you wanted.

code/game/machinery/suit_storage_unit.dm
```diff
@@ -655,21 +656,22 @@
 /obj/machinery/suit_cycler/pilot
 	name = "Pilot suit cycler"
 	model_text = "Pilot"
-	departments = list("Pilot Blue","Pilot")
+	req_access = list(access_pilot)
+	departments = list("Pilot", "No Change") //Pilot Blue needs fixing up
```
Zandario, line 660: Ditto

PR #2399, opened 2020-09-30
Removes floral somatoray pin requirement

ThatOneTeshari:
## About The Pull Request

Makes it so that Xenobotanists  can use it for Xenobotany without having to go to Sec or asking their RD for a pin.

Adds one line of code to ensure that the floral somatoray is usable on station for Xenobotany.

## Why It's Good For The Game

Allows Xenobotanists to actually use their equipment rather than having to take over the firing range for their projects. 

## Changelog
:cl:
adds: Removes the floral somatoray's pin requirement to be usable on station. 
/:cl:


ThatOneTeshari:
So reason this got spurred on was because someone sent me a screenshot of: 

HUEHUE

Yes, that is Science's Firing range being used for Xenobotany.

PR #2421, opened 2020-10-08
MAYBE fixes superfast rig AI movement but is just more snowflake and the entire rigcode forced movement system needs to just legitimately get nuked from the code at some point and rewritten

silicons:
yeah this is bad


lolman360:
is the issue only limited to protean rigs or

PR #2427, opened 2020-10-11
Hard Boiled: A Citadel Story (Gives the Detective their .45 instead of a taser.)

Captain277:
_The rain pattered away at the foggy glass window of my office, a rolling snare drum. Part of me wondered how it could be raining indoors, on a space ship, but most of me just wanted a stiff drink. I'd been on this damn case for years now. I felt the time lost on this job slipping through my calloused, cracked hands. It felt like sand pouring over the top of a silk nightgown. There had been a woman in that gown, when I was young. A .38 revolver, standard issue. A good gun. A reliable gun. My gun. The hallway lights cast a harsh shadow across my desk, the rain-stained fingerprints piercing the slats of my Venetian blinds, and presenting me with the truth. They'd stolen my gun and snuck a taser in its place. A taser - the weapon of the new generation. Hotshots. Techno junkies. Cyber freaks. All too eager to abandon the old ways of non-lethally taking down a man. Well, I wasn't gonna take it any more. My hands flew to the rolodex on my desk. I was searching for a name. A forbidden name I'd been told to never utter again. But here I was, with that name on my lips and a PDA in my hand. I was going to find that gun, and I was going to get her back. In the hallway outside, lightning flashed. How was it raining inside, seriously?_

1. _Removes the taser from the Detective's locker, and returns their custom .45._
**Why:** The Detective's supposed to have soul, and flavor, and a drinking problem. Here's something that provides two of those things, and may lead to the third.

2. _Adds a Medivend to the arrivals lounge._
**Why:** We've had some issues with bodies pilling up in Arrivals. This medivend seems like a prudent way to assist doctors in providing on-site care.

**Edit**
3. _Moves a pair of Brig holding airlocks two (three?) tiles to the right, where they're clearly meant to be._
**Why:** These airlocks have been oddly shifted to the left, even though this leaves one cell exposed. A pair of nearby fire shutters implies they were mistakenly placed here. This corrects the issue.

4. _Adjusts the access to the firing range and adds .45 practice speed loaders._
**Why:** Apparently Detectives can't open the firing range. Corrected.

**Double Edit**
5. _Adds a mass driver to the Chapel/Civilian EVA_
**Why:** Someone requested it, and we may as well have one if we're gonna stock coffins. It's added in here because this PR already edits the map.

6. _"Fixes" Acid and Blood turfs._
**Why:** Reverts Acid and Blood turfs to their original code. Known issues regarding splash messages and icon overlay overrides remain, but are minor.


lolman360:
.45s are 5 more brute on normal ammo and do the same amount of agony, I don't see a problem with giving them .45 gunbox wholesale at this point.

PR #2440, opened 2020-10-15
Xenohybrid Racial Adjustments

Captain277:
For some reason, Xenohybrids - literal Xenomorphs with either a genetic or technological separation from the hivemind - don't have basic Xeno traits. I've given them _some_.

1. _Adjusts Xenohybrid racial traits: Xenohybrids are now fully spaceproof, but still require Oxygen. They now take double fire damage, and half brute damage._
**Why:** Hybrids have been brought more in line with standard Xenomorph biology with this update, with small adjustments for balance.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station.dm
@@ -949,6 +949,17 @@
 	hunger_factor = 0.1 //In exchange, they get hungry a tad faster.
 	num_alternate_languages = 2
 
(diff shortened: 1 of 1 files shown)


lolman360:
double burn is really, really sharp for 0.5 brutemod

maybe adjust it to like 1.4, 1.5?

KasparoVy:
Burn baby burn!

Captain277:
Well, it's what the base Xenomorphs are statted for in the code. Xenos are typically portrayed as being pretty weak to fire, and Hybrids are still biologically Xenos.

KasparoVy:
I dig it man. I'm totally OK with burning, I take major burn on every build.

PR #2441, opened 2020-10-15
Adds automatic TGS DMAPI updates

Cyberboss:
Runs every UTC midnight, makes a PR if there's an update

Sample: https://github.com/Cyberboss/tgstation/pull/15


Zandario:
Thanks bruh, many thanks

PR #2447, opened 2020-10-18
Airlock Map Helpers are nice

Zandario:
## About The Pull Request
Airlock mapping helper from VOREStation/VOREStation#7072
## Changelog
:cl:
add: Airlock Helpers
imageadd: tweaked some of the icons from the port
imageadd: directional sprites for airlock components, like buttons, sensors, etc.
/:cl:


Zandario:
Got all of Triumph, outside of Deck1 due to Enzo mapping on that level at the same time, converted to use the helpers.

PR #2448, opened 2020-10-20
Triumph Update: New Engineering Bay & Reactor

Enzo-Leon:
## About The Pull Request

A MUCH needed re-work to the engineering bay was implemented finally w/ updates to the reactor and layout to atmos and engineering entirely. More maintenance was added. Also engines for movement were updated to fuel.

## Why It's Good For The Game

QOL and community feedback.

## Changelog
:cl:
add: Engineering floor re-work
add: Reactor change
add: engine changes.
/:cl:


Kraseo:
So I've got a lot of concerns with this engineering rework, primarily how there are so many design choices that are questionable at best, but also how this feels like a complete kneejerk reaction regarding how easy it is to overcharge the power grid with a few modifications to the supermatter.

I'll start off with the hallway, which is... way, *way* too long. I know deck 1 is the dedicated engineering department, but this feels like too much. The security checkpoint is also missing entirely as well, and it gives anyone who walks into the deck via the elevator easy access to the reactor room (which is now a fission core rather than a supermatter, which makes it arguably far less safer to be tampered with.)

imagen
imagen
imagen

(Notice the first image, where the smaller hallways splitting off from the crossroads lack any sort of light fixtures, and everywhere else but the section of the hallway near the CE's office lacks vents and scrubbers; with light fixtures peppered all over and lit up like a Christmas tree for a department only two or three people will be meandering around in.)

Then there's... this. Why is this huge box a thing? There's no decoration or anything along the way, just four rooms connecting eachother and wasting space. It would make sense if it was housing all the "extra" things stored at the corner of the storage room, rather than just jutting out in an ugly way.

imagen
imagen

Maintenance. Why are there borosilicate windows here, as opposed to the typical reinforced windows used everywhere else? They're not protecting anything vital. There's also the new windows connecting to the chief engineer's office, which don't really seem to serve much of a purpose either, and just make breaking in easier.

imagen
imagen

I don't understand the point of adding extra maintenance that doesn't really house anything else of note, like separate atmospherics, or even rooms out of sight? (The only places you can actually do extra stuff in are all in deck 2, and even then it might be too public for purposes like making another engine for extra power (more on that later.))

imagen

This is less actual criticism and more me nitpicking about the little things that annoy me, but: more empty rooms that don't seem to serve a purpose, and these doors here that lead to the tube that shoots out the fission core, which just seem weird to me that they're just there.

imagen
imagen
imagen

Moving on to the engine itself, I feel like switching out the supermatter for a fission reactor feels like it's been poorly thought of. While the supermatter is easy to "cheese" (that's more of a code issue than a playerbase issue, since there's literally nothing fun you can do with RP's supermatter compared to what main offers), the fission reactor requires lots of micromanaging, micromanaging which what little engineer players we have would probably come to despise. As an example, with this setup someone else made, here's power readouts from just after the engine was set up:

imagen

imagen
imagen

And then after an hour:

imagen
imagen
imagen

This PR isn't a quality of life change to engineering, because by forcing engineers to micromanage an engine that gives just enough power to supply the whole station moving with shields up, you're depriving them of being able to do anything else (so what's the point of going to the RP server, having to stare at the engine for up to 5 hours straight instead of interacting with anyone else?) unless they make an alternative engine to supplement or replace this one, and as I've said before, there's no space for it anywhere on the ship that isn't super public (while we have no antagonists, it makes no sense to set up something in those rooms that could be easily sabotaged and put the entire station in a blackout with little effort). I'd say it needs to be put either entirely on hold and revisited or scrapped entirely, because in my opinion this just feels like it wasn't thought on very well.

silicons:
not a fan of supermatter swapout.
the solution should be to make atmospherics/supermatter more interesting rather than go to an engine without nearly as much room for dynamics and expansion (there's simply no expansion room for fission that makes sense while supermatter can gain new interactions if the future if needed)

Enzo-Leon:
All points have been addressed properly in the first response (aside from hallway spacing, but that's more of the ships length layout. Nit-picks and issues with the reactor have been addressed and the original creator of the reactor has improved it to remove the micromanaging issue we caught after the initial transfer away from SM.

After this PR is pushed the next update for engineering is adding engine rotation to switch between RUS-T, Fission, SM. This will take a while but will be coming soon (not tm).

Zandario (approved):

PR #2457, opened 2020-10-22
Xenolove: One Size Fits All

JennaSilverShark:
In support of the new Protean RIG suits, I've sprited custom appearances so that every species can wear our lovely little nanites that already could anyhow.


JennaSilverShark:
Merge conflicts are fucking annoying

PR #2458, opened 2020-10-24
Xenolove: One Size Fits All

JennaSilverShark:
Adds all the sprites needed for protean RIGs to be worn on any species that can.


JennaSilverShark:
I'm crying and screaming right now

JennaSilverShark:
This is the cleanest file addition I have ever done, I have absolutely no way of fixing it right now.

Zandario:
I mean Protean rigs aren't even merged, so this wouldn't do much

JennaSilverShark:
Well the idea was that files could be there so the original pr would just have to code them in

PR #2475, opened 2020-11-08
Fixes soulcatcher mobs not being able to understand AIs 

silicons:
## About The Pull Request

Yes it removes the entire block. It made **zero sense.** If you wanted to prevent brains from hearing outside of MMIs it should just apply in general, not to only silicons. I'm not sure what this is for but I'm 99% sure it isn't needed so correct me if I'm wrong. Silicons aren't exactly a special scrambled voice that you can't understand unless you're human.

## Changelog
:cl:
fix: brainmobs can understand AIs without a snowflake MMI check.
/:cl:


lolman360:
can i put this on simplemobs too

PR #2486, opened 2020-11-12
MAKES SPACE NOT ROOM TEMPERATURE

silicons:
FUCK I AM AN IDIOT

### DIFF
--- a/code/game/turfs/space/space.dm
+++ b/code/game/turfs/space/space.dm
@@ -7,6 +7,7 @@
 
 	initial_gas_mix = GAS_STRING_VACUUM
 	thermal_conductivity = OPEN_HEAT_TRANSFER_COEFFICIENT
+	temperature = 2.7
 	can_build_into_floor = TRUE
 	var/keep_sprite = FALSE
 	var/edge = 0		// If we're an edge


KasparoVy:
I just thought the space-weather was nice :(

ForFoxSake:
Bruh, do you even space?

I don't wanna dump a lecture on how heat works, but I'm pretty sure this ain't what you're looking for.
Basically, the less gas is in an area, the less temperature can equalize between two objects.
Space turfs have a gas content of 0mol, ergo, temperature cannot equalize at all in this hard vacuum. (Technically: 293K * 0J/K = 0J, 2.7K * 0J/K = 0J)

If this changes the behavior of anything, then wherever that change is felt, is for sure something that's way more broken than just it's numeric value.

KasparoVy:
I think mechanically we're just trying to maintain our stance on space hurting people when they get out there and afaik I've always seen it done with a combination of low pressure + cold dmg

Not sure how we'd maintain that same level of lethality if we didn't give a fug about the temperature, or what temperatures we'd even simulate at that point

yourdoom9898:
Without this. it makes the cold loop for engine cooling effectively not function due to it not actually cooling the cold loop under 15 Celcius

ForFoxSake:
@KasparoVy For consistency's sake, space should not be dealing cold damage to mobs, but rather the mobs should be dealing heat damage to themselves while in low Jules per Kelvin environments.
This is already the case for FBP mobs, and is the primary function of the "suit cooling unit".
Having skimmed through life.dm to check for heat code it appears this is how it functions, as body temperature adjusts itself based on a ratio of gas density. It misses some values, but works in this scenario, as a pressure of 0 still produces 0 heat transfer.

@yourdoom9898 So. I coded the current engine that's being run on the Triumph. Nobody's told me it's not working, and I've tested it personally fairly extensively.
It seems to me that you're referring to something involving the radiator pipes. Now, their code is a mess, and definitely doesn't work as intended... _But_ in an ideal situation you'd already have seen a harsh drop in cooling power from them long before reaching 15C.
If this change is genuinely intended for the engine, then it's adding an additional 300K buff on top of the 1000K buff I've recently pushed to it, is it really in need of that much of a buff?

So, to repeat what my last comment was saying: The intentions of the current atmospherics system (and real world thermodynamics) suggest this change should not have any meaningful impact.

buffyuwu:
> FUCK I AM AN IDIOT

no, you're adorable pls be nice to yourself sweetheart

PR #2495, opened 2020-11-18
Minor Tweaks and Updates - November Edition

Captain277:
1. _Makes bucklers fit in the belt slot._
**Why:** Someone requested bucklers be able to fit on belts, and it seems thematically appropriate.

2. _Gives PARA their own Spawn Equipment listing._
**Why:** This will make bussing PARA class ERTs in much faster. Remember to set their minds to Holy still, as this is not handled by default.

3. _Removes Exploration Shields._
**Why:** This is a major balance change as my Exploration development continues. There is a reliance on these shields that makes most standard mobs trivial to deal with, leading to escalation, blah blah, removing for balance concerns.

4. _Fixes Mint Julep icon state._
**Why:** An underscore in the icon name broke its pathing. Small fix.

5. _Sabre Sheath now accepts stun swords._
**Why:** The Captain's sabre sheath can now hold a stun sword, instead of the sabre. Requested by a player, but also thematically appropriate.


silicons:
i'd actually recommend leaving exploration with one singular but very weak (like shouldn't resist more than 2-3 laser hits imo, maybe a subtype) shield for the pathfinder given the current state of exploration balance for defense purposes, and instead focusing on lessening the amount of focus on heavy weaponery it seems to have.

regardless i will merge this in a few days if there's no response

PR #2504, opened 2020-11-20
Intern Reassignment Initiative

Captain277:
Interns are no longer a class of Assistant. Instead, Intern titles are now sub-titles of the related base department job. You can still RP being an Intern/Trainee, but now you don't have to deal with guest passes and janky shit.

1. _Removes the Intern subset of Assistant, and redistributes its titles._
**Why:** QOL for "Intern" RPers.


KasparoVy:
Nice - good idea

Greenjoe12345:
uhh I think this breaks Server, the intern role of bartender/chef
since I don't see that anywhere

PR #2505, opened 2020-11-20
fixes simple_mob AI

buffyuwu:
readjusts existing AI code to be far more responsive when player is leaving a mob. mobs that have never been possessed by a player will not experience these increased response changes, as it only occurs between the STANCE_IDLE and STANCE_DISABLED.

this will allow an admin/player to possess an AI, use it/move it, and then ghost out and let it take over immediately

## Changelog
:cl:
fix: fixes simple_mob ai so that it doesnt freeze after a player leaves the mob
/:cl:


Captain277:
Travis is reporting an error with a map named "test" being included in the build.

PR #2515, opened 2020-11-24
this should not have compiled

buffyuwu:
fixes a terrible thing


Hatterhat:
arent mobtypes a bitflag? also does this even check for mobtypes to deal bonus damage

PR #2517, opened 2020-11-25
hair shapeshifting no longer balds you on cancel

silicons:
QOL


Captain277:
Travis is detecting active map edges.

PR #2520, opened 2020-11-25
Adjusts teshari balance

GrayRachnid:
## About The Pull Request
Reduces teshari speed and gives them a boost in health. This will drop their normal speed from 10 to 6.66 which is equivalent to normal tg/bee runspeed. Gave them a health boost so they can adjust.

## Why It's Good For The Game
With the code overhaul of movement speed and the fact that you no longer need less than 200 ping to properly play the game, teshari go a little too fast. This tones down their speed slightly while giving them a health boost to keep their speed more healthy for gameplay.

## Changelog
:cl:
balance: Reduced teshari speed but gave them a health buff to improve gameplay.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/teshari.dm
+++ b/code/modules/mob/living/carbon/human/species/station/teshari.dm
@@ -46,10 +46,10 @@
 
 	fire_icon_state = "generic" // Humanoid is too big for them and spriting a new one is really annoying.
 
-	slowdown	= -1
-	snow_movement = -2		// Ignores light snow
+	slowdown	= -0.5
+	snow_movement = -1		// Ignores light snow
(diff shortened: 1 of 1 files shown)


silicons:
before there's a shitstorm i'm going to say right now i fully support this pr 
a species that runs 10 tiles per second by default on a hrp server when even servers like /tg/ realized this was an insane speed is not kosher.
if they become too weak they can be rebalanced in other ways  
6.66 speed is pretty alright.

silicons (approved):

silicons:
(for reference default runspeed on rp is 5, so you're already going 30% faster)

KasparoVy:
Fuck.

At least we get a meme number.

PR #2527, opened 2020-11-26
buffs teshari brute/burn damage resistaances to 10% vuln instead of 35%, and item slowdown mod to 1.25 from 2

silicons:
## Why It's Good For The Game

i missed item slowdown mod in the original pr and now they move like a snail in a hardsuit which is too hard of a nerf
plus 1.1x mod with 75 health is already really weak, we can tune it back again if it turns out this is bad

## Changelog
:cl:
balance: teshari have been given a bit less damage vulnerability and item slowdown mod.
/:cl:


KasparoVy:
Teshari galaxy-wide experiencing a sudden surge in robustness
Xenophysicians confirm it is indeed because they finally figured out how to take vitamins

PR #2534, opened 2020-11-30
puts an explorer/offstation firing pin in the pathfinder's m41a instead of an unrestricted one

silicons:
## Why It's Good For The Game

>all of the other guns are locked but not the funny burst rifle gun they shouldn't even have

pog

## Changelog
:cl:
fix: pathfinder's m41a is now explorer-pinned.
/:cl:


Hatterhat:
theres a crate that comes with a single Oversized Burst Pistol, you might want to add an explorer pinned subtype and replace the one in the locker with it

Captain277:
This PR is shortly going to be made irrelevant. I'm removing the m41a from the Pathfinder's closet and replacing it with something more reasonable and less of a meme. I just keep forgetting to do that.

silicons:
@Captain277 kk

PR #2538, opened 2020-11-30
Removes the m41a

Captain277:
After forgetting to actually do this for a while, I've finally remembered.

1. _Removes the m41a from the Pathfinder's locker._
**Why:** This gun should never have been put in. In the interests of giving the Pathfinder a necessary and versatile ballistics alternative, I have instead given them one p90.


dwinters99:
This doesn't seem necessary considering how poorly the rest of the station is balanced vis a vis firepower. Also, the P90 is next to useless. Not a good trade-off. Welcome back to the phaser meta.

yourdoom9898:
@dwinters99 You can order it from cargo, not that you really need a powercreeped AR for explo

dwinters99:
Yes, I mean- explo is perfectly possible with just phasers. In fact, it's EASIER with phasers because you don't need ammo. Not to mention the randomness of ballistics. The matter is, though, that ballistics are _fun_.

I would understand this change if explo was used for anything that wasn't the fun of the players or if someone was abusing the weapon (how?).

With the exploration firing pin change that was added recently it won't be used on-station unless it's emagged (or a firing pin is mooched from the RD). Additionally, like I mentioned, the general balance for exploration here is SO POOR that I don't see why this is the fix that is deemed needed. 

Like, exploration gets their own medic rig? Easy-to-access syndie rigs? Fully equipped? Mercs have a 100% weapon droprate? Of all the things to be fixed, removing something fun and replacing it with a weapon that is next to useless seems very very backwards vis a vis the quest for fun.

code/game/objects/structures/crates_lockers/closets/secure/explorer.dm
```diff
@@ -212,10 +212,10 @@
 		/obj/item/material/knife/tacknife/survival,
 		/obj/item/material/knife/machete/deluxe,
 		/obj/item/gun/energy/frontier/locked/carbine,
-		/obj/item/gun/projectile/caseless/usmc,
-		/obj/item/ammo_magazine/m10x24mm/small/snark,
-		/obj/item/ammo_magazine/m10x24mm/small/hp/snark,
-		/obj/item/ammo_magazine/m10x24mm/small/ap/snark,
+		/obj/item/gun/projectile/automatic/p90,
+		/obj/item/ammo_magazine/m9mmt,
```
VailTheWolf, line 216: This is the wrong magazine. Just use /obj/item/ammo_magazine/m9mmp90 or /obj/item/ammo_magazine/m9mmp90/hunter.

Captain277: According to the p90's code, it can apparently accept the m9mmt magazine. The m9mmp90 base mag is described as AP by default, so I included the m9mmt as a non-AP/non-Hunter alternative.

VailTheWolf: That's extremely odd and not at all how that should behave. I'll see about changing that in the future if possible. A WT-550 magazine and an H90K (P90) shouldn't be compatible.

PR #2539, opened 2020-12-01
fixes protean vore settings because if i get hardstunned by a carp again i'm going to go ape and punch a hole through the code 

silicons:
qol

why is mobvore defaulting to on again


lolman360:
oh so that's why my mobvore keeps breaking

yourdoom9898:
Too bad you'll still get hardstunend with mobvore off, because it doesn't stop the pounce. Having mobvore off is honestly _more_ lethal

PR #2544, opened 2020-12-03
Speech bubbles appear above objects you're inside of.

BlackMajor:
See title.

Ported from polaris again, makes speech bubbles appear above containers if you're inside of one.

### DIFF
--- a/code/modules/mob/living/say.dm
+++ b/code/modules/mob/living/say.dm
@@ -338,6 +338,17 @@ proc/get_radio_key_from_channel(var/channel)
 	//var/image/speech_bubble = image('icons/mob/talk_vr.dmi',src,"h[speech_bubble_test]") //VOREStation Edit. Commented this out in case we need to reenable.
 	var/speech_type = speech_bubble_appearance()
 	var/image/speech_bubble = image('icons/mob/talk_vr.dmi',src,"[speech_type][speech_bubble_test]") //VOREStation Edit - talk_vr.dmi instead of talk.dmi for right-side icons
+	var/sb_alpha = 255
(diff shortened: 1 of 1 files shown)


KasparoVy:
Nice.

code/modules/mob/living/say.dm
```diff
@@ -338,6 +338,17 @@ proc/get_radio_key_from_channel(var/channel)
 	//var/image/speech_bubble = image('icons/mob/talk_vr.dmi',src,"h[speech_bubble_test]") //VOREStation Edit. Commented this out in case we need to reenable.
 	var/speech_type = speech_bubble_appearance()
 	var/image/speech_bubble = image('icons/mob/talk_vr.dmi',src,"[speech_type][speech_bubble_test]") //VOREStation Edit - talk_vr.dmi instead of talk.dmi for right-side icons
```
KasparoVy, line 340: D'you figure we can ditch the VR edit at this point?

code/modules/mob/living/say.dm
```diff
@@ -338,6 +338,17 @@ proc/get_radio_key_from_channel(var/channel)
 	//var/image/speech_bubble = image('icons/mob/talk_vr.dmi',src,"h[speech_bubble_test]") //VOREStation Edit. Commented this out in case we need to reenable.
```
KasparoVy, line 338: Ditto, is this commented comment still necessary?

code/modules/mob/living/say.dm
BlackMajor: Probably, but that's not part of what I changed.

PR #2561, opened 2020-12-07
small cleanup of visible message

silicons:
might fix ghost points


Captain277:
Lifted testmerge. No confirmation, but I believe this PR may have been related to the "That's a Citadel Station 13, how can you see this?" message that was appearing upon joining the round. Will be testing more later.

PR #2566, opened 2020-12-07
let's make vampirism not bad

buffyuwu:
🆑 
add: adds new functionality to vampirism, including more immersive text feedback for parties involved
add: victims who have been bitten will now appear to have been bitten in their examine until treated
fix: changes the 'fat and bloated' examine text after feeding to be more appropriate
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station_special_abilities.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station_special_abilities.dm
@@ -383,19 +383,19 @@
 		return
 
 	last_special = world.time + 600
-	src.visible_message("<font color='red'><b>[src] moves their head next to [B]'s neck, seemingly looking for something!</b></font>")
+	src.visible_message("<font color='red'><b>[src] moves their head next to [B]'s neck, leaning in closer...</b></font>")
 
-	if(do_after(src, 300, B)) //Thrirty seconds.
+	if(do_after(src, 150, B)) //Fifteen seconds, because thirty is just awkward.
 		if(!Adjacent(B)) return
-		src.visible_message("<font color='red'><b>[src] suddenly extends their fangs and plunges them down into [B]'s neck!</b></font>")
-		B.apply_damage(5, BRUTE, BP_HEAD) //You're getting fangs pushed into your neck. What do you expect????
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/carbon/human/species/station/station_special_abilities.dm
```diff
@@ -383,19 +383,21 @@
...
 
-	if(do_after(src, 300, B)) //Thrirty seconds.
+	if(do_after(src, 150, B)) //Fifteen seconds, because thirty is just awkward.
 		if(!Adjacent(B)) return
-		src.visible_message("<font color='red'><b>[src] suddenly extends their fangs and plunges them down into [B]'s neck!</b></font>")
-		B.apply_damage(5, BRUTE, BP_HEAD) //You're getting fangs pushed into your neck. What do you expect????
+		src.visible_message("<font color='red'><b>[src] suddenly extends their fangs and sinks them into [B]'s neck!</b></font>")
+		sleep(25)
+		to_chat(B, "<span class='danger'>You feel light headed as the bite of those fangs grip your senses!</span>")
+		//B.apply_damage(5, BRUTE, BP_TORSO) //changed to torso from head
```
KasparoVy, line 393: Not sure what this comment is for, diff says it used to apply damage to the head already.

buffyuwu: > Not sure what this comment is for, diff says it used to apply damage to the head already.

the original comment was "//You're getting fangs pushed into your neck. What do you expect????"

KasparoVy: I'm just wondering why the comment exists at all instead of just nixing the whole line

buffyuwu: > I'm just wondering why the comment exists at all instead of just nixing the whole line

not everyone is going to check diffs on git, and so, we don't write code comments with the intention of them being read alongside a diff. it is important to notate what changes have been made straying from what was originally present, but also, to aid whoever comes along and reads it in the future in understanding what we've done.

to better show you what i mean, here's an example of you, yourself, notating relatively simple changes in simple terms, alongside what both a diff or just simply reading the line would explain: https://i.imgur.com/f0tavx3.jpg

PR #2569, opened 2020-12-07
Make chemical patches more usefull

TheLordME:
## About The Pull Request
Chemical Patches made at the ChemMaster3000 are currently pretty useless, given there are only a few chemicals that function on touch, most are harmfull. So I went and basicly copied some elements from bruise packs to allow the patches to seal wounds as the name and icon suggests.
In addition to that I added a check to avoid applying patches onto open surgery sites, as is present in bruise pack code.
During testing I discovered a bug that would trigger the check for Robolimbs while on harm intent, even though there was clearly no robotic limb targeted or affected. Not sure where that comes from, my fix would be to have a harm-intend check before hand and kill the function before giving the wrong information to the user.

The Original bruis Pack code I copied was supplied by Kevinz000, just making sure to give them credit for their work.

## Why It's Good For The Game
It adds useability to the chemical patches although there are currently only Tricordrazine, Spaceacillin and Sterilise, that have affect_touch functions.
Next step would be to add receipts for the "topical gel"-variants of some chemicals, that are already present in code, and maybe even add some more varients to it.

## Changelog
:cl:
add: Chemical Patches work as a single use bruisepack. Applying thier Chemical and bandaging a single wound.
fix: Chemical Patches can no longer be placed on surgical sites.
fix: Chemical Patches now correctly check for robotic limbs.
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/patch.dm
+++ b/code/modules/reagents/reagent_containers/patch.dm
@@ -29,12 +29,16 @@
 			if(!affecting)
 				to_chat(user, "<span class='warning'>The limb is missing!</span>")
 				return
-			if(affecting.status >= ORGAN_ROBOT)
+			if(affecting.robotic >= ORGAN_ROBOT)
 				to_chat(user, "<span class='notice'>\The [src] won't work on a robotic limb!</span>")
 				return
 
 			if(!H.can_inject(user, FALSE, L.zone_sel.selecting, pierce_material))
 				to_chat(user, "<span class='notice'>\The [src] can't be applied through such a thick material!</span>")
+				return
+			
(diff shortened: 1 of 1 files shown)


TheLordME:
Upon closer inspection, the issue with the robotic limb is not singular to harmintend interactions, it also applies to broken and bleeding limbs.
The issue is the wrong variable is used for the check, while affecting.status is used, affecting.robotic should be used, the ORGAN_ROBOT is a define not meant for status.
I will provide a fixing Commit in a few minutes.

TheLordME:
Now squashed and the issue should be fixed, atleast from the little testing I have done on my machine.

PR #2570, opened 2020-12-08
Increases Mining Void Rad Resistance So People Stop Dying, Even If It's Funny When They Do

Captain277:
1. _Increases Mining Voidsuit rad resistance to 65._
**Why:** It was originally 20. This should help drop first-run miner lethality.


KasparoVy:
Ahaha. Learned this lesson and try to teach it as much as I can, but can't avoid people getting radsick entirely

Cheers

PR #2572, opened 2020-12-09
marks food as opencontainer so you can pois- errrr take toxins out of them ^w^

silicons:
is this even a fix or a feature i don't know but let me poison food please


VailTheWolf:
You can already poison food as long as you have a syringe.

PR #2573, opened 2020-12-09
Unlocks Exploration Jump Console

Captain277:
1. _Unlocks Exploration shuttle Jump Console._
**Why:** I hate that I'm still having to find and fix this stuff. Yeah I get that Pilots and the Pathfinder should be the ones with primary access to the shuttle controls, but news flash: Sometimes they aren't available, or they're _dead or incapacitated_. We're all adults here. If someone takes the shuttle that isn't supposed to, we can punish them. Not everyone preemptively.


lolman360:
make it hackable?

BlackMajor:
Not sure how I feel about this one to be honest.

PR #2584, opened 2020-12-14
removes some of the goddawful shitcode from protean rigs, removes their complete radiation and taser immunity

silicons:
## About The Pull Request

no more internal unremovable cells that selfcharge
no more snowflake emp protection
no more ai override because IT DOESNT EVEN WORK
their gloves are insulated but the rest of them only halves conductivity
radiation is now 90% instead of 100%

## Why It's Good For The Game

if you want an unremovable cell that links to their nutrition, do it right instead of something that makes the powersink useless
also unironically powercreeped to shit while not even working right

## Changelog
:cl:
balance: protean rigs are slightly less godawful
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/rig/suits/alien.dm
+++ b/code/modules/clothing/spacesuits/rig/suits/alien.dm
@@ -127,28 +127,18 @@
 	name = "nanosuit control cluster"
 	suit_type = "nanomachine"
 	icon_state = "nanomachine_rig"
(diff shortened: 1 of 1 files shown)


CharlesWedge:
You do realize these chems were already obtainable through the cargo ordered Chem Distiller right?

lolman360:
emp_protection = -100 gives them 2x more damage from emps
it's negative protection

PR #2596, opened 2020-12-21
Restores Dual E-Swords

Captain277:
As it says, restores the dual e-sword.

1. _Adds the Dual E-Sword Weapon._
**Why:** I tried to make one of these the other day, and was shocked to find they didn't exist. Thanks Bay or Vore or whoever.

2. _Adds the ability to make Dual E-Swords._
**Why:** How else are you gonna get it? Telecrystals?


code/game/objects/items/weapons/melee/energy.dm
```diff
@@ -313,13 +310,52 @@
 
 	return 1
 
+/obj/item/melee/energy/sword/attackby(obj/item/W, mob/living/user, params)
+	if(istype(W, /obj/item/melee/energy/sword))
```
BlackMajor, line 314: You forgot that science can print eswords too, and are a subtype of this.
/obj/item/melee/energy/sword/charge
Therefore, unless you make this an explicit type, then sci can produce deswords too.

Captain277: I'm not necessarily against Science being able to make them, since they can already print actual e-swords, but I should probably look into adding some kind of barrier to that.

BlackMajor: The one's science make require a battery and use power per strike. The deswords that the charge swords combine into do not share this property. I'd suggest either not doing that, or making a charge variant.

Captain277: Finally getting back around to this. I'll look into adding a step in the path momentarily.

PR #2620, opened 2021-01-11
Medical Gear Unhugbox - Infections

Enzo-Leon:
## About The Pull Request

Medical gear is no longer infinite in terms of CLEAN. You need to start replacing your syringes and gloves and make sure everything is washed and sterile before operating or jamming syringes in people.

## Why It's Good For The Game

Feedback request, and needed change that was removed way long ago.

## Changelog
:cl:
tweak: Re-enabled syringes giving infections if re-used
balance: Adjusted Germ Levels to prep for Virology rework
/:cl:

### DIFF
--- a/_maps/map_files/triumph/triumph-02-deck2.dmm
+++ b/_maps/map_files/triumph/triumph-02-deck2.dmm
@@ -9343,7 +9343,6 @@
 /area/hallway/primary/starboard)
 "BI" = (
 /obj/machinery/conveyor_switch/oneway{
-	convdir = -1;
 	id = "QMLoad2"
 	},
 /obj/machinery/light{
@@ -26153,12 +26152,12 @@ iM
 ai
 oE
 oE
-Ou
+oE
 LW
 oE
 Mr
 OQ
-oE
+Ou
 KT
 cp
 ii
--- a/_maps/map_files/triumph/triumph-03-deck3.dmm
+++ b/_maps/map_files/triumph/triumph-03-deck3.dmm
@@ -516,6 +516,12 @@
 /obj/machinery/atmospherics/pipe/simple/hidden/supply{
 	dir = 9
 	},
+/obj/machinery/button/windowtint{
+	dir = 8;
+	id = "ward_a";
+	pixel_x = 24;
+	pixel_y = 24
+	},
 /turf/simulated/floor/tiled/white,
 /area/medical/patient_a)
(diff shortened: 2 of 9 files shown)


VailTheWolf:
Thank you for bringing this back. Please make sure to drop an announcement in the discord that people need to look out for general infections again. (Which are different from diseases.)

CameronWoof:
Can you generate an infection by injecting the same person with the same syringe multiple times?

Enzo-Leon:
@CameronWoof The bug for that was fixed after testing and I've been able to inject the same patient multiple times in a short spawn with no risk of infection that I've seen so far. However, waiting like a minute and then using that old syringe WILL have the infection risk % tied to it. We're keeping eyes on this to ensure that stuff isn't unfun for medical players for that. 

Also, hypo-sprays are set to never spread infection thus giving science another good item to research and supply the medical team with later into rounds.

@VailTheWolf Aye I've been giving people updates that play medical and stuff consistently on this. So far all the usual medical players are up to speed and I haven't seen an uptick in people going "haha infection"

Enzo-Leon:
Oh yeah, we're also re-doing the ENTIRE medbay to make it handle upcoming infections and the soon we promise change to virology. https://cdn.discordapp.com/attachments/766415627060772885/799103064165580850/medbay_wip.png

Enzo-Leon:
Fix made to the syringe code
```	
//Dirtiness should be very low if you're the first injectee. If you're spam-injecting 4 people in a row around you though,
//This gives the last one a 30% chance of infection.
	var/infect_chance = dirtiness        //Start with dirtiness
	infect_chance += (targets.len-1)*5    //Changed from a 10% increase to 5%.
	if(infect_chance <= 20 && (hash in targets)) // Adjusted number as this should now work properly without infecting the first patient instantly.
		infect_chance = 0
	else
		if(prob(infect_chance))
			log_and_message_admins("[loc] infected [target]'s [eo.name] with \the [src].")
			infect_limb(eo)
```
The old block was running BOTH checks and thus put the first patient at risk of getting infected even though the syringe was new. This new check should prevent the first patient getting infected now for a few extra syringe uses.

Dirtiness goes up by 1pt for EVERY Injection/draw from a player. The 5% increase is per TARGET (or unique user DNA). If you're only using a syringe on one person, in theory there'll only be a 5% increase added to the flat dirtiness.

PR #2626, opened 2021-01-17
Splits Exploration From the Research Department

Captain277:
This PR adjusts multiple files to make Exploration its own department on the station.

1. _Recolors and reconfigures Explo headsets._
**Why:** Standard Explo headsets no longer have Science comms. The Pathfinder variant does.

2. _Adjusts access for Exploration._
**Why:** Standard Explo no longer has Research access. The Pathfinder's access has been removed, but may be re-added.

3. _Consolidates Access Datums into one file._
**Why:** Removes a useless _vr file.

4. _Adjusts the prices on Exploration Gimmick Packs._
**Why:** Getting 4 p90s for the same price as some weak Roman armor was never the point of these packs, obviously. After the extensive testing period, I'm happy with the current pack layouts, and have priced them accordingly.


code/game/objects/items/devices/radio/headset.dm
```diff
@@ -553,6 +553,18 @@
 	desc = "A bowman headset for search and rescue."
 	icon_state = "sar_headset_alt"
 
+/obj/item/radio/headset/explorer/pathfinder
```
VailTheWolf, line 556: Change this to /obj/item/radio/headset/pathfinder/

code/game/objects/items/devices/radio/headset.dm
```diff
@@ -553,6 +553,18 @@
...
+	name = "pathfinder's headset"
+	desc = "Headset used by Pathfinders for coordinating and executing expeditions."
+	icon_state = "exp_headset_path"
+	adhoc_fallback = TRUE
+	ks2type = /obj/item/encryptionkey/pathfinder
+
+/obj/item/radio/headset/explorer/pathfinder/alt
```
VailTheWolf, line 563: Change this to /obj/item/radio/headset/pathfinder/alt

code/game/objects/structures/crates_lockers/closets/secure/explorer.dm
```diff
@@ -198,7 +198,8 @@
 		/obj/item/clothing/mask/gas/explorer,
 		/obj/item/clothing/shoes/boots/winter/explorer,
 		/obj/item/clothing/gloves/black,
-		/obj/item/radio/headset/explorer,
+		/obj/item/radio/headset/explorer/pathfinder,
```
VailTheWolf, line 201: Update these paths respectively 
/obj/item/radio/headset/pathfinder/
/obj/item/radio/headset/pathfinder/alt

VailTheWolf (commented):
No clue why the code doesn't like the item state paths. You might have to use a different name scheme like "path_headset" and "path_headset_alt" in your .dmi instead of "exp_headset_path" and "exp_headset_path_alt". This may be confusing the code. 

Simplify the obj pathing rather than making it a child of the exploration headset. Sometimes I run into problems with Byond hating the relation of parent/child objects.

Captain277:
> No clue why the code doesn't like the item state paths. You might have to use a different name scheme like "path_headset" and "path_headset_alt" in your .dmi instead of "exp_headset_path" and "exp_headset_path_alt". This may be confusing the code.
> 
> Simplify the obj pathing rather than making it a child of the exploration headset. Sometimes I run into problems with Byond hating the relation of parent/child objects.

Good eye. Thank you for spotting this. I was able to fix it quickly with your help. <3

PR #2631, opened 2021-01-18
Protean rad immunity, because it makes sense

Ghostcoffeee:
As protean blobs are rad immune and their humanoid forms aren't, I though it made little sense for a species with 2 forms to have differing resistances, so here I am, making them more similar.

The entire PR is me editing two lines of code, with the express purpose to make proteans more OP.

Also, a personal gripe, there're too many protean apps right now and I will spare a lot of effort due to my laziness to reduce the amount and freely given acceptance of them.
Protein

-Ashet
<3


Zeldazackman:
Looks good, and aye there's quite an amount of applications vibing to be one but for the amount of ones that get accepted? it's far and few between really, which isn't so bad.

Then there's the player count of folks that I'd like to bring up that actively are playing proteans in 2021, of the nineteen or so that have been accepted and whitelisted for the species/race only one person is really ever playing them anymore, and that's sparingly.

PR #2638, opened 2021-01-21
Adds Hyper Pads, Aka Quantum Leap pads, for mappers.

BlackMajor:
Yeah hey, just adding a mapping asset I need for a map I'm working on.

These are hyper pads, or quantum leap pads.
They're effectively fancier quantum pads, cover a 3x3 area, work without power and are safe to use bags of holding with.

Also, fixes a small bug with do_teleport code, and removes the baked in sparks proc from quantum pads, since that's handled by the fixed do_teleport code.


KasparoVy:
Badass, good idea

PR #2646, opened 2021-01-23
Fix: Sleep Indicator being stuck after Surgery

ThatOneTeshari:
## About The Pull Request

So after a while of trying to understand why and how things do. I've opted to just compare the port of Polaris' indicators to that of Chomps, and added in the differences. This in high hopes, fixes the floating sleep icon after surgery. If not, I uh - I'll keep looking into it.

## Why It's Good For The Game

Fixes the stuck status of Sleeping when coming out of surgery.[Hopefully]

## Changelog
:cl:
fix: Fixes the 'stuck' Sleeping indicator after surgery.
/:cl:

## Comments

So while doing this I have discovered, and more harrowing - there is no way to toggle off the indicators. I however am just a simple caveman who doesn't understand how Byond is coded and how it works in general. I would encourage someone with better knowledge and skill to look into this. If not well, get cozy with the icons. ^^'

Sadly no pictures, as it's a fix.


ThatOneTeshari:
This was after comparing code from both Polaris, Chomp and Cit 13 RP.

PR #2653, opened 2021-01-26
Makes The Frontier Phasers Sound Like Guns, Not Slide-Whistles (and the PDW sounds nicer, too)

drofoljaelisglis:
## About The Pull Request

Improves the Frontier Phaser by giving it an actually good sound effect, and also adds a sound effect that (I thought) sounded better than the standard gunshot effect due to request. Also adds an extra sound file in for future use with heavier energy weaponry, since I was there.

## Why It's Good For The Game

The Frontier Phaser currently sounds like a slide-whistle, and it's just laughable, really. The new sound will make it sound more like a gun, and not that you're killing someone or something with a toy.

The PDW sound I changed on request, and although I think that the effect is better than the generic one, it is possible there are better for it. It was a side concern compared to the awful slide-whistle gun.

## Changelog
:cl:
tweak: Changed the Frontier Phaser, Holdout Phaser, and Personal Defense Weapon's firing sounds.
soundadd: New weapon sounds, including one (currently) unused one.
/:cl:


VailTheWolf:
I like the new PDW sound, however it's a little too punchy. Take a listen to the sound profile of the P90 sometime. It's a lot more snappy, enclosed. Not an uzi chambered in .45.

PR #2655, opened 2021-01-27
Nerfs Baymiss and Buffs Intrinsic Accuracy Across the Board

Captain277:
1. _Changes Base Accuracy Bonus from 0 to 65._
**Why:** Guns miss way too much in this codebase, and I think this is the cause. All guns now have 65% accuracy intrinsically. The accuracy rating of guns with custom stats have been adjusted to match this new balance metric. Ideally this will reduce the amount of misses.


lolman360:
> 65% accuracy intrinsically

captain it literally says in the comment that 15 accuracy = 1 tiles closer for accuracy calculations, not % likelihood to hit

Captain277:
The line before that in the same comment: "Accuracy is measured in percents."

PR #2661, opened 2021-01-28
lets fix some grammar

buffyuwu:
minor change that will prevent people from typing strings like:
 "Hello i am Buffy"
and instead fix the grammar into:
 "Hello I am Buffy"
## Changelog
:cl:
fix: minor autocorrects for grammar in chat
/:cl:

if they type a lowercase 'i' where it should be an uppercase, it will make it uppercase. same with i'm -> I'm


lolman360:
can you make this a pref or something

buffyuwu:
> also typos funny

not on my HRP minecraft roleplay server they're not

PR #2664, opened 2021-01-28
Introduces Nano-Peridaxon

Captain277:
1. _Adds a new chemical called Nano-Peridaxon._
**Why:** Nano-Peridaxon can be made by combining equal parts Peridaxon and Nanite Repair Juice. Nanite Repair Juice can be created by feeding a NIF Repair Device nanopaste. Nano-Peridaxon works exactly like standard Peridaxon, but it also repairs synthetic organs. Nano-Peri's enhanced nanomachines repair synthetic and standard organs, making this an advanced chem. However, its primary design intent is to assist Explorers or Security - and to a lesser extent civilian collateral - exposed to Ion blasts, without having to perform surgery. If the use of Nano-Peri becomes ubiquitous, I will rebalance to make it harder to acquire. This is not meant to replace field surgery.

2. _Reduces Eye Blurry Thingy on Peri._
**Why:** Buffy was kind enough to point this out. The RNG can make your eyes blur for somewhere North of five minutes, which has caused tons of player complaints. This should fix that?

3. _Adjusts the Light Blocking on Dense Webs._
**Why:** This wasn't apparent before, due to dynamic lighting not being active, but the dense webs spawned by spiders block light a little _too_ well.


VailTheWolf:
Only concern I have is that on site "bone healing juice" and "organ be not hurt juice" are supposed to have side effects that hamper them away from a combat stimulate use and used more towards an actual "sit down and roleplay while you recover" Medical Bay treatment. I myself rarely use Peri unless I can observe someone taking O2 damage after a hard fight, and even then that's on the trip back just to mitigate O2 loss. 

Blurry vision is combated by walking at half speed, since you get blurry vision and dizziness side effects both. You're not meant to take it and be able to olympic sprint out of medbay immediately. 

Tl;dr: Be careful with this balance please. I don't want to see people doping Peri every 10 minutes out in the field because "funny chemist made me 20 bottles of this and tricord".

Captain277:
Yeah, this is designed specifically to help offset the the incoming damage from Ion mobs on people with synthetic organs. I'll be keeping an eye on it in regards to the potential for abuse.

PR #2682, opened 2021-02-05
adds lavaland

buffyuwu:
this will be the beginning of our rendition of lavaland. please keep in mind this PR is not the 'finished' or 'final' product, and that it may change with future development and changes as time goes on.

adds a lavaland z level, with scorching ash storms on random intervals and plenty of ore to mine.
there are several submaps that randomly spawn each round around the map as well

things needed for future development:
-hostile fauna
-megafauna
-bosses
-additional submaps

i inadvertently figured out POIs and how broken they are. essentially, unless you give your map a truly rectangular shape, it is highly likely to be rotated to frick in a square shape. a lot of POIs made recently were perfect squares, and thus, get rotated on a whim at roundstart. i don't know the math involved, but the simplest solution is as follows:

keep your POI submap a long rectangle, say 25x 50y, and ensure all others follow that structure. make sure the area the seed_submaps() proc is placing them into is of ample size, OR, portioned out into 26x 51y sections for perfect POI placement. that way, it is hiiiiighly unlikely the math will manage to fit them all in your zlevel at odd angles and potentially break things like light placements.

## Changelog
add: Lavaland
/:cl:

without submaps view:
unknown


KasparoVy:
RIP asteroid mining?

buffyuwu:
> Is this being added in parallel to asteroid mining?

this has nothing to do with asteroid mining

KasparoVy:
> 
> 
> > Is this being added in parallel to asteroid mining?
> 
> this has nothing to do with asteroid mining

Wicked, cheers. It's my favourite implementation of mining so I can't help but be curious

Captain277:
We're keeping Belt Mining in, so the players have options depending on what kind of mining experience they want to engage in. Player polls indicate they want options.

TheLordME:
Can we on the current state reach lavaland? and if yes how?

buffyuwu:
> Can we on the current state reach lavaland? and if yes how?

mining has a bluespace teleportation pad next to the external airlocks

PR #2693, opened 2021-02-11
Rearranges the Kitchen

Captain277:
1. _Rearranges the Kitchen._
**Why;** Feedback from Kitchen mains has been consistently negative. In an effort to make the kitchen more user friendly and ensure QOL concerns are addressed.
*Expanded the freezer.
*Fixed the Gibber placement.
*Removed the oil tank and put it in Maint.
*Gave the Kitchen a door _to_ Maint.
*Expanded the food presentation table to the South.
*Gave a second door to the diner area.
*Added more stools so we can stop stealing them.
*Moved the fridges so they aren't neighbors.

2. _Adjusts Turfs on Space Food POI._
**Why:** I'm pretty sure this sole POI is the root cause of all of our active turf issues lately. I'm not **sure** it is, but I'm pretty sure. I will edit POIs via this PR until we crack that fucking issue at last.


BlackMajor:
How many times are y'all gonna remap the bar/kitchen this has been happening basically daily.

PR #2715, opened 2021-02-15
changes nanoui script tag to application/javascript to match web server

silicons:
*scream
thanks to nevimer for pointing this out!


silicons:
nope

PR #2721, opened 2021-02-16
fixes energy guns runtiming

silicons:
dispersion is a list


Putnam3145 (approved):
dynamic typing is and always will be a mistake

PR #2731, opened 2021-02-17
Shuttle Thruster Band-Aid

Captain277:
1. _Provides a TEMPORARY Fix to Fuel Economy Breakages._
**Why:** So something recently broke fuel economy. Shuttle thrusters aren't generating thrust at the same level that they were beforehand. Until we isolate that issue, this temporary fix simply reduces the fuel consumption of all existing thrusters. These are **not** the long-term consumption values I want, but they are necessary at the moment. This is being marked Test Merge Only.


Captain277:
Band-Aid successful. Actual shuttle fix completed, gonna close this one now.

PR #2733, opened 2021-02-17
minor autocorrect fix

buffyuwu:
removes whos

### DIFF
--- a/code/__HELPERS/text.dm
+++ b/code/__HELPERS/text.dm
@@ -273,7 +273,6 @@
 											"dont"="don't",
 											"shouldnt"="shouldn't",
 											" ive "=" I've ",
-											"whos"="who's",
 											"whove"="who've",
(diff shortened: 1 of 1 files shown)


drofoljaelisglis:
What's the problem though, who's is the contraction of "who is", ie "who's responsible for this?". It might be better to just make it 
" whos " -> " who's " 
if it's causing issues with words involving whos as part of the word

buffyuwu:
> What's the problem though, who's is the contraction of "who is", ie "who's responsible for this?". It might be better to just make it
> " whos " -> " who's "
> if it's causing issues with words involving whos as part of the word

too ambiguous in usage, and it's mean to be a simple typo fixing proc, not handle nuanced words or instances where different intentions could render different results. for example, if the user means to type 'whose' but they type 'whos' by mistake

PR #2832, opened 2021-03-07
Protean de-jello-fication project

Ghostcoffeee:
## About The Pull Request

So, Protean limbs were 10 hp per limb, meaning a misplaced candle flame or the like would've ashed it. This annoyed me to no end, so I changed it.
I also added a rather minor regen effect to non-blob Proteans (2/s), and buffed the blob regen to 10/s, so that people would have a reason other than ventcrawling/tentacle erp/near-death to go to blobform.

As I am incredibly balance-blind, people will probably think that either the blob regen is too much, or that the entire existence of non-blob regen, a la prommies, is too OP.

## Why It's Good For The Game

Makes the Protean limbs not made of jello. This change also makes them not turn to limbless torsos when exposed to weak EMPs, and makes the limbs able to survive 2 laser rifle shots.
The Protean in general will be able to survive 2 ion rifle shots or 1 medium EMP before blobbing. A heavy EMP, also known as sitting on the grenade, will still kill them very dead.

Also, I'm a bad coder, so you'll have to disable material modifiers to turn off the regen, or it'll tick over to zero. If anyone can tell me how to prevent that, please do.

## Changelog
:cl:
add: Protean non-blob regen, similar to prommies
tweak: Tweaked Protean limb HP to make them not made of nano-jello
tweak: Reduced Protean brute resist as a cause of them not being made of nano-jello
balance: Buffed blobform regen
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/protean/protean_blob.dm
+++ b/code/modules/mob/living/carbon/human/species/station/protean/protean_blob.dm
@@ -171,8 +171,8 @@
 /mob/living/simple_mob/protean_blob/Life()
 	. = ..()
 	if(. && istype(refactory) && humanform)
-		if(!healing && health < maxHealth && refactory.get_stored_material(DEFAULT_WALL_MATERIAL) >= 100)
-			healing = humanform.add_modifier(/datum/modifier/protean/steel, origin = refactory)
+		if(!healing && health < maxHealth && refactory.get_stored_material(DEFAULT_WALL_MATERIAL) >= 50)
+			healing = humanform.add_modifier(/datum/modifier/protean/steelBlob, origin = refactory)
 		else if(healing && health == maxHealth)
 			healing.expire()
 			healing = null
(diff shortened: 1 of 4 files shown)


Captain277:
Howdy howdy, ideally I'll format this all right. I pulled all the Maintainers together to discuss this PR, and here's the overall feedback. Protean Resistances should ideally be brought into the range proposed by Kevin in his Protean Rework draft. I've gone ahead and posted that segment here:

**Temperature**
_Immunity to cold temperatures and its effects
Escalating damage at high (>1000-2000C) temperatures
If this is not enough, at high (>300-500C) temperatures it can be made to give them escalating debuffs until they take burn.
Not final, values should be adjusted to be able to survive small fires but not full blown plasma/tritfires._
**Pressure**
_Pressure immunity_
**Direct Fire**
_Immune - Tied into Temperature_
**Radiation**
_High radiation resistance but not immunity, giving them system instability.
For an idea of how weak/strong, direct exposure to engine core for 1-2 minutes should cause severe issues.
For the love of god give all synths radiation warnings so they don’t suddenly go from “no issues” to “dying” in the span of 5 seconds out of the blue._
**Brute, Burn**
_No brute/burn mods. If really needed, at most 0.8/1.2. Brute/burnmods tend to result in issues because of how damage works in the game and how hard it becomes to balance things, so more than not likely to just be 1/1._
**Conductivity, Electrocution Damage**
_Higher conductivity
Electrocution damage results in burn and instability, instead of just burn. This will, however, result in less damage than now - they will not have inherently super high electrical damage plus burnmod, see below for instabiilty.
All electrocution damage should be treated similar to how electrical stuns work, as they are basically made out of highly conductive material._
**Tox/Instability**
_Could do this for everyone or just proteans, but honestly it’d be neat if this wasn’t limited to proteans._
**Ion/EMP**
_Brute, burn, instability. They’re a cloud of nanites, ions should absolutely ruin their day.
Not instakill levels. Balance towards 2-3 ion shots to near-incapacitate, but 5-7 to full kill. Numbers up to adjustment.
This is a direct hit from an ion rifle. In the below case of floodfill, this would result in 70-100 strength, depending on how we normalize EMP floodfilling and wave spread/dispersion falloff._

On top of the above design input re: resistances, we've agreed - I think - that Proteans should need to actively enable regeneration, and therefore should not passively regenerate when not in blobform.

silicons:
Oh, uh, I meant the segment at the top of "what to do now"
the above is.. not very doable at current time by really anyone but major coders, imo. Up to them.

PR #2836, opened 2021-03-08
maybe fixes walls

buffyuwu:
lets see

:cl:
fix: wall deconstruction
/:cl:


silicons:
the above lines that are being removed isn't live yet
if it continues to break after a server update ping me
this pr will cause an infinite loop in destroy logic

PR #2860, opened 2021-03-13
Removes unscrubbable fire hazards: lowers MINIMUM_MOLES_TO_FILTER to 0.01 (which is still too high, this is zone atmos not tile atmos there's zero excuse for it requiring this level of cheapskating for performance). 0.015 moles is now required for combustion.

silicons:


Putnam3145:
this seems like a premature microoptimization that atmos code is absolutely covered in that you can honestly just remove completely

PR #2867, opened 2021-03-14
Moves the Medigun to Cargo and Adds Defib Sickness

Captain277:
1. _Moves the Medigun and Tier 0 Medigun Cells to Cargo._
**Why:** The Medigun is pretty OP and kinda shit in terms of "gameifying" Medbay, so I'm moving it to Cargo instead of leaving it to be crafted in the first half hour of the round. Tier 0 Medigun cells are also being moved to Cargo. Tier 1+ cells and specialized cells are still craftable via Research, which should make them feel more like researchable upgrades.

2. _Adds 'Defib Sickness'._
**Why:** Players that are revived via Defib now experience Defib Sickness upon rez. This translates to approximately two minutes of Confusion, and ten minutes of Weakness. This addition exists specifically to address concerns regarding LRP attitudes regarding death, especially in the face of the Defib Timer hard limit being removed. Since Resleeving has become less vital for field operations, this mechanical limitation should help rebalance things.


nik707 (approved):
e

PR #2880, opened 2021-03-21
separates global and local narrates again .local narrates now require a target to use the view() of, because why would you kill the muscle memory of everyone who admins main like this 

silicons:
## Why It's Good For The Game

QUALITY OF LIFE
*SCREAM

## Changelog
:cl:
admin: Global-Narrate is back. Local-Narrate is now a thing, instead of being combined with Narration.
/:cl:


nik707 (approved):
e

PR #2888, opened 2021-03-23
my friend lagged into lava so what if we have every client ping the server every 200 milliseconds and halt input if they dc for more than 500.

silicons:
i'm insane

### DIFF
--- a/code/controllers/subsystems/input.dm
+++ b/code/controllers/subsystems/input.dm
@@ -18,6 +18,9 @@ SUBSYSTEM_DEF(input)
 	/// Macro set for classic.
 	var/list/input_mode_macros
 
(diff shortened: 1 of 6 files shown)


silicons:
autopings break reconnect as it'll cause byond to try to reconnect to a closed socket due to it never stopping
we could fix it by mkaing reocnnects shut it off but it isn't worth it, as then hard reboots won't reconnect
closing, until i have time to find a way to hook this into SSserver_maint

silicons:
here we go with experimental fixes again

silicons:
connection problems continue :/

need to make it a client pref too

buffyuwu:
just when i think you can't find something incredibly complex to rework you find a new way <3

PR #2909, opened 2021-03-27
synthetic heat sinks now only activate if the user is actually over species.heat_discomfort_level. 

silicons:
life() continues to be a joke that makes me lose faith in every coder who's ever worked on it


buffyuwu:
> life() continues to be a joke that makes me lose faith in **every coder who's ever worked on it**

hey thats
_checks notes_
us

nik707 (approved):

PR #2910, opened 2021-03-28
adds engine loading weighting, defaulting to 40/40/20 supermatter/rust/fission (untested webedit PR that touches subsystems btw)

silicons:


maps/submaps/engine_submaps/engine.dm
```diff
@@ -14,7 +14,7 @@
 	allow_duplicates = FALSE
 
 /datum/map_template/engine/rust
-	name = "R-UST Engine"
+	name = "EngineSubmap_RUST"
```
silicons, line 17: this is required because config entries have their key/values separated by spaces, it's possible to do it the other way around but i'd like to avoid it

nik707 (approved):

PR #2912, opened 2021-03-28
fixes #2701

### LINKED ISSUE #2701: NIF can display null-name daydreaming text at spawn
#### Brief description of the bug
There's a 1% chance to display a daydreaming message whenever a NIF notify proc happens, which seems to display a nulled-out name when triggering on the 'welcome back' text.
This is pretty minor but might hint at something weird going on behind the scenes, since there doesn't seem to be any reason for it to happen. It might also be an isolated case, but it's hard to tell considering the rarity.

#### Steps to reproduce
- Log in with a NIF 100 or more times on average.
- You might want to fiddle with the proc chance to actually test this.

#### Code Revision
BYOND Version: 513.1533
Server revision compiled on: 
Master commit: bfd2ad249c7850bbd63de2a0b1f2e13ca31496ea 
The following pull requests are currently test merged:
#2695: 'makes soulcatcher verbs clientside' by silicons at commit 177ffb74c9 
#2694: 'fixes colormate matrix mode labels ' by silicons at commit ef45cdda37 
#2696: 'changes NIF init order to make the soulcatcher/commlink hopefully AFTER the new character mob is spawned in, instead of before.' by silicons at commit b7f26013fd 
#2698: 'Mapping consolidation pt. 1: AKA everyone will now conflict. Featuring several bug fixes.' by BlackMajor at commit a997455df5 
#2697: '[Testmerge only] Adds new exiting medicine for patches and replaces/reworks the old ones' by TheLordME at commit c20ff7b7c4 

Local commit: e1f239a7b44e98ec3a43a3fdcb69498adad5baa7
Server tools version: 4.8.0

buffyuwu:
just using [human] here will make it return blank space, so we force it to use human.real_name to pull the real_name var from the mob/living/carbon/human

closes #2701

### DIFF
--- a/code/modules/nifsoft/nif.dm
+++ b/code/modules/nifsoft/nif.dm
@@ -370,7 +370,7 @@ GLOBAL_LIST_INIT(nif_id_lookup, init_nif_id_lookup())
 	if(!human || stat == NIF_TEMPFAIL) return
 
 	to_chat(human,"<b>\[[icon2html(thing = src.big_icon, target = human)]NIF\]</b> displays, \"<span class='[alert ? "danger" : "notice"]'>[message]</span>\"")
-	if(prob(25)) human.visible_message("<span class='notice'>\The [human.real_name] [pick(look_messages)].</span>")
+	if(prob(1)) human.visible_message("<span class='notice'>\The [human.real_name] [pick(look_messages)].</span>")
 	if(alert)
 		SEND_SOUND(human, bad_sound)
 	else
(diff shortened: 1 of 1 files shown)


code/modules/nifsoft/nif.dm
```diff
@@ -370,7 +370,7 @@ GLOBAL_LIST_INIT(nif_id_lookup, init_nif_id_lookup())
 	if(!human || stat == NIF_TEMPFAIL) return
 
 	to_chat(human,"<b>\[[icon2html(thing = src.big_icon, target = human)]NIF\]</b> displays, \"<span class='[alert ? "danger" : "notice"]'>[message]</span>\"")
-	if(prob(1)) human.visible_message("<span class='notice'>\The [human] [pick(look_messages)].</span>")
```
silicons, line 373: not a fan of nerfing personally

PR #2925, opened 2021-04-01
why is this broken

nik707:
## About The Pull Request
this cable is fucked in SDMM but ??? works on main???? so idk what this PR is

## Why It's Good For The Game
e

## Changelog
:cl:

fix: fixed something that might not even be broken?

/:cl:

### DIFF
--- a/_maps/map_files/triumph/triumph-01-deck1.dmm
+++ b/_maps/map_files/triumph/triumph-01-deck1.dmm
@@ -14099,6 +14099,7 @@
 "XP" = (
 /obj/machinery/power/terminal,
 /obj/structure/cable{
+	d2 = 8;
 	icon_state = "0-8"
 	},
 /turf/simulated/floor/plating,


_maps/map_files/triumph/triumph-01-deck1.dmm
```diff
@@ -14099,6 +14099,7 @@
 "XP" = (
 /obj/machinery/power/terminal,
 /obj/structure/cable{
+	d2 = 8;
```
silicons, line 14102: why?

_maps/map_files/triumph/triumph-01-deck1.dmm
```diff
@@ -14099,6 +14099,7 @@
...
 	icon_state = "0-8"
```
silicons, line 14103: this should be 8-0 shouldn't it? don't quote me btw i don't do cabling

PR #2936, opened 2021-04-02
"the codebase never ceases to amaze me with its amazing ideas and piss-poor implementation": air injectors.

silicons:
## About The Pull Request

adds ui to air injectors. you can now modify on/off and pump rate without a console to control it

## Why It's Good For The Game

why in the name of all that is holy do you have it so you can only control pre-mapped injectors
what happens if you lose the console??
even this isn't preferable but it's slightly less awful than not being able to do it at all.

## Changelog
:cl:
add: air injectors are now OP.
/:cl:

### DIFF
--- a/code/modules/atmospherics/components/unary/outlet_injector.dm
+++ b/code/modules/atmospherics/components/unary/outlet_injector.dm
@@ -96,6 +96,46 @@
 
 	flick("inject", src)
 
+/obj/machinery/atmospherics/components/unary/outlet_injector/ui_interact(mob/user, datum/tgui/ui)
+	ui = SStgui.try_update_ui(user, src, ui)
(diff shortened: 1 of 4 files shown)


code/modules/atmospherics/components/unary/outlet_injector.dm
```diff
@@ -158,10 +198,9 @@
 	update_underlays()
 
 /obj/machinery/atmospherics/unary/outlet_injector/attack_hand(mob/user as mob)
-	if (!src.allowed(user)) // ID check, to prevent randos from switching off secure atmos equipment.
```
silicons, line 161: yes, i removed the access check
if you don't want people switching off your "secure atmos equipment", secure it better.

PR #2944, opened 2021-04-04
adds bodydrop sounds when resting

buffyuwu:
and so it begins


buffyuwu:
TODO: add silicon specific rest sounds

buffyuwu:
consolidated into another PR #2949

PR #2945, opened 2021-04-04
maybe fixes a spam of runtimes and overlays from zshadow.dm

buffyuwu:
let's find out
WIP


silicons:
the issue is they aren't deleted, not that owner doesn't exist
this hides the issue and leaves a memory leak

PR #2951, opened 2021-04-07
Makes fishing less fucking dogshit

nik707:
## About The Pull Request

Reduces max fishing time from 99s to 70s
Removes chances of any fish which use sif meat which to my knowledge do not work in any recipes
Increases chances of getting actual fish over junk

## Why It's Good For The Game

Fish isn't used often enough and that's because fishing takes fucking ages and is more likely to give you shitty items than actual useful fish, and when it does there's a solid chance you get useless fish you can't even cook with.

## Changelog
:cl:
del: removes several sif fish types
tweak: tweaks time on fishing
tweak: tweaks droprates for fish
/:cl:

### DIFF
--- a/code/modules/fishing/fishing.dm
+++ b/code/modules/fishing/fishing.dm
@@ -3,15 +3,15 @@
  */
 
 GLOBAL_LIST_INIT(generic_fishing_rare_list, list(
-		/mob/living/simple_mob/animal/passive/fish/solarfish = 1,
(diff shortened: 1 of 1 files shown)


Auris456852:
Removing all Sif fish makes sense for fishing in the ship's/station's pond... But do remember, all those Sif fish give explorers points for scanning them. May want to make it possible to catch those in different planets sometime in the future.

PR #2968, opened 2021-04-13
more horror character design options

buffyuwu:
🆑 
add: more bone heads, under 'body markings'
/:cl:

requested


nik707:
ok bonehead

PR #2982, opened 2021-04-15
[s] removes the hugbox that is bluespace shielding every area that is associated with erp because holy FUCK WHY 

silicons:
The only ones left are
- supply shuttle (player exploit risk)
- armory (damn security goons get a pass I guess because it'd be unfun if someone just harpooned all the guns out)
- vault (same as above)


nik707:
@silicons conflict, fix and ping so i can merge

PR #2990, opened 2021-04-17
fixed the borg adv health analyzer module desc that referred to it as a carbon dioxide jetpack

Mount0:
im not even going to humor the blurb that goes here normally man


buffyuwu:
ahhh the wonders of copy pasted code

PR #2993, opened 2021-04-19
max mob size reduced to 150% from 200%

buffyuwu:
## About The Pull Request

max mob size reduced to 150% from 200%

## Why It's Good For The Game

mobs shouldn't be 3 tiles long or tall considering for headgear and different turf layers. dog borgs (and normal player mobs) at 200% can take up an entire hallway and no longer interact with things predictably; they enter doors above their sprite, couches can cover parts of them, etc

## Changelog
tweak: max mob size reduced to 150% from 200%


nik707:
YESA

PR #2995, opened 2021-04-19
lets not spam THUNDERDOME.ogg indefinitely in the holodeck

buffyuwu:


nik707:
Leave it playing in the actual thunderdome thing but the rest makes sense

PR #3005, opened 2021-04-20
hacky-ish thing to prevent nifs from wearing out from auto installed software

silicons:


buffyuwu:
thank u, this was annoying

PR #3012, opened 2021-04-22
Medicalborg Rework

LordDoge:
## About The Pull Request
Just a minor buff to medihounds.  

## Why It's Good For The Game
At the moment?  Medihounds are somewhat useless.  While they are fully capable of doing chemistry, the systems given are clunky and the ability to do any sort of surgery is entirely unavailable.  This is especially an issue for single-staffed medical rounds where there's only one hound.  Players sacrifice necessary tools by going to the hound modules, so this buff will hopefully make it not as big of a pain.

## Changelog
:cl:
Adds all necessary surgery tools (no autopsy scanner, though)
2 grippers - Organ + medical
2 new chems in hypo - tricord + tramadol

Combines surgery and crisis borg into one
Balances borg vs hound
/:cl:


LordDoge:
An additional adjustment, up to you all if it's still worthy.  But I don't really see a reason to take off two chems from the hypo compared to the regular medical borgs.

Only difference is tramadol and tricordrazine.

Kraseo:
this seems like a bit of an unnecessary buff and makes the surgeon module incredibly irrelevant, and even if it wasn't, why are you giving it only to the medihound instead of including the normal medborg module?

LordDoge:
> 
> 
> this seems like a bit of an unnecessary buff and makes the surgeon module incredibly irrelevant, and even if it wasn't, why are you giving it only to the medihound instead of including the normal medborg module?

Primarily because I essentially never see any other medical borgs other than medihounds.  The majority of the players who do play borg will be playing hounds, and adding it to the normal medborg module isn't exactly necessary as surgeon is almost exactly the same just with some surgery tools.

At the moment?  Medihounds, in comparison, are almost entirely useless.  They can't even pick things up.  So it seems pretty reasonable to me to buff them.  Essentially, if people want to play a hound?  They can.  But they also have to deal with the downsides, which isn't really helping anyone when medical needs the players.

silicons (changes requested):
if you're giving medihounds these buffs give it to regular borgs too.
there's supposed to be a module difference between crisis/normal/surgeon, too.
if you're just removing that, go the full mile and combine normal cyborgs.
dog modules do not get special treatment.

code/modules/mob/living/silicon/robot/dogborg/dog_modules_vr.dm
```diff
@@ -146,7 +146,7 @@
 	desc = "An advanced chemical synthesizer and injection system utilizing carrier's reserves, designed for heavy-duty medical equipment."
 	charge_cost = 10
 	var/datum/matter_synth/water = null
-	reagent_ids = list("inaprovaline", "dexalin", "bicaridine", "kelotane", "anti_toxin", "alkysine", "imidazoline", "spaceacillin", "paracetamol")
+	reagent_ids = list("bicaridine", "kelotane", "alkysine", "imidazoline", "tricordrazine", "inaprovaline", "bicaridine", "dexalin", "anti_toxin", "tramadol", "spaceacillin", "paracetamol")
```
silicons, line 149: there's duplicates in herei

LordDoge: Yeah I noticed that after, added an additional commit to fix them.

LordDoge:
> if you're giving medihounds these buffs give it to regular borgs too.
> there's supposed to be a module difference between crisis/normal/surgeon, too.
> if you're just removing that, go the full mile and combine normal cyborgs.
> dog modules do not get special treatment.

Yeah that absolutely works.  The reason I didn't give the regular borgs buffs was that they pretty much all have the same abilities.  Though I also believe that having _three_ modules for mediborgs is kind of redundant.

PR #3038, opened 2021-04-28
okay this is going to be a horrible idea but it's this or weeks of manual replacement - icon sync to vorestation via overwrite-or-append merge from cit onto vore icons 

silicons:
branch setup:
- reset all files to vorestation's copy 
- grab all modified files via grep

script logic:
- read in our file and their file from modified files list
- merge in our own icons, overwriting any conflicts. if there are none, appends our icons.

in theory this should be an automated solution ensuring maximum compatibility

downsides
- this is sure to bloat the codebase and we'll have to clean it up later.


AffectedArc07:
image
its ok

PR #3040, opened 2021-04-29
removes fun - no more spawning in dust clouds :( and i guess faster/more organized overmaps init 

silicons:
:(


nik707:
make it for ion clouds n shit too unless you did

silicons:
it applies to all events @nik707

PR #3046, opened 2021-04-29
Adjustments to the R-UST submap + other small map changes

Auris456852:
## About The Pull Request

EngineModified

Adjusts the R-UST further. While a part of it is solely aesthetic, there are 3 changes that are actually noteworthy.

1. The meter on the point where the pipes overlap has been moved to the side, to measure the red end of the hot loop.
2. The blast door has been nudged down, to properly align with the bits of plating leading to the shield diffuser outside. Sure, the R-UST doesn't need to be ejected, but as it is now, the shields will block off any attempts to vent it.
3. The scrubbers were previously right next to the red hot loop pipes, making it mildly difficult to tell which is which. The scrubber and air supply pipes have been moved so that they do not overlap with the actual engine in the slightest.

## Why It's Good For The Game

Simply because it makes the R-UST easier on the eyes while working more like it should.

I do expect to fail some merges because of another PR that was recently passed that didn't pass checks, but alas.

## Changelog
:cl:
tweak: Minor changes to the R-UST.
/:cl:


Auris456852:
Now, I wanted to add this next commit to a separate PR, but alas, I merged and now it's part of this one. Anyhow.
Before:
Before
After:
After

This just makes the fuel lines not be connected to the scrubbers.

Auris456852:
Really unsure on this one not breaking anything, so a testmerge may be in order.

PR #3048, opened 2021-04-30
fixes and adjustments to slimes/xenobiology/fire extinguishers

silicons:
- slimes no longer run at you while immune to projectiles that aren't directly targeted
- mobai respects stuns now
- slimes are stunned for 1-2 seconds instead of 9-10 on split.
- however, injected phoron-gray reactions still get the 9-10 seconds stun.
- water puffs get refactored. they can now hit more than one mob in a tile. this fixes slimes being immune to extinguishers while attached.
- amber slimes now actually prevent a self-cascade.

actual balance changes:

buffs slime cells to 750 chargerate instead of 25 because what's even the point of doing the most dangerous department on the station if you can't do anything fun lol

slimes  no longer die or lose friends from starvation - i'll be honest. this is a dangerous department (and in some cases it should be more dangerous), but if i'm being punished for leaving the lab for an hour after already risking death to get it to end-status, fuck off.


silicons:
this pr wasn't tested btw i'm coding rigs on main rn sorry it should work tho

PR #3095, opened 2021-05-13
>protean code for dying never worked in the first place >this 4 line fix that no one ever bothered to do 

silicons:
i hate fetish coders


silicons:
this isn't tested btw

PR #3140, opened 2021-05-30
makes it so mediborgs actually have burn kits (god i'm really dumb)

LordDoge:
## About The Pull Request
Yeah, I forgot to add the line of code that makes adding the ointment module actually happen.  I quickly realized this when trying to do surgery on a burn victim

## Why It's Good For The Game
I'm dumb and forgot to do this before

## Changelog
:cl:
adds burn kit to medical cyborg module
/:cl:


LordDoge:
why u fail

PR #3191, opened 2021-06-14
Adds Curtains Between the Sauna and Pool, and adds an Onsen to Gaia

Captain277:
1. _Adds curtains separating the sauna and pool._
**Why:** You know why this is here.

2. _Adds some "hot spring" pools to the Gaia mountain._
**Why:** The mountain cave is kinda boring. This expands it some to encourage digging, and also gives more natural springs to bump up engagement.


TheLordME:
The hot springs, both outside and inside the cavern, are unroofed, not sure if its intentional
Meaning it rains into them

PR #3197, opened 2021-06-17
Makes the cigarette but icon slightly larger which might look weird but I HATE clicking butts from assholes who leave them around everywhere

nik707:
tin


SandPoot:
Try putting a border on the sprite with an alpha of 1, helps with clicking with no visual changes.

PR #3211, opened 2021-06-25
Silicons wants them gone

TheLordME:
## About The Pull Request
@silicons 

## Why It's Good For The Game
@silicons 

## Changelog
:cl:
balance: Removes the CE-Hardsuit from Cargo, yell at @silicons 
/:cl:

### DIFF
--- a/code/datums/supplypacks/misc.dm
+++ b/code/datums/supplypacks/misc.dm
@@ -261,7 +261,7 @@
 	containername = "ami hardsuit crate"
 	access = access_rd
 
-/datum/supply_pack/misc/ce_rig
+/*/datum/supply_pack/misc/ce_rig
 	name = "advanced voidsuit (empty)"
 	contains = list(
 			/obj/item/rig/ce = 1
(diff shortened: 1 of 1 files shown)


silicons:
requested due to it being insanely powercreeped and since we've been having constant issues with this to the point of bi weekly CC audits.
i don't have a huge ball in the game right now but we should probably find an alternative suit to give some temperature protection.

VailTheWolf:
> requested due to it being insanely powercreeped and since we've been having constant issues with this to the point of bi weekly CC audits.
> i don't have a huge ball in the game right now but we should probably find an alternative suit to give some temperature protection.

If that's the reason people have been using the Advanced Hardsuit, the normal EVA and the S&R R.i.G.s have that function already. Cannot confirm if the Hazard R.i.G. shares this function too. Public EVA voidsuits are also provided with both O2 and Suit Cooling units. There's no excuse to grabbing the CE's suit.

TheLordME:
I think the temperature protection refers to environmental protection, as you would find on for example lavalands lava.
Its not the temperature controll for synths, but hazard protection

silicons:
ill fix this myself actually, soon :)

PR #3242, opened 2021-07-10
Fixes a few issues with mirrors

nik707:
## About The Pull Request

I never expected people to actually try to remove mirrors with surgery instead of the convenient tool, so I never made mirrors null out when removed through surgery which basically bricked them and made it impossible to implant a new one without var editing. Now it's fixed for when people do it wrong/accidentally do it wrong because they tried to remove shrapnel.
(Because I also forgot how dogshit surgery is for shrapnel removal)

ALSO.
Synthetics don't spawn with mirrors. This is now fixed.
You also are now incapable of installing mirrors into mobs that have no client attached. Use the resleeving console. An error message has been added to facilitate this.

## Why It's Good For The Game

obvious tbh

## Changelog
:cl:
fix: synthetics now spawn with mirrors as intended
tweak: Mirrors can now properly be removed through surgery without bricking the mobs respawning
tweak: it is now impossible to install a mirror into a body with no attached client. Use the resleeving console.
/:cl:

### DIFF
--- a/code/modules/mob/new_player/new_player.dm
+++ b/code/modules/mob/new_player/new_player.dm
@@ -665,7 +665,7 @@
 		new_character.dna.SetSEState(GLASSESBLOCK,1,0)
 		new_character.disabilities |= NEARSIGHTED
 	if(client.prefs.mirror == TRUE)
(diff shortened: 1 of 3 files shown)


buffyuwu:
an

nik707:
> an

what

PR #3287, opened 2021-07-23
removes a really weird song from the jukebox

nik707:
it really shouldn't be there


nik707:
this is entirely untested and im assuming the jukebox isn't complete shit

PR #3298, opened 2021-07-26
Makes Shields Behave Less Stupidly

CharlesWedge:
## Free the Station Pets!

Shields now draw from a much more stricter list when determining which tiles are "Outside" the station. This means in general that shields will no longer block interior pet pens, and will in general behave better when applied to the surface of the Tether map. Further specific turfs can be added at later points to further tweak shield gens and prepare them for any future maps.

## Changelog
:cl:
tweak: tweaked the shield generator's blocked tiles list so the shields can behave more precisely on the surface
/:cl:


CharlesWedge:
This will merge fine when my map PR is merged however I consider too fundamental a change to the mechanics of shields to put it in that PR, so I am keeping it separate here until the map PR is merged

PR #3319, opened 2021-08-05
Ports Triumph Desert World To V-4

CharlesWedge:
## No Content Left Behind!
Explo has taken quite the hit lately with the severe reduction of content from the hardsync. The porting of some of Triumph's explo content should help with giving them more interesting things to do. In addition many of the former POIs of the Desert world have been updated increasing their functionality and to account for the use of dynamic lighting on V-4 and integrating them with some of the new things brought by Hardsync.

## Explorers Rejoice
Desert World Has Returned and better than ever, with all new dynamic lighting and weather. Many locations have been touched up making them more aesthetically pleasing and changing their loot around some. 3 New POIs including a new dungeon style location and an old POI has been fixed to actually spawn. A New Map with entrances to both V-4 Beach and the UD to fully integrate these locations into V-4, making it a 3 Z level planet.

## Changelog
:cl:
add: New Virgo 4 Desert Z-Level, 3 New POIs
tweak: Former Desert World POIs are reworked for use with dynamic lighting.
V-4 Beach/Cave: Have been slightly edited to connect to the new desert location. Also V-4 Beach Solars are now actual solars rather then assemblies, they are also now fenced.
tweak: External Plating and external Tiles are now considered Outdoors for purposes of planet lighting.
code: All the code related to making a new linked Z level loaded
/:cl:


TheKnightofAura:
would it be possible to make the desert entirely sand, rather then a mix of beach and sand? I can't exactly see the diffrence between the two, but from my gloss over the desert is almost entirely beach, and the beach area is like 80-90% sand

Also, did this touch existing V4 PoI in the caves? if so, I don't know how I feel about having the Xenomorph pods from the Airless spawning in the V4 caves- in lore, Xenomorphs are a bit more... devastating then they are in practice- I don't think they're a good fit for V4.

CharlesWedge:
> would it be possible to make the desert entirely sand, rather then a mix of beach and sand? I can't exactly see the diffrence between the two, but from my gloss over the desert is almost entirely beach, and the beach area is like 80-90% sand
> 
> Also, did this touch existing V4 PoI in the caves? if so, I don't know how I feel about having the Xenomorph pods from the Airless spawning in the V4 caves- in lore, Xenomorphs are a bit more... devastating then they are in practice- I don't think they're a good fit for V4.

The sand would look different but something is screwing with the way icons are generating (they don't actually look the same in map maker. We will see if I or someone else tries to fix it). Also the PR doesn't touch the way ruins gen in V-4 caves, so the xenos are meant to be there. They definitely shouldn't mind you considering their lore (On triumph I kept them on the Airless Class-D World for a reason). I am not touching the caves for a bit, so the xenos will be around a while still until I can screw with ruin spawning on V-4 caves.

PR #3335, opened 2021-08-12
Window melting fix

Ghostcoffeee:
## About The Pull Request

It makes windows of all kinds melt similarly to walls. Rather simple, but it will probably annoy engi players.

Basic windows will break at 500 C, R-windows will break at 1000 C.
For engineering and toxins reasons, borosilicate will not melt.

Enjoy your hellburns.

This is only if they're next to a fire, same as the walls.

## Why It's Good For The Game

Consistency. Why can walls melt, but windows not?

## Changelog
:cl:
fix: Windows can now melt.
/:cl:

### DIFF
--- a/code/game/objects/structures/window.dm
+++ b/code/game/objects/structures/window.dm
@@ -486,7 +486,7 @@
 	icon_state = "window"
 	basestate = "window"
 	glasstype = /obj/item/stack/material/glass
-	maximal_heat = T0C + 100
+	maximal_heat = T0C + 500 // Bumping it up a bit, so that a small fire doesn't instantly melt it. Also, makes sense as glass starts softening at around ~700 C
 	damage_per_fire_tick = 2.0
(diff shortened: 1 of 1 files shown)


TheLordME:
Id bump the temp for 'normal' glass up to perhaps 250°C, because windows that melt from steam sound a bit stupid.
Most windows are atleast reinforced anyways, and usually a phoron fire heats the room much higher.

nevimer:
This kills atmos experimentation if anyone actually did any.

nevimer:
Big issue with default temps is that nothing can contain a plasma burn anymore, not even nacelles on triumph if you did this. Supermatter's at high temps would casually burn down their containment. The RUS-T in improper shutoffs would burn down more than just the walls to space. Having the highest tier phoron glass invulnerable to atmos is probably the best choice, and normal glasses already break.

PR #3340, opened 2021-08-18
CMO Locker's were missing things, I fixed it.

Jaybirdnerd:
## About The Pull Request
CMO's locker was missing the cloak and the spiffy hand-analyzer that we usually have for CMO's. Figured this a bug, so it's fixed with this PR.

## Changelog
:cl:

fix: CMO's locker was missing the Job Role Cloak and the Advanced Medical Hand Scanner.
/:cl:


Solaris-Shade:
I can dig this! plus one and all that!

PR #3341, opened 2021-08-18
Part 1/4 of my update to explo/cargo

TheKnightofAura:
## About The Pull Request
Splitting this up into four parts, because i'm super skittish about touching the shuttle points, and this is a good break spot anyways- This moves the QM and PF's offices, adds a 'secret' escape route, fills in the old space with some nice things, and as a bonus, fixes three other map issues on A1 while I was at it. ((A wall/door that lead to nothing, a timeclock terminal, and a stray pipe.))

Unlike the later PRs, which is why I'm doing the split here, this keeps the departments as closed up as they where before, and thus, I don't imagine there'll be any fuss over the changes. Maybe this'll also get me a maintainer actuially asking what the fuck I'm planning, so I can point them to the dozens of times I asked them specifically if what I'm planning was gonna fly or not. We'll see!

## Why It's Good For The Game
It gives the cargo pets a slightly bigger pen, adds a fun little thematic escape route for the PF, fitting the frontier nature they're supposed to have, adds in a room we had on triumph but are missing here, (the security station barely attached to explo, for security to yell at them about contraband. This one's lacking the medical tools Triumph had, but, meh.)

As a side note, it doubles the number of gift vending machines we had on the station from 1 to 2. the lack of boxes of chocolate to give out is criminal. 

## Changelog
:cl:
Moves QM/PF office, adds things in new empty space
Fixes some minor map issues
/:cl:

### DIFF
--- a/_maps/map_files/tether/tether-05-station1.dmm
+++ b/_maps/map_files/tether/tether-05-station1.dmm
@@ -21834,7 +21834,10 @@
 /turf/simulated/floor/tiled,
 /area/security/checkpoint/science)
 "gfU" = (
-/obj/structure/closet/crate/secure/weapon,
+/obj/structure/closet/crate/secure/weapon{
+	name = "Security Lockup";
+	req_one_access = list(38,63)
+	},
 /turf/simulated/floor/tiled,
 /area/security/checkpoint/science)
 "gfZ" = (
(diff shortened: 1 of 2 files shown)


lightcat:
Yes

TheKnightofAura:
As a side note, if it's not clear, this is good to test merge/fullmerge as normal- I'm poking at coding items needed for the final product next, which is why I cut the map progress where I did.

PR #3360, opened 2021-09-09
Creates: Apidaen/Vetalan Races, Potential Incoming Tweaks to Auril/Dremachir

Captain277:
## About The Pull Request

1. _Adds a new Insectoid "Bee Person" race to the game, similar to Vasilissans, with the ability to fly and regurgitate honey._
2. _Adds the Vetalan, a dualistic "race" of other species that have been infected by a biologically altering pathogen. They're basically a lore explanation for vampires. I think I maybe made them too robust. We'll find out._
3. _Adds a dedicated .dm for Angels and Demons, so we can get some more species diversity and get people off of Custom Race memes._
4. _Attempts to add a toggle to Eye Glow._

## Why It's Good For The Game

1. The concept has been requested a few times, and I think some more insectoid race representation would be neat. We would have bees as well as spiders with this.
2. The lore team had plans to do this and never did. I felt like doing it. Needs extensive testing and iterating.
3. The same as number two.
4. The layer thing is pretty goofy, we should find a way to turn it off and on.

## Changelog
:cl:
add: Adds Bee race to the game. Adds vampire race. Formalizes Angel and Demon races.
/:cl:


AnalWerewolf:
Vetalan's having the strong versions of regular attacks is pretty OP, not to mention their 0.75 brute resistance and 110 base health. More powerful than the most gamed custom species you can possibly make!

And even then as a custom species you can't get the strong attacks. 

Strong attacks e.g. /claws/strong allow the character to _CONSTANTLY_ hit for 5 damage whereas a normal character's attack can range anywhere from 1 damage up to 5. They get max damage, what- why? I understand the biting attack being constantly full damage but not the kicks and punches.

PR #2782, opened 2021-02-25
Traffic Control Updates & Hostile Migration Fix

Enzo-Leon:
## About The Pull Request

This does a slight overhaul of the traffic controller system to focus more on NT specific ships and fleets with a new added feature of fleet interaction with the Marksman itself reporting on the going-on of the area around us. A new fleet variable was added for future maps and lore development too. 

This also helped fixed an ongoing bug with the rodent migration event and should now ALWAYS fire removing the "false alarm" excuse we had used prior.

Also Sol-Gov stuff was updated to "The Orion Confederation"

## Why It's Good For The Game

QOL and story building to the environment and giving more options for mappers/lore team members to expand on the on-going story.

## Changelog
:cl:
add: new ATC options
tweak: ATC timings & weights
fix: Rat migration no longer is broken
code: air_traffic & organizations files were updated with the new fleet information. More NT ships were added, homeage to the NSB Adephagia added as a station destination for NT vessels to travel off to.
/:cl:

PR #3371, opened 2021-09-13
if you cannot code your air traffic spam right it does not deserve to exist

silicons:
someone please fix this when they can be bothered to, this is causing runtimes every trigger.


KillianKirilenko:
Culprit is #2782. Enzo apparently removed the "NO" state from that line without understanding that the system would shit itself.

Revert
```
		"a current system time sync" = list("sending time sync ping to you now"),
```
to
```
		"a current system time sync" = list("sending time sync ping to you now", "your ship isn't compatible with our time sync, set time manually"),
```

Captain277:
Hey @silicons I'm ready to fullmerge this, but would you look over Killian's comment to determine whether it just needs to be tweaked first for me?

Captain277 (changes requested):
"a current system time sync" = list("sending time sync ping to you now"),

To:

"a current system time sync" = list("sending time sync ping to you now", "your ship isn't compatible with our time sync, set time manually"),

silicons:
ah thanks!

Captain277 (approved):
It isn't showing the original request thingy. This should clear that I think? Git is weird.

KillianKirilenko:
Score one for `git blame`. 👌

Shit like this is why I always gave the system several multi-hour runs to catch possible issues.

PR #3384, opened 2021-09-16
removes telescience shielding from everything but the armory  and vault

silicons:
there's no reason for this to be on most of the station
 scenes do not prioritize over gameplay


Solaris-Shade:
Now no table is safe from teli-boi.

PR #3385, opened 2021-09-16
dark purple slime detonations now use an atmos floodfill instead of view() 

silicons:
having a slime phase gas through windows is annoying


Captain277:
@silicons code/modules/mob/living/simple_mob/subtypes/slime/xenobio/subtypes.dm:212:error: GLOB.cardinals: undefined var

PR #3389, opened 2021-09-16
fixes 'sticky limbs' bug in character creation

rpilcrow:
## About The Pull Request

fixes the 'sticky limbs' bug in character creation previews by making update_icon_preview() under the preferences datum check a variable to see whether it should regenerate the mannequin's limbs. if true, it does so and resets the var back to false.
alongside this, the procs that are used to alter limb states from the character creation menu now set this var to true.

none of this touches player mobs, only the preview mannequin.

_this no longer does any of the below, but i'm keeping it for posterity. there's always a simpler solution. which is probably also one that doesn't blow people up._

> altered copy_to_mob() in 03_body.dm to check for whether it needs to generate a new external organ. 
> if it does (ie amputated limb to normal or robotic, changing robotic limb model, or robotic or amputated limb to normal) then it removes the limb and regenerates it. if it doesn't, it does what it needs to do to the existing limb.
> this required changing the order in which it goes through the limbs so that it won't try to, ie, put a hand back onto a missing arm, and this doesn't seem to have any apparent side-effects.
> 
> tested this with human characters and various other races including zaddat, skrellian, and tajaran, tried various combinations of adding/removing limbs and changing limb states including using the full body option.
> 
> there's almost certainly a better solution, but this bug has existed for at least a year now and makes creating an FBP character an exercise in frustration.

## Why It's Good For The Game

i just want to make an FBP, dude
it won't btfo prometheans anymore either

## Changelog
:cl:
fix: Fixed 'sticky limb' bug in the character creation preview. Prosthetic/Amputated limb states won't end up stuck.
/:cl:


buffyuwu:
thank you

Solaris-Shade:
Does this fix also work when you change tails and or wings?

rpilcrow:
i hadn't noticed any issues with tails or wings. what's wrong with them?

Solaris-Shade:
none that I know of, I was just asking if the other tab updated as well.

+1

Captain277:
I'm pretty sure this PR is murdering Proteans and causing significant issues with Synthetic organ detection/generation

Protean Jelly
.

rpilcrow:
fun. FBPs work perfectly fine but i didn't think to test proteans. i also hadn't realised that this code is also applied at spawn, but in hindsight that makes sense.
knowing that happens, i think that's because protean species code robotizes each limb, and then this runs afterwards.

evidently that's not the solution, then, but at least i still have a good understanding of what the bug is and how to fix it. there might be somewhere better to move the limb regeneration to make the previews work.

rpilcrow:
looks like limb states are still sticky when loading a new slot since i haven't added anything to that. should be a simple fix. will probably push something tonight.

PR #3400, opened 2021-09-19
defibs no longer lie to players because there is no such thing as "further attempts futile"

silicons:


code/game/objects/items/devices/defib.dm
```diff
@@ -282,10 +282,13 @@
...
-			return "buzzes, \"Resuscitation failed - Severe damage detected. Begin manual repair before further attempts futile.\""
+			return "buzzes, \"Resuscitation failed - Severe buildup of toxins and metabolic stress detected. Begin manual repair before attempting again.\""
 
-	else if(H.health + H.getOxyLoss() <= config_legacy.health_threshold_dead || (HUSK in H.mutations) || !H.can_defib)
-		return "buzzes, \"Resuscitation failed - Severe tissue damage makes recovery of patient impossible via defibrillator. Further attempts futile.\""
+	else if(HUSK in H.mutations)
+		return "buzzes, \"Resuscitation failed - Severe tissue damage makes recovery of patient impossible via defibrillator. Begin manual repair before attempting again.\""
+	
+	else if(!H.can_defib)
+		return "buzzes, \"Subject biological state incompatible with defibrillation.\""
```
silicons, line 291: DNM until logic here is rechecked, this is not just feedback this is defib logic and I do believe my PR will break it in its current state so @BlackMajor if you can help out

BlackMajor: Why does that first check check if they're a synthetic for checking oxyloss. 
I'm assuming the first one was for synths, and the second one was for organics.

can_defib only gets flicked to 0 if the patient has been decapitated, and gets flicked back to 1 if you do the head reattachment surgery.

PR #3429, opened 2021-09-28
Janky Widescreen Port

nevimer:
WIP

## About The Pull Request

Ports widescreen mode from TG, kinda janky, probably best for a TM only for now. 
https://cdn.discordapp.com/attachments/472887611757953034/892261227764514846/4snHFuyS3e.mp4

## Why It's Good For The Game

It's widescreen, you get to use more of your screen. 
dreamseeker_s2edToRleg

## Changelog
:cl:
add: widescreen mode
code: changed to widescreen

/:cl:

### DIFF
--- a/code/modules/client/preference_setup/global/setting_datums.dm
+++ b/code/modules/client/preference_setup/global/setting_datums.dm
@@ -368,3 +368,4 @@ datum/client_preference/debug/age_verified
 	key = "USE_WIDESCREEN"
 	enabled_description = "Show"
 	disabled_description = "Hide"
+	enabled_by_default = FALSE //defaults off by request


deathride58 (changes requested):
Needs a comb-through of the code to make non-1:1 view calculations compatible with the existence of widescreen.
Additionally needs a pref to allow people to toggle widescreen on/off

nevimer:
> Needs a comb-through of the code to make non-1:1 view calculations compatible with the existence of widescreen (without this, code that attempts to use the widescreen view as-is will runtime, as it's a string instead of a number)
> 
> Additionally needs a pref to allow people to toggle widescreen on/off

Ah, well, uh, things that change your screen size at the moment just send you back to 1:1 every time you stop using them - binoculars, overmaps, etc. I added two verbs and a preference to toggle and choose. The ugly is overlays like mesons and scanner goggles, they're only covering 1:1 and left align. Pain hud is center align but is also 1:1.

silicons:
needs further discussion before we do this

PR #3442, opened 2021-10-01
A Fate Worse Than Chess!!!: A Chess Board for the Holodeck

CharlesWedge:
## Because Chess is Fun
During those long boring shifts of SS13 I have taken to playing chess in the field of the mind with another player. The problem is that field of the mind chess is hard. I then remarked we needed a chess holodeck program, and our holodeck is a 10x10 so there is really no excuse to not have an 8x8 chessboard. So a did it and it only took an hour. Now I have yet another means of entertaining myself and the wider crew on those long and boring shifts (and the EM team has an excuse to bring back the chess wizard).

## Changelog
:cl:
add: New Chess Program in the Holodeck, with cardboard cutout pieces and a nice board using one of the two currently unused holodeck slots.
/:cl:


CharlesWedge:
And random Xenoarch Runtime has failed me...

TheLordME:
make another minor commit, or ask @silicons to rerun the checks

PR #3447, opened 2021-10-02
Talon Redesign Take 3

UrsaMajora:
## About The Pull Request
Reworks the Talon a decent sum. 

## Why It's Good For The Game
Makes the Talon a bit more imposing while on it, while keeping continuity of how the outline looks vs the skybox image.

## Changelog
:cl:
add:Engine Bay for the engines only.
add:Kitchen/Mess Hall.
tweak:Makes the Cargo bay and Boat a lot larger.
tweak:Adds an airlock for the Exploration shuttle to dock directly.
tweak:Moves the captain's room and vault a little.
code:Renormalizes the FuelDepot 'cause that's a thing.
code:Adds three total areas to the Talon area file.
/:cl:


Knouli:
First off, allow me to say that I really like these changes. They've some really great quality of life implications for those on the Talon. Ranging from the simple 'expand the boat' issue of Talon merchants being assholes to elbows when transporting cargo, to the addition of a kitchen so they have an excuse to eat rather than drink all their calories. I applaud your efforts!

While there are parts I'd modify as a 'preference' side, I'd like to focus my present efforts in performing my role as the dreaded QA tester and start poking at all the parts that stuck out to me.

Biggest and most critical is the fact that Talon's solar power has been cut in half - the starboard side solar tracker is disconnected from the solar array. Despite it's physical presence, no amount of 'searching for devices' will detect it forcing manual configuration of the solar panels to operate. I suggest making corrections to the solar grid such that the starboard solar array's solar tracker is connected to the array.

image

This critical flaw is responsible for a lot of the grief I experienced through my first trial run of new Talon and the ship was under near constant rolling brownouts throughout the course of its operation. This chiefly manifests in the new thruster room at the aftmost section of the ship. The single APC there is responsible for maintaining two maximized fuel pumps on top of the thrusters themselves - which will commonly compete and overwhelm each other. Leaving the options of either fueling or running the thrusters at any single given point of time. I suggest either upgrading the APC with a larger cell to serve better in these high-load situations, or, better yet, breaking up the chamber into three rooms - that each pump is tied to a single APC, and a third APC would be responsible for powering the thrusters.

image

This also manifested as sensors and cockpit losing power - but is a consequence rather than a stand-alone 'bug'.

Both atmosphereic closets are completely disconnected from the power grid. From what I presume to be roundstart, starboard and port engineering, responsible for waste gas and distro gas accordingly, have no power. This cripples Talon's ability to condition the air without manually charging either APC.  I suggest adding these APCs to the power grid.

image

While I do appreciate the use of the large air tanks for distro, I would appreciate the inclusion of inlet and outlet valves and connectors for the array. With all the air pumps that have been added, they cannot be refilled for further use beyond their initial discharge. Further, the chance to refill the Talon's air might prove pertinent in a breech situation where a large section of the ship is depressurized. I suggest the addition of a pump refilling connector in the distribution chamber, or elsewhere for that matter, as well as an inlet to re-pressurize the air tanks with air canisters within the distribution chamber.

image

The new airlock added to permit the expo shuttle's docking with the Talon has improperly set permissions for the docking buttons. With an ID with Talon access you are able to cycle INTO the Talon, as if you were aboard the expo shuttle or in space. However, with the same access you are unable to cycle OUT of the Talon. I suggest that both should be made Talon accessible with ID access for station IDs stripped from the external airlock button.

image

With the redesign of the maintenance tunnels around security an airlock-airlock configuration has cropped up. This is an awkward design, and really encompass the worst of both worlds. I recognize the use-case for having an egress route into maintenance tunnels in a hazardous environment situation, but the present configuration would force exposure to both maintenance tunnels and the main docking chamber regardless. I suggest moving the security doorway one tile port-wards, such that it opens directly into the main chamber rather than opening into the maintenance airlock. Even better would be a means of connecting the security room with the adjacent maintenance tunnel - but this would require extensive redesign.

image

The new messhall is fantastic, but all three stores of food - the meat freezer, the kitchen cabinet, and the refrigerator are all ID access locked to station roles, making them inaccessible to Talon personal. I suggest either removing the ID access lock to these lockers, or custom spawning the associated locker's contents into all-access lockers.

image

The vehicles made available to the Talon previously was the ATV and Space Bike - with this revision they instead have two space bikes as opposed to their long standing complement. I suggest removal of a Space Bike to be replaced with an ATV as before.

image

This is what I've uncovered through my first playthrough and I intend to run a few more tests here and there to see what works and what doesn't!

KillianKirilenko:
Speaking as someone who recently redesigned the Talon for Virgo (crunching it to a single z-level in the process, by staff request), I have to say this is actually a fairly decent job, if somewhat on the large side.

Props to @Knouli for offering all that feedback too. Looking at the solars in StrongDMM, it appears that they're all part of the same grid now, and there are no solars *at all* on the lower deck - which may be where your power issues are coming from.

I suggest stripping out those weird triple-wide windows into the engine bay and placing some more solars on the flanks of the lower level, possibly along the wings as well.

Knouli:
After having had my world view entirely shattered concerning underplating, I'm able to see that the Talon isn't 'wireless magic', but that there's three layer. I'll bite my tongue concerning the greater implications of this discovery and narrow it down to bugfixes for this specific PR.

Chiefly, that the disconnect for the port and starboard atmos closet APCs are a result of the APCs not being wired in correctly. Similarly, that the distribution closet port-side is not connected to the amopshereics network as a whole requiring a single 'L' pipe to reconnect it. I suggest amending these disconnects to tether the closets to the Talon's power grid, and distribution pipe system.

image

Using a similar technique I was able to discover that the starboard side solar track isn't wired to the main gird. I suggest replacing this wire so that the system will have a functional tracker.

Further in-game testing with the Talon's full power grid restored reveals further issue with the thruster APC being unable to contend with the power output of both fuel pumps and thrusters while in-use. Segmentation of the room into three discreet 'grids' is once again encouraged - or at least having thrusters draw power from an alternative power source to prevent thrust lost in-use.

The expansion of the boat has come with other issues - chiefly the lack of roofing for the boat while it is docked within the Talon's internal hanger. While the boat is space-worth while docked at other sites, this results in the precarious situation where you are unable to contain atmosphereics within the boat while docked. Such as in the situation where you took on 'bad air' prior to docking and made a landing within the Talon's hanger, this air would diffuse immediately upon landing regardless of the airlock's integrity. I would suggest re-working the catwalks above the hanger to install plating over the open air space to rectify this issue.

image

Another small nit-pick - but the captain's quarter's bolt button is on the window while there's still available space on the wall itself.

image

Disposals, too, has been reversed somewhere and medbay's trash is emptying into the bar.

image

Beyond the scope of bugfixes, I highly suggest the complete removal of underplating. Maybe I'm ignorant to its greater uses, but from a player perspective all it does is get in the way of repairs and obscures the actual underworkings of the Talon. If you could / will just banish it all from the Talon through the course of your PR, that would be great. Otherwise, I'm like as much to take a crack at it once this rework is full-merged.

Captain277:
Un-TMing due to reports of several mapping issues. It's fine to go back up once they're addressed! Just pulling it for now.

PR #3466, opened 2021-10-06
The Great Randoming

UrsaMajora:
Applies the Randomization code to all creatures I find logically would have genetic differences.

## About The Pull Request
Sets a standard for "genetics" when it comes to organic creatures, in the terms of a randomized code. This makes their health, damage, movement, size, and some cases even their meat be from 50%-150%. Special cases need not apply.

## Why It's Good For The Game
Too long have people gotten to go along thinking "I know this mob has exactly this much health, so I hit it this many times and it will die." This will add in a random factor to the world, making things a bit more exciting. Ranging from a spider that's smaller and faster, to a deathclaw that's larger and tankier. You'll have to pick and choose your targets a bit more than before.

## Changelog
:cl:
code:Added the Randomize code factor, applying it to a large number of creatures (Pets omitted when I was able to)
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/vore/deathclaw.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/vore/deathclaw.dm
@@ -10,7 +10,7 @@
 	desc = "Big! Big! The size of three men! Claws as long as my forearm! Ripped apart! Ripped apart!"
 	catalogue_data = list(/datum/category_item/catalogue/fauna/deathclaw)
 
-	icon_dead = "deathclaw-dead"
+	icon_dead = "deathclaw_dead"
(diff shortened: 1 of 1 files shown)


UrsaMajora:
It's all an early project, so tweaking things like the min-max adjustments is doable.
I chose +-50% because when I was initially trying with +-25%, it sometimes was just hard to notice it being applied at all.
I'm learning how things interact, so I wasn't aware you could overwrite with the initialize, so I can go ahead and put the stats defined in the base again.
On the terms of the movespeed, it's not SUPER noticeable for some creatures. There are several who's base move cooldown is 0 by default, but I went ahead and set that to 1 making them natively slower in all cases.

PR #3474, opened 2021-10-07
Improves Headphones and subtypes with Alt-click on and off

TheKnightofAura:
Not perfect. Can still telepathically toggle them, but I can't figure out how to make them not right now.

## About The Pull Request
Adds alt-clicking to turn headphones, and any subtypes of headphones, on and off for music

## Why It's Good For The Game
Not needing to go into the right click menu for an on and off toggle is really nice.

### DIFF
--- a/code/modules/clothing/ears/ears.dm
+++ b/code/modules/clothing/ears/ears.dm
@@ -37,14 +37,13 @@
 
 	update_clothing_icon()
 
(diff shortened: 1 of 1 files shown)


KillianKirilenko:
Something along the lines of
```
if(!Adjacent(user))
	return
```

ought to fix people being able to toggle them at a distance.

PR #3490, opened 2021-10-09
slippery blood.

UrsaMajora:
## About The Pull Request
Makes blood puddles and gibs have a chance to slip you.
You won't slip if you are walking, wearing non-slip shoes, the blood is dried, or people have already stepped all over it.

## Why It's Good For The Game
Makes people be a bit more mindful of where they step.

## Changelog
:cl:
add: Slipcode to blood.
/:cl:

### DIFF
--- a/code/game/objects/effects/decals/Cleanable/humans.dm
+++ b/code/game/objects/effects/decals/Cleanable/humans.dm
@@ -97,6 +97,8 @@ var/global/list/image/splatter_cache=list()
 				S.overlays += S.blood_overlay
 			S.blood_DNA |= blood_DNA.Copy()
 			perp.update_inv_shoes()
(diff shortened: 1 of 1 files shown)


silicons:
this is terrible for gameplay and instantly gets my veto
don't merge unless the rest of you are 100% sure you want this.

KillianKirilenko:
Having been an Explorer for a long time I can guarantee this would get people killed, because all it takes is one extra click from a machete or knife and the spider you were fighting has been butchered into coating a 3x3 square in gore, leading to a zone of death.

Walk mode won't help either, because then you'll just die when the mobs catch you, rather than because you fell down.

Interesting in theory but it's going to be terrible in practice.

PR #3574, opened 2021-10-27
Adds meme gun

TheLordME:
and moves size gun to the other guns in files

## Changelog
:cl:
add: Adds the stripper gun, same effect as the cell, just as a normal gun
refactor: Sizegun and stripper gun file moved to the other gun files
/:cl:


MalricB:
i believe this is somehow causing size changes (100% > 200% at least) to default to 'fuzzy' scaling mode instead of 'sharp' if you have it selected
spawning at 200% doesn't give you fuzzy scaling
but using size gun / NIF's size program does

Captain277:
Temporarily removing Testmerge to troubleshoot. Will confirm whether PR tangentially affects size scaling.

MalricB:
uhh guess i was wrong, my bad
not a clue wth is causing the issue then

TheLordME:
Nicc and me already trouble shoot that issue, its definitly not cause by the PR, it doesnt happen on my local, even after spawning in as fuzzy and then changing it back

PR #3584, opened 2021-11-03
Fixes Roaches becoming Jesus

yourdoom9898:
## About The Pull Request

It turns out that spiderlings have a bunch of snowflake code when it comes to processing their deaths in order to make it so they don't leave a corpse. However, this caused a issue in which when the code was...borrowed without intent to return...from the spiderlings, it would cause any roach that had to chance to grow up to _always_ grow up unless the roach was killed and gibbed.

## Why It's Good For The Game

 I'd rather not have roachlings starting any religions about coming back from the dead, and it makes no sense for a dead creature to suddenly come back to life new and improved.

## Changelog
:cl:
fix: Roaches no longer come back from the dead by becoming swole
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_mob/subtypes/animal/roach/roach.dm
+++ b/code/modules/mob/living/simple_mob/subtypes/animal/roach/roach.dm
@@ -173,7 +173,7 @@
 		amount_grown = 1
 	get_light_and_color(parent)
 
-/mob/living/simple_mob/animal/roach/roachling/Destroy()
+/mob/living/simple_mob/animal/roach/roachling/death()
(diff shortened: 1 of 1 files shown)


Captain277:
Oh yeah! I remember realizing this was still an issue during later iterations. At the time I thought it was funny, since there's the whole "roaches are hard to kill" thing, but I'm very happy to see a solution to it. Thanks Cass.

CharlesWedge:
part of me wants to code it so all roaches have a 1% chance of reviving (determined on death) a few minutes after death

PR #3606, opened 2021-12-07
no more huds on preferences preview 

silicons:


silicons:
(this isn't tested we'll assume it works)

silicons:
hud rewrite in process

PR #3677, opened 2022-01-14
[RFM] Adherent alright

TheLordME:
## Changelog
:cl:
add: Added Adherent space squids
fix: Fixes us not having adherent
/:cl:


LordPapalus:
Important things:
- Adherent cannot be fixed via normal means, they need to use the mineral bath
  > Any other species that uses the mineral bath, even synthetics, sustains burn damage
- Adherent can charge both via the Electron Reservoir (aka the Pylon) but charging via normal means is slower
   > Any other species that touches the pylon gets burned and knocked back

TheLordME:
Adherent body color can be adjusted via synth body color, which is a seperate variable in the char setup

PR #3696, opened 2022-01-26
welding a shard while holding it no longer embeds a pane of glass in your body

silicons:


TheLordME:
can we keep this as a feature?

PR #3699, opened 2022-01-28
Allows The Illegal Removal of Firing Pins From Guns Without Destroying Them (Not Clickbait)

Captain277:
## About The Pull Request

1. _Enables the illegal extraction of firing pins._

## Why It's Good For The Game

1. Currently, firing pins can only be overwritten by other pins. There is no way to extract extant pins with bootleg modification. Although I can see this being rife with abuse potential, I also believe it frees up a lot of choice and experimentation potential for the players. I'm willing to revisit this to tweak the RNG weights on failure and success chances, after further testing.

## Changelog
:cl:
add: Adds the ability to extract firing pins from guns.
/:cl:

### DIFF
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -287,23 +287,23 @@
 
 	if(A.is_multitool())
 		if(!scrambled)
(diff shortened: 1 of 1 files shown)


lolman360:
just make it screwdriver and 100% of destroying the pin, which is inline with the balance of firing pins

PR #3749, opened 2022-02-22
should fix champagne not interacting with shuttles

TheLordME:
## Changelog
:cl:
fix: fixed that champagne doesnt interact with shuttle consoles
/:cl:


TheLordME:
doesnt actually work yet

PR #3750, opened 2022-02-22
Moves the adherent maintenance room and makes ita bit baldie proof

TheLordME:
## Changelog
:cl:
fix: Fixed idiots touching the forbidden Horse Cock
tweak: Mining and solars space suits now actually use suit storages.
/:cl:


TheLordME:
Floor is not craftable

TheLordME:
crowbaring either breaks or drops normal tiles
placing tiles down should give them the crystal optics again

PR #3805, opened 2022-03-16
Permadeath char as a NPC

TheLordME:
Julia Harding was turned into an abomination on Lavaland east

### DIFF
--- a/_maps/map_levels/140x140/lavaland_east.dmm
+++ b/_maps/map_levels/140x140/lavaland_east.dmm
@@ -80,7 +80,10 @@
 /turf/simulated/floor/tiled/steel_dirty/lavaland/exterior,
 /area/lavaland/east/colony)
 "aW" = (
-/mob/living/simple_mob/mechanical/cyber_horror/surt,
+/mob/living/simple_mob/mechanical/cyber_horror/surt{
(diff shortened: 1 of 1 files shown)


Captain277:
Not sure how I feel about a player character being turned into a simplemob and mapped across Zs.

TheLordME:
general chat seemed to like it, and in this case it was a death that happend on lavaland east, they were killed by a nanite husk thingy, and as lore seems to be that nanites tend to turn people into new things it seemed fitting

Captain277:
A dead DNR character becoming a husk makes sense. Giving the husk a name is a little more strange, considering this PR means you'd just be killing Husked Julia every time the server restarts. For an event where you encounter the mob once it makes sense, but it just doesn't make sense from a long-term gameplay perspective to have the same reanimated corpse respawning every round.

silicons:
i've no opinion on this

on one hand it's cool
on the other hand it's a pandora's box and hurts my sensibilities
on the other other hand we don't have enough devs to make it a pandora box

so
w/e not good but also not really harmful

that said what is harmful is yeah, you're canonically killing the same mob .. on every restart?

TheLordME:
alright, if you feel it unfitting feel free to close it

PR #3829, opened 2022-03-22
Makes people a little less Buttholes

TheLordME:
## About The Pull Request
Cigarette butts now try to be placed in the players hands rather than be droped on the floor

## Changelog
:cl:
fix: Fixed People spitting their cigarette butts on the floor rather than hold onto them.
/:cl:


Zandario (approved):

Captain277:
This might be one of the most genuinely obnoxious PRs I've run into. Putting the cigarette butt in a player's hand just sucks. I'd rather it just equipped to their mouth, if we're going to do this.

Captain277:
Had to re-un-TM it. It works as intended for cigarettes, but it appears to delete cigar stubs and leave their icon sprite permanently in the mask slot now.

TheLordME:
yeah Idk whats causing that, I made an attempt but only ended with cigarettes doing the same

PR #3831, opened 2022-03-24
[RFM]"CRyP70 - 9er" content

TheLordME:
## About The Pull Request
Adds two new machines. One turns energy into points and the other turns points into equipment.

- [x] Points need more and more power per point generated.
- [x] All energy used is converted into heat, to add some challenge beyond, haha engine go brrr.
- [x] Balancing of points generated/spend for items.
- [x] Fix the Vendors UI, currently the top saying how many points you have is still the same as the mining vendors.
- [x] Vendor always says you dont have enough points, even tho you do. Item is properly dispensed if enough points are present.

## Why It's Good For The Game
Gives Engineering a reason to overclock the engine if they want.
and have some balanceing challenge.
Point generators might later be added/ changed to provide more of a challenge than a glorified heater.

## Changelog
:cl:
add: Adds a Crypto Mining server that turns power into engineering points
add: Adds a vendor to spend Engineering points.
fix: Fixes subtypes of the equipment vendor trying to vend twice
/:cl:


yourdoom9898:
Can you please also add a thaler exchange to the vendor?

~~I've wanted turn the cargo warehouse into a crypto farm for ages and I can sell thalers~~

TheLordME:
resolved it

PR #3846, opened 2022-03-29
New knives. New spirits. What could go wrong?

CrystalClearCC:
## About The Pull Request

Adds three new knives (two butterfly, one stiletto)
Butterflies are identical to current butterfly knife - stiletto is incredibly dangerous with the shank mechanic (blue grab, target torso, red click) but otherwise does less damage than a kitchen knife.

Also adds a Malibu knockoff. Doesn't include any cocktails at the moment because I can't sprite for shit. Available from any booze-o-mat

## Why It's Good For The Game

Rare, collectable knives for those who like them.
More spirits is more good.

## Changelog
:cl:
add: Cuban Pete's has released a new rum! This one is white.
add: Three new knives have been added. Knife nuts rejoice.
/:cl:


Captain277:
This is basically good to go, except for two small things!
1. The stiletto should be able to fit into boots, since it's a concealable weapon.
2. The White Rum spawns empty. This is likely because you didn't call an /Initialize with the reagents after creating the bottle!

PR #3856, opened 2022-04-03
Reduces sizeshift trait cost to 2

Mouseofthecake:
## About The Pull Request

Reduces the trait cost for sizeshift to 2.

## Why It's Good For The Game

Since using size mechanics for combat and such is against the rules anyway, I don't think this trait should cost the equivalent of haste. I wanted to reduce it down to 1 point, but 2 seems much more agreeable.

## Changelog
:cl:
balance: reduced sizeshift trait cost to 2
/:cl:


yourdoom9898:
TBH, since size is meant to be purely a non-combat feature, why is this not just a neutral like Brutal Predation or Succubus Drain?

Mouseofthecake:
> TBH, since size is meant to be purely a non-combat feature, why is this not just a neutral like Brutal Predation or Succubus Drain?

good question. i'd be all up for it being neutral tbh

PR #3868, opened 2022-04-06
Add tactical and speed reloads to projectile weapons

Shadowtail117:
## About The Pull Request

Projectile-based weapons can now be tactically- and speed-reloaded. A tactical reload is exchanging a magazine in your hand for the one in the gun (useful if you want to change to a full magazine while keeping a partially-full one on you), while a speed reload drops the magazine in the gun onto the floor while quickly reloading it with one in your hand (useful for dumping an empty magazine and quickly putting a new one in).

When on **help** or **disarm** intent, nothing is changed.

When on **grab** intent, clicking a gun in your hand with the appropriate magazine will initiate a tactical reload. This will start a 1-second timer, after which the magazine in the gun and the magazine in your hand will swap places. You must remain still while doing this.

When on **harm** intent, clicking a gun in your hand with the appropriate magazine will initiate a speed reload. This will start a 0.5-second timer, after which the magazine in the gun will drop onto the floor and the magazine in your hand will be loaded into the gun. You must also remain still while doing this.

__This does not apply to energy-based weapons like tasers as they have separate logic and also have their own delay for putting new cells in that I did not want to interfere with.__

## Why It's Good For The Game

Adds a way for quickly changing magazines during a firefight without having to do inventory sorcery like dropping something on the floor to get a free hand to take the mag out, drop it, pick up the fresh one from the floor to put it in, etc. Clean and simple.

## Changelog
:cl:
add: You can now tactical- and speed-reload a ballistic weapon by clicking on it with the appropriate magazine on grab or harm intent, respectively.
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/projectile.dm
+++ b/code/modules/projectiles/guns/projectile.dm
@@ -28,6 +28,7 @@
 	var/allowed_magazines		//determines list of which magazines will fit in the gun
 	var/auto_eject = 0			//if the magazine should automatically eject itself when empty.
 	var/auto_eject_sound = null
+	var/can_special_reload = TRUE //Whether or not we can perform tactical/speed reloads on this gun
 	//TODO generalize ammo icon states for guns
 	//var/magazine_states = 0
 	//var/list/icon_keys = list()		//keys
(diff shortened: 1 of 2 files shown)


Zandario (approved):
Me and Cap approoooove

PR #3870, opened 2022-04-07
Wall actions are based on intent

Shadowtail117:
## About The Pull Request

Wall actions are now based on intent. If you click a wall on help intent, nothing happens. On disarm and grab intents, you will push it. On harm intent you will punch it (which will hurt yourself if the wall isn't weakened).

## Why It's Good For The Game

Some people like yours truly like to rotate our character by clicking in the direction that we want to face. If we end up clicking on an adjacent wall we push it for no reason which just looks silly. This fixes that.

## Changelog
:cl:
tweak: You will no longer push a wall on help intent.
tweak: On harm intent, you will punch a wall instead of pushing it.
/:cl:

### DIFF
--- a/code/game/turfs/simulated/wall_attacks.dm
+++ b/code/game/turfs/simulated/wall_attacks.dm
@@ -104,8 +104,33 @@
 		else
 			fail_smash(user)
 			return 1
+	if(iscarbon(user))
+		var/mob/living/carbon/M = user
+		switch(M.a_intent)
+			if(INTENT_HELP)
+				return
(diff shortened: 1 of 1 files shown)


yeayea130:
Once again. Borgs don't have disarm or grab intents. so. this makes opening false walls a lot harder for them. if it affects them.

Shadowtail117:
This only affects carbons, which robots/cyborgs are not. Should not have any impact on them.

Shadowtail117:
Linter should be appeased

PR #3887, opened 2022-04-14
fixes skills removal having removed player employment records from being printed

silicons:
guys c'mon can we please check what we're deleting??


Shadowtail117:
Please forgive me I thought that was listing the mechanical skills that they had and not their employment records I didn't know I didn't know I didn't

PR #3896, opened 2022-04-16
Add back autohiss verbs, add type discrimination

Shadowtail117:
## About The Pull Request

Refactors the autohiss file, now the maps are stored in their own datums and not on the species. This is done because now the species you're playing as and the species you autohiss as might not be the same anymore!

You can once again change your autohiss type with the new "Toggle Autohiss" verb in the OOC tab, but you can now use "Toggle Autohiss Type" to change what species you autohiss ass. Playing as a custom lizard but want to autohiss as an unathi? There you go. Catgirl human that wants to speak like a tajaran? Bam.

This also comes with its own preferences selection. **Just like with the old autohiss, changes made mid-round will not be saved for future rounds.**

image

## Why It's Good For The Game

Custom species ahoy, and it can't hurt to be able to change your autohissing midround if something happens.

## Changelog
:cl:
add: Re-added a verb to toggle your autohissing mode, named "Toggle Autohiss" in the OOC tab.
add: You can now change what kind of species you autohiss as (such as hissssing of unathi or rrrrrolling of tajarans) with the new "Toggle Autohiss Type" verb in the OOC tab.
del: Removed the chimera's ability to change its vocal biomorph since the above verbs accomplish the same thing.
/:cl:

### DIFF
--- a/code/modules/client/preference_setup/vore/04_autohiss.dm
+++ b/code/modules/client/preference_setup/vore/04_autohiss.dm
@@ -1,12 +1,17 @@
 #define AUTOHISS_OFF 0
 #define AUTOHISS_BASIC 1
 #define AUTOHISS_FULL 2
+#define AUTOHISS_NUM 3 //The number of autohiss types that we have. Used for sanitization
 
-#define AUTOHISS_NUM 3
+#define AUTOHISS_TYPE_NONE  0//Equivalent to being off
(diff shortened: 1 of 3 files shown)


Shadowtail117:
Would like feedback regarding changing the autohiss *type* midround since that might be abused to change identities. It's one thing to feign it by adding extra sss and rrrs, it's another to completely change a species-determined accent

PR #3907, opened 2022-04-19
Fix nutrition I think

Shadowtail117:
## About The Pull Request

There is a bug where despite spawning with normal nutrition amounts, you'll get the HUD icon for being overfed anyway. I think it was due to a problem with the max_nutrition value so the value was hardcoded instead. (It's the same 450 but when I tested it it didn't break the hud anymore.)

## Why It's Good For The Game

bugfix

## Changelog
:cl:
fix: Potentially fixed an error where you would spawn with the "overfed" nutrition icon despite having normal nutrition amounts.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/life.dm
+++ b/code/modules/mob/living/carbon/human/life.dm
@@ -1271,8 +1271,8 @@
 
 		if(nutrition_icon)
 			switch(nutrition)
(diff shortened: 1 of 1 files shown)


silicons:
i wanted the hud to work with speies max nutrition but fuck it i hate byond

PR #3913, opened 2022-04-22
Add sprinting and stamina

Shadowtail117:
## About The Pull Request

Adds sprinting and stamina as mechanics from https://github.com/Baystation12/Baystation12/pull/25773. Or, more specifically, changes running to be stamina-based so you can no longer do it all the time.

The way this works is that mobs now default to walking. When using the move intent key (default alt), you start running, and a stamina bar appears above your move intent, which goes down with each tile you run across. If it reaches zero, you become exhausted and are no longer able to run for a few seconds, until you recover some stamina.

Stamina replenishes over time when not running, and is (typically) faster if you are lying down or otherwise relaxing.

A couple new verbs were added to the IC tab, default runs and default walks. Basically changing what move intent you go to when holding down the sprint button and when you release it, respectively, and is used for species who have ways to move beyond just walking and running. Currently it's just those two though, so the verbs don't have much use.

-------

Of course, new config options are needed for this.
`SPRINT_COST` is the stamina cost for each tile, defaulting to `0.8`. (Stamina caps out at 100.)
`MINIMUM_STAMINA_RECOVERY` is the least you can recover every time you start replenishing some stamina, defaulting to `5`.
`MAXIMUM_STAMINA_RECOVERY` is the most you can recover every time you start replenishing some stamina, defaulting to `5`.

(The minimums and maximums are the same because...well that's how they were ported. So unless changed, you'll recover stamina at a static rate no matter what.)

## Why It's Good For The Game

No more zooming around everywhere, people can walk at more reasonable speeds and run to places if they need to.

## Changelog
:cl:
tweak: Running is now stamina-based. When running, a bar will appear above your move intent icon that shows your remaining stamina. If it reaches zero, you become exhausted and are forced to walk. Stamina replenishes over time when not running. (Ported from BS12)
/:cl:

### DIFF
--- a/code/__DEFINES/_flags/movement.dm
+++ b/code/__DEFINES/_flags/movement.dm
@@ -11,3 +11,6 @@
 #define FLOATING				(1<<4)
 /// Ventcrawling
 #define VENTCRAWLING			(1<<5)
+
+#define MOVING_DELIBERATELY(X) (X.move_intent.flags & MOVE_INTENT_DELIBERATE)
+#define MOVING_QUICKLY(X) (X.move_intent.flags & MOVE_INTENT_QUICK)
--- a/code/__DEFINES/chemistry.dm
+++ b/code/__DEFINES/chemistry.dm
@@ -43,6 +43,7 @@
 #define CE_SPEEDBOOST "gofast" // Hyperzine
 #define CE_SLOWDOWN "goslow" // Slowdown
 #define CE_ANTACID "nopuke" // Don't puke.
+#define CE_ENERGETIC "energetic" // Speeds up stamina recovery.
 
 #define REAGENTS_PER_SHEET 20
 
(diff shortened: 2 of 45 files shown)


Captain277:
Testmerged. Then, on re-review before round end, noticed that baseline walk speed was *not* buffed in tandem with the addition of sprint mechanics, which was explicitly part of the necessary changes for this mechanic to be considered. Un-testmerged until this has been rectified, pending further Maintainer discussion.

Shadowtail117:
PR has been updated to fix an issue with spawning at the run intent, as well as moving the config options to the modern system. In addition, the default move delay has been lowered significantly resulting in faster movement, as requested. Though, it's still preferable to manage the final amounts on the serverbox and not via this PR.

Video of the new functionality:

https://user-images.githubusercontent.com/25853190/164748704-7e9aa0ea-0ebb-49f4-a5c4-0e9b4f277506.mp4

PR #3917, opened 2022-04-22
what a dizaster

silicons:
we all need our little pet projects to feel more at ease while being one of the three people carrying the burden of not letting RP die

i'll be working on this on and off given this isn't too important but it'll be fun probably 

what's going to be done if i can muster the motivation to finish this:
- auxgm is ported for gasmixtures
- everything uses auxmos procs
- zone rebuilds/whatnot are completely redone
- unsimulated zones are reworked
- most of the game will use turf air api instead of return_air()

limitations:
- atmos machinery will still be a mess even after this - their operations on the zones aren't canonical 
- reactions will not be sorted out after this
- the way i'm doing new zones will not unnecessarily hold onto gas mixtures, but this results in a relatively amount of gas mixture churn
- this system has no way of dealing with planetary turfs - without zone dividers, it'll be just as expensive as before to rebuild
- a heuristic might help here, since we don't want to write "true" planetary turfs


silicons:
this needs a rethinking.

PR #3920, opened 2022-04-23
Creates Modular Limb Plates

Captain277:
## About The Pull Request

1. *Adds Modular Limb Plates and EMT Subtype.*

## Why It's Good For The Game

1. _Player request/commission. These armor plates are lightweight and impart a small protective value. They take the suit slot, so you cannot wear them in tandem with other armor or protective equipment. They're intended to be largely cosmetic, but due to the complexity, minor protective values were assigned, as I am wont to do sometimes. These currently only naturally spawn in the EMT configuration in Paramedic lockers. However, an unpainted base set is spawnable by Admins. I didn't provide any other natural vectors since they're negligible, and are primarily meant to serve an aesthetic purpose._

## Changelog
:cl:
add: Adds Modular Limb Plates
/:cl:


yeayea130:
MY BRAND

PR #3928, opened 2022-04-24
Makes ERIS floors craftable

Ghostcoffeee:
## About The Pull Request

Makes ERIS floors craftable from steel and plastic.

I've removed the multicrafting, because it seems to be bugged in my tests, so turn on your autoclickers

## Why It's Good For The Game

Fewer ahelps for tiles

## Changelog
:cl:
add: Added ERIS floor crafting
/:cl:


Zandario:
Do we really want to start integrate Eris floors? I've very against mixing our visual style with ERIS's. I plan to unify our style completely, which may include nuking these tiles, so I'm unsure if it's a good idea do this before they end up getting removed. Which I know people will complain about.

PR #3965, opened 2022-04-30
[WIP] [MDB IGNORE] [IDB IGNORE] Gemma Colony Map

TheLordME:
## About The Pull Request
Can't be that we have only one open Map PR

## Why It's Good For The Game
Maps I guess

## Changelog
:cl:
add: Added new Map
/:cl:


TheLordME:
My current idea for lore behind the map is:
NT has a claim on the fronteir system, but doesnt have the manpower to properly establish the exploitation of local resources, so they send their _System occipyers_ (the players) in to make sure they dont use the claim, until proper work force can be send.

PR #3972, opened 2022-05-02
Real World Material Statistics

CharlesWedge:
## Trying My Best to Make Mats More than Reskinned Steel

Materials Now have proper weights based on their density, hardness based on the Mohs Scale, their IRL melting points and proper conductivity. This is all in the effort to make materials more diverse so they don't all function like reskinned steel sheets. Perhaps this will also encourage the use of more crafting reciepes with the various materials 

This PR is a bit experimental and I am not quite sure how the new values will effect various buildables with these values. As such please Testmerge thoroughly before merging in full and give me feedback about any issues you may find.

## Changelog
:cl:
tweak: adjusted the variables for various materials based on real world values for such.
/:cl:


silicons:
bad plan
haphazardly enforcing realism

PR #4030, opened 2022-05-11
Adds +1 to Most Racial Language Limits

Captain277:
## About The Pull Request

1. *Increases General Language Selections by One.*

## Why It's Good For The Game

1. _Honestly, I'm not entirely convinced it is. This should only be tested for a while. The motive here is that I want to be able to speak Vernal but I don't have room. I would've just added it to Custom Species, but I mean, that just disincentivizes playing other races so I can't do that. Consequently, I've lifted the language count on nearly every race by 1, just to see how it impacts the setting._

tweak: Adds +1 to the language count for every race.
/:cl:


nik707:
This feels kinda like it'd just make languages even less unique, and hard to find ones that other people don't all understand. Granted I think the language system entirely can kinda lead to metabubbles forming but who knows. Interesting to see how testing works

PR #4079, opened 2022-05-18
Added Utility Coat Clothing Item

CommissarCorg:
Added a suit clothing item that I sprited, called Utility Coat, its in the loadout as well.


CommissarCorg:
So, I don't know why its also trying to add that map edit commit I did months ago.

PR #4104, opened 2022-05-23
lifeweb

silicons:
circuit reagent generator now produces far more power if the blood is from a player


code/modules/integrated_electronics/passive/power.dm
```diff
@@ -109,7 +109,8 @@
 	desc = "Produces electricity from chemicals."
 	icon_state = "chemical_cell"
 	extended_desc = "This is effectively an internal beaker. It will consume and produce power from phoron, slime jelly, welding fuel, carbon,\
-	 ethanol, nutriments and blood, in order of decreasing efficiency. It will consume fuel only if the battery can take more energy."
+	 ethanol, nutriments, in order of decreasing efficiency. It will consume fuel only if the battery can take more energy. Blood is also able to be used as \
```
TheKnightofAura, line 112: Replace the comma between Ethanol and Nutriments with and instead.

PR #4151, opened 2022-06-02
makes ghost tags better

silicons:
appearance flags --> don't inherit parent bullshit


silicons:
ok can't do this because resizes
ugh

PR #4160, opened 2022-06-05
The Lythios Facility Emerges

CharlesWedge:
## Nanotrasen Employee Annoucement
As you are all probably aware the NSB Atlas was only recently required by Nanotrasen. Its acquisition from the People's Republic of Adhomai came after a disasterous reactor failure at the original facility. As a result the Facility was sold to Nanotrasen and quickly declared an exclusion zone. Recently however the facility has once again emerged form the snow North West of the Science outpost. It goes without saying that the site is still currently heavily irradiated and should be avoided in all cost. All expeditions into the region should only be conducted with extreme radiation protection and lower levels where radiation is most extreme should be avoided entirely. 

## Changelog
:cl:
add: Lythios Disaster site at the North West Corner of the Wester plains. A multilevel radioactive disaster site.
add: 3rd Extreme Radiation Emitter, since 100 Bqs is not enough to simulate an Elephant's foot.
/:cl:


Captain277:
Not sure what happened but this PR broke hardcore.

Zandario:
loading ../../_maps/map_files/rift/rift-09-west_caves.dmm
error: '[' should be escaped (\[) or followed by ']'

```
448 /obj/item/paper{
449 	info = "[center]Check the vault, they couldn't get it out without the Commonwealth finding out. Careful they booby trapped the lower levels. [br][i]- Colonel Lii'jar. [/i][/center]";
450 	name = "fancy note"
451 	},
```

PR #4198, opened 2022-06-14
refactors /Life() a little into Physical and Biological life ticks

silicons:
## About The Pull Request

doesn't do this fully
see: i didn't entirely refactor every process especially biological metabolism and stasis because that'd take forever
in the future we should heavily look into having stasis/whatnot handled at the root, perhaps using signals/additives for more than one stasis effect (status effects soon!).

## Why It's Good For The Game

more granular way to handle life ticks so we can hopefully unfuck this utter mess in the future.

## Changelog
:cl:
refactor: Life(), a little.
/:cl:


Zandario (approved):
Finally...

PR #4203, opened 2022-06-15
Fighters Collision Damage

racobio:
## About The Pull Request
Causes a flat amount of damage when fighters collide with objects.

## Why It's Good For The Game
Adds some realism to piloting fighters. Flying a fighter indoors is now properly dangerous to the fighter, unless the pilot is sufficiently skilled as to not collide with objects.

## Changelog
:cl:
add: Added collision damage to fighters
add: Added alert message on fighter collision
/:cl:

### DIFF
--- a/code/game/mecha/combat/fighter.dm
+++ b/code/game/mecha/combat/fighter.dm
@@ -222,6 +222,13 @@
 	else
 		who << sound('sound/mecha/fighter_entered.ogg',volume=50)
 
+//causes damage when running into objects
+/obj/mecha/combat/fighter/Bump(var/atom/obstacle)
+	if(istype(obstacle, /obj))
+		occupant_message("Collision Alert!")
+		take_damage(25, "brute")
+		playsound(src, 'sound/effects/grillehit.ogg', 50, 1)
+
 ////////////// Gunpod //////////////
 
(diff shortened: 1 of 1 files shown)


code/game/mecha/combat/fighter.dm
```diff
@@ -222,6 +222,13 @@
 	else
 		who << sound('sound/mecha/fighter_entered.ogg',volume=50)
 
+//causes damage when running into objects
+/obj/mecha/combat/fighter/Bump(var/atom/obstacle)
```
silicons, line 226: no var/, also needs ot . = ..()

also this check .. it WAS tested yes? this doesn't work on turfs. furthermore the damage is pretty high.

code/game/mecha/combat/fighter.dm
```diff
@@ -222,6 +222,13 @@
...
+	if(istype(obstacle, /obj))
+		occupant_message("Collision Alert!")
```
silicons, line 228: visible message and make this obvious.

code/game/mecha/combat/fighter.dm
racobio: it was taking damage when running into turfs as well as objects. dunno how, but it was.

PR #4206, opened 2022-06-16
Fixes nutriment causing puking

Zeldazackman:
Hopefully it should prevent nutriment from causing people to puke.

It was apparently never listed for a proper reagent ID.

Shamelessly swiped from verks' fix.


deathride58:
One digit away from being #42069

PR #4207, opened 2022-06-16
adds an unit test for reagent ID uniqueness

silicons:
"oh no, everyone's vomitting again"


Zandario (approved):
Don't see any fuckups

PR #4266, opened 2022-07-15
reworks defibs, cpr, brain decay, not so slight organ processing refactor

silicons:
## About The Pull Request

- CPR only doable once every 3-7 seconds, no more healing oxyloss, instead it provides "manual breath" trait that makes it as if someone wasn't in crit (gains instead of loses oxy from respiration. bloodloss/etc can still give oxy dam)
- CPR stabilization of breath works for 10 seconds
- CPR stabilization of brain damage works for 15 seconds. Brain damage doesn't go up, it gets stabilized.
- CPR forces one reagent tick always regardless of death status.
- brains damage while someone's dead. I'd like for it to be all organs damage but .. let's not give medical too much of a pain yet
- defibs no longer do brain damage, this was stupid and is causing issues.
- defib sickness is removed because the way it's doing it is literally adding to the problem right now and isn't on standards at all. It'll be readded at a later date with more sane implementations/values.
- defibs now work on logged out people, a la CM. 

## Why It's Good For The Game

current revival meta is literally broken, let's fix it

## Changelog
:cl:
add: CPR now has a use! It will stabilize people who are in crit properly and prevent brain decay while dead. Furthermore, people being CPR'd slowly metabolize reagents, even when dead.
del: Defib sickness removed until rework
refactor: organs have been refactored a bit
balance: organs now decay while dead
/:cl:


silicons:
update:

lol i don't get a break

organ refactor time.

PR #4270, opened 2022-07-20
removes seed_rand from random maps 

silicons:


code/modules/random_map/random_map.dm
```diff
@@ -31,6 +31,7 @@ var/global/list/map_count = list()
 	// Test to see if rand_seed() can be used reliably.
 	var/priority_process
 
+// SEED DOES NOTHING
```
deathride58, line 34: contraception be like

silicons: :death:

PR #4274, opened 2022-07-25
attempts to separate tgui store from other servers

silicons:
noticing chat history being shared across codebases when it shouldn't, this _may_ fix it. testmerge first, please.


tgui/packages/common/storage.js
```diff
@@ -10,9 +10,9 @@ export const IMPL_MEMORY = 0;
 export const IMPL_LOCAL_STORAGE = 1;
 export const IMPL_INDEXED_DB = 2;
 
-const INDEXED_DB_VERSION = 1;
+const INDEXED_DB_VERSION = 2;
```
LetterN, line 13: no need to bump this

LetterN (changes requested):

PR #4285, opened 2022-07-29
finally removes tgui build artifacts from the codebase and forces build script usage

silicons:
i do not care how lazy people are
it is an ONE LINER to do this

in exchange, we can stop merge conflicting on tgui and having weird desync situations.


silicons:
MAKE SURE TO TESTMERGE THIS ONCE FIRST.
We need to make sure the server is building tgui properly.
I already checked the scripts but the best way to test is a live test.

PR #4317, opened 2022-08-03
fixes scientists being able to materialize infinite energy swords out of thin air

silicons:


code/game/atoms_movable.dm
```diff
@@ -264,7 +264,7 @@
 
 //Called when touching a blood pool.
 /atom/movable/proc/blood_act()
-	blood_act(null, 500, 50)
+	// blood_act(null, 500, 50)
```
silicons, line 267: im just removing it because features that result in an infinite loop get removed

PR #4257, opened 2022-07-08
Removes all nurses, the spider queen, and spiderlings from the cave

yourdoom9898:
## About The Pull Request

Exactly what it says on the tin

## Why It's Good For The Game

Having a cave nearby with things for Security to shoot at when bored is fine and dandy, but allowing this threat to be able to multiply to a constant threat to a valuable research facility and it's staff is utterly unacceptable, and there is no logical reason why NanoTrasen wouldn't have completely cleared out this cave, given it's proximity to the station. I find neutering their growth a acceptable compromise

## Changelog
:cl:
del: Removed a bunch of spiders, and the spider cave's exponential growth
/:cl:

PR #4340, opened 2022-08-08
Removes all nurses, carriers, the spider queen, and egg piles from the outside cave

yourdoom9898:
## About The Pull Request

See #4257 

## Why It's Good For The Game

This cave is the absolute barest form of "content" possible, and having constant incursions in the same area players spawn in the game to is outright unacceptable. I would remove the cave entirely given much more interesting POIs already exist nearby the station, but too many people are complaining, so neutering the absolutely insane growth of the cave will have to do.

### DIFF
--- a/_maps/map_files/rift/rift-06-surface3.dmm
+++ b/_maps/map_files/rift/rift-06-surface3.dmm
@@ -11335,10 +11335,6 @@
 /obj/random/trash,
 /turf/simulated/floor/plating,
 /area/maintenance/commandmaint)
-"uw" = (
-/obj/effect/spider/spiderling/virgo,
-/turf/simulated/floor/outdoors/safeice/lythios43c/indoors,
-/area/rift/surfacebase/outside/outside3)
 "ux" = (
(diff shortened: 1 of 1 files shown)


FreeStylaLT:
I'm gonna leave a local server running for about 8-10 hours to see if spiders can actually wander into any parts that'd result in them attacking arrivals.

If that doesn't happen (in the span of roughly 2 rounds) I can guess any incursions happened because other players _led_ the spiders out into maint or arrivals, which isn't exactly an unwelcome thing, barring obvious grief.

FreeStylaLT:
image
damn one wandered out in 1:39 hours

FreeStylaLT:
image
lmao one ended up in the shuttle after 3:23 hours

silicons:
already discussed this
move the thing somewhere else if you don't want to map anything new
don't kill the reason it's dangerous that's lower than no effort

yourdoom9898:
Lower than no effort would be leaving this dogshit """POI""" as is, and mapping it somewhere else implies I even wanted to keep it around in the first place

PR #4389, opened 2022-08-29
fixes rig/dufflebag spawning issues

Captain277:
1. *Swaps RIG and Duffle Bag names in interface.*

1. _This was so easy to do, there's no excuse for the bug taking months to fix. All you had to do was look in-game, see the words "Duffle Bag" and "RIG" were swapped. Then you just had to search for the phrase "Duffle Bag". It's literally right at the bottom of the results. You can see where the two words are swapped in the load order. Then you just swap the two words. That's it. That's how easy it was to fix the bug you've all been pinging about for months. Unacceptable._

## Changelog
:cl:
fix: Fixes bad naming on RIG and Duffle Bag selections.
/:cl:


nik707:
lol

PR #4399, opened 2022-09-07
Chlorine Cosmetics, Courser Corrections, and Creepy Coronas

Captain277:
## About The Pull Request

1. **Adds a new icon state for Chlorine gas.**
2. **Adds a hallucinogenic/emp aura to Horrors.**
3. **Corrects Courser dir.**

## Why It's Good For The Game

1. _Mostly a cosmetic change. I think it'd be cool to "see" chlorine gas filling a room._
2. _A mechanical addition I've long had planned, this would make aberrations and other spooky mobs a little more accurate to the lore._
3. _The Courser has been unable to fly due to an improper definition of its facing direction. I don't know how long this has been going on because it was never reported. It's now fixed._

## Changelog
:cl:
add: Adds cosmetic chlorine gas icon.
add: Adds horror aura for...horrors.
fix: Fixes incorrect Courser dir.
/:cl:


code/datums/components/horror_aura.dm
```diff
@@ -0,0 +1,10 @@
+/datum/component/horror_aura/Initialize()
+	if(ismob(parent))
+		RegisterSignal(parent, .proc/aura_effect)
```
silicons, line 3: you don't actually need a signal
a signal means you want to listen to something 
for you, you realistically just need to 
START_PROCESSING(SSobj, src)

code/datums/components/horror_aura.dm
```diff
@@ -0,0 +1,10 @@
...
+
+/datum/component/horror_aura/proc/aura_effect()
```
silicons, line 5: then call aura_effect() somewhere from /datum/component/horror_aura/process()

code/datums/components/horror_aura.dm
```diff
@@ -0,0 +1,10 @@
+/datum/component/horror_aura/Initialize()
```
silicons, line 1: file should start with a description of component and a /datum/component/horror_aura for neatness
generally defining a path by defining procs without defining the actual path is bad practice

code/datums/components/horror_aura.dm
```diff
@@ -0,0 +1,10 @@
...
+	if(ismob(parent))
+		RegisterSignal(parent, .proc/aura_effect)
+
+/datum/component/horror_aura/proc/aura_effect()
+	var/atom/movable/AM = parent
+	for(var/mob/living/carbon/human/H in view(7,src))
```
silicons, line 7: if you want to be even fancier you could make this range variable by making it a variable and having /Initialize() be /Initialize(radius) --> src.radius = radius.

code/datums/components/horror_aura.dm
silicons: for the Initialize(), you want to 
1. check . = ..() for COMPONENT_INCOMPATIBLE, if so, return at the top
2. check if the parent is an atom/movable. if not, return COMPONENT_INCOMPATIBLE, because it's pointless to attach this component to a non-atom/movable.

code/datums/components/horror_aura.dm
Captain277: Allrighty. Think I did everything. Aura still isn't triggering, however.

PR #4490, opened 2022-10-05
Miscellaneous Atlas Map Fixes Vol. 1

AlphaM01:
## About The Pull Request
A collection of small fixes for the Rift Map.

## Why It's Good For The Game
Makes the map slightly better in many small ways, details in the changelog.

## Changelog
:cl:
tweak: Changed Atmos' Phoron tank hazard sign to correctly call the gas "Phoron" instead of "Plasma".
tweak: Renamed the doors inside Centcomm's Medbay to accurately say what they are instead of all being the same copied "Operating Theatre" door.
fix: Split the Memorial monument in Centcomm into two different objects, each occupying their own tile to prevent phasing through it.
fix: Anchored the memorial room's statues in place.
fix: Moved Surface 1 airlock buttons to the outsides of the airlock, instead of inside the doors, to prevent them being interactable from the wrong side of the doors.
add: Connected the central heating area of the Emergency Bunker to the power grid by adding a missing section of wire under a door.
add: Added plating to the underside of the catwalks on the Mining Outpost to prevent situations where it was possible to fall through the catwalk because of the empty space it was on (in a few spots leading to insta-death).
del: Deleted outdated/floating path edge overlays.
/:cl:


ktoma36:
I would say the catwalks were intentional but I myself actually just fell off the from the surface level. Though I think its more to do with the funky railings I put on this corner
image

AlphaM01:
I don't know what the precise mechanism causing it in the code is, but I know that it seems to happen if you move diagonally between ground w/ no catwalk directly onto catwalks over empty space. This doesn't happen on the bridge because it has the spots with plating at the ends, and doesn't happen on the landing pads (anymore) because I already added plating to them after having this same issue while I was building them.
The spot you sent is one of the deadly and probably most phased through ones, but it doesn't happen only there. Furthermore, it also no longer happens with the plating under that tile. I've tested the fix in local and could walk diagonally back and forth onto the snow without falling, where as you will if you try that currently. So I it's probably not the railing's fault.

ktoma36:
Did some testing and yeah... its definably a thing and it only happens if you're going from catwalks with open turf underneath them and catwalks with a different turf underneath them

AlphaM01:
I have
1) Mis-named one of the Centcomm doors.
2) The ability to throw in the pipe fix w/ the rest of this stuff.
However, I don't know how this TM stuff works so I'm not touching anything for now.

PR #4522, opened 2022-10-11
Removes Nanopaste Self-Heal from Cyborgs

Captain277:
## About The Pull Request

1. **Removes Nanopaste Self Heal from Cyborgs.**

## Why It's Good For The Game

1. _Cyborgs retain their access to Nanopaste stacks, but can no longer heal themselves._

## Changelog
:cl:
balance: Adds a check to prevent self-healing with nanopaste modules.
/:cl:

### DIFF
--- a/code/game/objects/items/stacks/nanopaste.dm
+++ b/code/game/objects/items/stacks/nanopaste.dm
@@ -16,6 +16,9 @@
 /obj/item/stack/nanopaste/attack(mob/living/M as mob, mob/user as mob)
 	if (!istype(M) || !istype(user))
 		return 0
+	if (istype(M,/mob/living/silicon/robot) && istype(user))
+		to_chat(user, "<span class='notice'>You are unable to apply this paste to yourself.</span>")
+		return 0
 	if (istype(M,/mob/living/silicon/robot))	//Repairing cyborgs
 		var/mob/living/silicon/robot/R = M
 		if (R.getBruteLoss() || R.getFireLoss())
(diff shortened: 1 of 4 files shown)


yeayea130:
Instead of this. Fix the ability to infinitely use nanopaste without having any Nanopaste supply left.

It's fucking broken. I've reported it to multiple people as I have no idea how to fix it.

Captain277:
> Instead of this. Fix the ability to infinitely use nanopaste without having any Nanopaste supply left.
> 
> It's fucking broken. I've reported it to multiple people as I have no idea how to fix it.

Yeah, on this topic. The initial band-aid has been removed, as was intended. Upon further experimentation, charge cost doesn't seem to be factoring into the consumption at all. This means the issue can be traced back to /obj/item/stack/proc/use() itself. There's some curious math at play re: synth charge costs. I think this issue might extend across the board to other medications and cables. I'm going to have to source input from various cyborgs to confirm how widespread the issue is - I think it's all stacks, probably. In the meantime, this adds a check on Nanopaste that prevents self-application if the user's a cyborg. I'll be happy to remove that check once we fix this charge cost issue, which hopefully won't take long.

code/game/objects/items/stacks/nanopaste.dm
```diff
@@ -16,6 +16,9 @@
 /obj/item/stack/nanopaste/attack(mob/living/M as mob, mob/user as mob)
 	if (!istype(M) || !istype(user))
 		return 0
+	if (istype(M,/mob/living/silicon/robot) && istype(user))
```
TheLordME, line 19: something like `if(M == user)` would be better suited here
currently you are verifying user is a mob (which is done in the if in line 16 as well AND checking of your target is a robot/borg

TheLordME (changes requested):

yeayea130:
please check #4523. likely fixes exploit. needs testing. probably. can't test myself as currently going out to get groceries!

PR #4533, opened 2022-10-13
Updates DreamAnnotate to v2

deathride58:
Github deprecated Node12 for workflows, which DreamAnnotate V1 uses. DreamAnnotate V2 uses Node16 instead. This PR isn't urgent, but shouldn't affect much in theory.


Zandario:
Just read the details and yeah, better sooner than later.

PR #4566, opened 2022-10-18
Doubles Announce Time for Vines

Captain277:
## About The Pull Request

1. **Increases Vine Announce Time From 1 Min to 2 Min.**

## Why It's Good For The Game

1. _Vines often get caught before they can really spread far enough. This adds one minute to the announce timer to give the vines a better chance. May drop if necessary._

## Changelog
:cl:
tweak: Adds 1min to vine announce timer.
/:cl:

### DIFF
--- a/code/modules/events/spacevine.dm
+++ b/code/modules/events/spacevine.dm
@@ -1,7 +1,7 @@
 /var/global/spacevines_spawned = 0
 
 /datum/event/spacevine
-	announceWhen = 60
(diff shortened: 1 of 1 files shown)


Zandario:
For those who aren't informed, something that's planned is to have the timer (and other things) dynamically scale depending on the number of active people, this is a temp solution to give vines a minor buff without changing their stats.

PR #4620, opened 2022-10-24
Spider Cave Change

MediHound:
## About The Pull Request

Turns the Spider Cave into an opt-in dangerous area.

## Why It's Good For The Game

People shouldn't need to select someone at roundstart to venture into the cave right next to the landing pads to block it off, out of fear a spider might walk out of the cave and break in at arrivals.
As such, the previous closer part of the cave has been cleared out, and people can explore this part to get some trash loot.
the actual spider cave was expanded up to the north, with an entrance at the north, together with a warning sign and spiderwebs, making it clear that people are entering danger. This turns the spider cave into an opt-in dangerous area and reduces the likelyhod of a spider accidently walking out.
At the end of the spider cave are multiple random item spawns, as a reward for those who got that far.

## Changelog
:cl:
tweak: Changed the spider cave
/:cl:


TheLordME:
how do you ensure the spiders stay in their cave?

MediHound:
> how do you ensure the spiders stay in their cave?

I don't. I just change it so that if a spider does end up walking outside, they are further away from the station.
Honestly did not want to put down a door.

This is what it's looking like:
GQhzywu

PR #4624, opened 2022-10-25
icon_exists

Zandario:
## About The Pull Request
Proc: icon_exists
Checks if the given iconstate exists in the given file, caching the result. Setting scream to TRUE will print a stack trace ONCE.

## Changelog
:cl: Zandario, Kapu1178
code: Added a new proc: **icon_exists** - It checks if a given icon_state exist and can scream bloody murder if it doesn't.
/:cl:


Zandario:
> question is this going to be used for rendering i'm not sure how list hashing works with vv'd icons

Uhh here's what Kapu said:
> If the implication is "does this work if i modify the icon file mid round" the answer is "it would always fail if you looked up the icon_state before adding it" and my follow up question would be "why are you modifying icon files mid round"

But yeah his uses for this I mainly saw inside of carbon icon gen, so you wouldn't really use this for just normal icons and such but you can use it as a temp dev tool as well to see if icons are creating correctly

Zandario:
I didn't notice I put this up as a draft 💀

PR #4625, opened 2022-10-25
Creates Catborgs

Captain277:
## About The Pull Request

1. **Adds Catborgs.**
2. **Renames Dogborg Selections.**

## Why It's Good For The Game

1. _I was commissioned to do some catborg sprites. As discussion went on, the Dogborg infrastructure (namely resting and wreckage) was selected. All in all, I think the shells are cute - and not just because I spent all day making them. Versatility of cyborg shells and designs would almost certainly be common in the future. Concerns about "other" potential issues regarding implementation here should be discussed as part of larger Server Rules precedent._
2. _Now that it isn't just canines, I changed the names to make more sense. Dogborg code is a total mess, and I'm probably gonna go through it later and shuffle things around to make it less of a nightmare. Maybe. That's not really the scope of this project._

## Changelog
:cl:
add: Adds catborgs.
tweak: Changes Hound naming selections.
/:cl:


yeayea130:
For future reference. if you're going to be going into dogborg code and messing with it. Involve me. I need to learn and tagging along with someone is the best way to do it.

Captain277:
> For future reference. if you're going to be going into dogborg code and messing with it. Involve me. I need to learn and tagging along with someone is the best way to do it.

I'm gonna be refactoring Dogborgs here, ultimately. There's probably gonna be a lot of repathing over the next couple days too. Gonna try and bring them into alignment with normal cyborgs, like I always say I will.

PR #4635, opened 2022-10-27
How Did I forget to add the client var for view-data

Zandario:
## About The Pull Request
Apparently when I updated view_data I forgot the most important part, the client var.


nevimer:
very important, we can finally get rid of some old procs :D

PR #4671, opened 2022-11-02
[WIP] Miscellaneous Atlas Map Fixes Vol. 2

AlphaM01:
## About The Pull Request

This contains a couple more mapping fixes, including improvements upon some of the changes from Vol. 1 that needed them.
Particularly, unbandaids the phoron sign fix, and re-does the airlock changes, but more consistently and with interactable buttons where people expect them.
The two entirely new additions are, a split of the "memetic" engineering hazard sign into "infohazard" which uses the memetic sprite, and "cognitohazard" which uses a new sprite but performs the function of the old memetic sign, just more accurately (and dare I say more effectively, since the new sprite should make the warning a lot clearer to the end user compared to a lightbulb!).
Also fixes the space-hole in atmospherics Papalus found while I was making the PR.

## Why It's Good For The Game

Small improvements and bugfixes to Rift good.
The airlock change in Vol. 1 was controversial, but I've put the feedback to use on this new iteration.

## Changelog
:cl:
add: Added new warning sign: "Cognitohazard", replaces memetic hazard warning on Supermatter areas.
tweak: Repurposed "Memetic" hazard sign into "Infohazard" (currently unused)
tweak: Remaps airlocks to now have double access buttons, one on the outside, one on the inside, additionally, somewhat standardizes the placements for them.
fix: Un-bandaids the Phoron sign from Vol. 1 as an edited instance to now being called "Phoron" in the code itself.
fix: Plates over the space-hole in atmospherics.
/:cl:


AlphaM01:
Additionally, a small report on local testing of airlocks:
 - Two buttons seems to work as expected, all buttons including the panel and sensor were tested.
 - Both pre and post-remap airlocks can have both sides opened sometimes if you cycle out, then run in and hit the cycle in button fast enough. I've no idea why they do this. This isn't too catastrophic as hitting any of the buttons makes it fix itself, but will vent areas if left open.
 - Some of the new airlocks which are hooked to air tanks get stuck trying to cycle in once the tank runs out.
 - The science outpost airlock is mapped differently, it's sensor does not act as a button, and it may be possible to re-close both doors by hitting the button for the side that's open, it doesn't seem to always happen.

ktoma36:
I can give some insight to the science outpost airlocks. Those ones have all their variables set manually instead of using map helpers, they're pretty old in that regard.

PR #4674, opened 2022-11-03
fixes falling for good; removes dropnom

silicons:
refactors falling + mz movement to use a new system.

removes dropnom because no one uses it and i don't feel like maintaining vorecode that no one uses anymore in drop code when i need to rewrite drop code partially because of its existence


ktoma36:
Nooooo my plan to direct all people who fall from the sky into one giant slime mob is ruined now!!! In all seriousness though I've personally only seen one happen once with someone pulling the RD's slime pet down those stairs in tether. I doubt anyone will miss it

code/modules/multiz/movement.dm
```diff
@@ -38,12 +38,9 @@
...
 
-	if(!start.CanZPass(src, direction))
-		to_chat(src, "<span class='warning'>\The [start] is in the way.</span>")
-		return 0
-
-	if(!destination.CanZPass(src, direction))
-		to_chat(src, "<span class='warning'>\The [destination] blocks your way.</span>")
+	var/atom/obstructing = start.z_exit_obstruction(src, direction)
+	if(obstructing)
+		to_chat(src, "<span class='warning'>\The [obstructing] is in the way.</span>")
```
Zandario, line 43:
```suggestion
		to_chat(src, SPAN_WARNING("\The [obstructing] is in the way."))
```
You touch it I correct it: :trollface:

Zandario (dismissed):

Zandario (approved):
Based

PR #4679, opened 2022-11-05
Tooltips Galore!

Zandario:
## About The Pull Request
Started a small idea to "Lets tooltip it all :)"

Also took this time to get rid of the old admins global list so we just have GLOB.admins

## TL;DR
Cool tooltips so I can stop having to explain emoji and admin commands.

Also a micro optimization for the span macros using tips from LummoxJR

## Changelog
:cl:
add: Emojis now show tooltips for what command it was, just like discord!
code: SPAN_ Macros now use the "[a][b][c]" method instead of "a"+"b"+"c", though that only saves minimal time in compile.
code: Added a macro for Blockquotes
code: Unified our debug_log into the correct chat categories.
/:cl:


Zandario:
For some reason it isn't removing the Merge Conflict tag

TheLordME (approved):

PR #4687, opened 2022-11-06
[WIP] [DNM] More vehicles ported, vehicle tweaks

ktoma36:
## About The Pull Request

This is my ongoing project to port vehicles over from the old system as well as turn a few things into 'vehicles'.  Main one right now is the wheelchair, list will expand if I find more I think would be good to convert

## Why It's Good For The Game

Old vehicle system sucks, new good. 

## Changelog
:cl:
add: More vehicles to new system
imageadd: Thank you Pawn for new Tug Icons 
code: Fiddling with vehicle code
/:cl:


code/modules/vehicles/ridden/wheelchair.dm
```diff
@@ -59,7 +91,7 @@
 		R.color = src.color
 		qdel(src)
 
-/obj/vehicle/ridden/wheelchair/OnMouseDropLegacy(over_object, src_location, over_location)
```
TheLordME, line 62: Is this intended to overwrite the Legacy mouse drop?
I know silicons recently added the legacy exact

code/modules/vehicles/ridden/tug.dm
```diff
@@ -0,0 +1,47 @@
+/obj/vehicle/ridden/tug
```
TheLordME, line 1: doesnt this file need adding to the .dme?
You dont need to add .dmi filee, but this one is a .dm file

code/modules/vehicles/ridden/tug.dm
```diff
@@ -0,0 +1,47 @@
...
+
+
+/datum/component/riding_handler/vehicle/ridden/tug
+	vehicle_move_delay = 3
+	allowed_turf_types = list(
+		/turf/simulated,
+		/turf/unsimulated
+	)
+
+/datum/component/riding_handler/vehicle/ridden/tug
```
TheLordME, line 18: this line shouldnt be needed, you literally define the datum a part above

TheLordME (changes requested):

code/modules/vehicles/ridden/wheelchair.dm
ktoma36: Oh, I saw that there was OnMouseDropLegacy and OnMouseDrop and thought OnMouseDrop was the more recent proc. OnMouseDrop in game works pretty much the same from what Ive tested

code/modules/vehicles/ridden/tug.dm
ktoma36: Was having merge conflicts earlier and took it off the .dme list temporarily, thank you for reminding me to put it back on XD

code/modules/vehicles/ridden/tug.dm
ktoma36: Fair enough, I was simply following how silicons was organizing them (in the boat one. ) IMO it helps keep it from being a bit of a mess to look at, does it harm it at all to do it like this?

TheLordME: you can always slap a comment here if you want a visual seperation
I am not sure if it has a noticeable impact, but it might increase compile time
and imo its confusing rather than helping with readability

code/modules/vehicles/ridden/wheelchair.dm
TheLordME: Legacy is depricated use non legacy

ktoma36:
Will visit this again in the future but consider this legitimate salvage if someone else wants to use this

PR #4708, opened 2022-11-12
overhauls radiation

silicons:
:cl: 
refactor: radiation now uses a wave system
add: radiation can now contaminate items
add: geigers can now detect radiation contamination, and work from range
add: showers can decontaminate radioactive items
tweak: whole-level irradiation has been reworked; it'll probably end horribly. you have been warned.
balance: radiation damage and armor heavily reworked
/:cl:


code/__DEFINES/traits/_helpers.dm
```diff
@@ -0,0 +1,69 @@
+
+// trait accessor defines
+#define ADD_TRAIT(target, trait, source) \
```
TheLordME, line 3: the \ is to keep it as a macro?

TheLordME (dismissed):
Lets be real, its a silicons PR, if they leave the server dies, and if they make a mistake they fix it themselves

code/__DEFINES/traits/_helpers.dm
Zandario: It's the newline indicator? It's **required**.

Zandario (approved):
I agree to the terms of service

PR #4716, opened 2022-11-16
Removes Holster Selection from Loadout

Captain277:
## About The Pull Request

1. *Removes Holster Selection from Loadout.*
2. *Adds a Holster to the Pathfinder's locker.*

## Why It's Good For The Game

1. _I have general issues with the existence of this listing to begin with. However, there is a bug in the nature of Selection that causes the machete scabbard to select its longest path, which in this case is a pre-loaded Durasteel machete. This fault, paired with my general dislike of this item being made available in the loadout, has led to this outcome._
2. _Consequently, I have given the Pathfinder a spare holster in their locker, as this was apparently the most frequent use of this loadout option. This is how we should address holster shortcomings - not by giving them out to anyone in a role restricted slot. I will accept further requests to put holsters in lockers within reason._

## Changelog
:cl:
add: Adds a holster to the PF's locker.
del: Removes Holsters from Loadout Selection.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/explorer.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/explorer.dm
@@ -152,6 +152,7 @@
 		/obj/item/ammo_magazine/m57x28mm,
 		/obj/item/ammo_magazine/m57x28mm,
 		/obj/item/clothing/accessory/holster/machete,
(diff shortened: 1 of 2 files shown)


TheLordME (approved):

TheLordME:
When are we adding a holster for the RPD and RCD?

Zandario (approved):

PR #4726, opened 2022-11-16
fixes prosthetic limb customization

silicons:
refer to code comment for how this works

this is shitcode
i accept it's shitcode
and there's jack shit we can do about it because frankly fuck organ code.

refactor coming in the next few months.


TheLordME (approved):

Zandario (approved):
Don't we just love organ code? :)

PR #4752, opened 2022-11-22
adds a donator item

MalricB:
adds a backpack for Jackson#4111, replaces their old donator item (funny hat)


TheLordME (dismissed):

code/modules/preferences/preference_setup/loadout/loadout_donator.dm
```diff
@@ -152,8 +152,8 @@
 	ckeywhitelist = list("certifiedhyena")
 
 /datum/gear/donator/woolhat
```
Zandario, line 154: Shouldn't this path be changed

MalricB: yup ty

TheLordME (approved):
Is it working now?

PR #4759, opened 2022-11-23
[idb ignore] completely migrates sprite accessory to id system

silicons:
Shit like this is why I hate everyone.

tl;dr: Since everything is stored by type, I can't do the organization pass I need to do.

So I need to migrate every single savefile on the disk at once.


Zandario (approved):
I've read the terms of service

PR #4763, opened 2022-11-24
fixes some robolimb stuff

silicons:
same issue as before, only properly fixed

tl;dr lets you robotize limbs forcefully

yes this is shitcode
no i don't care because i can't do it better without refactoring organs.
i'm refactoring organs next month
fuck this shit
so done with it

### DIFF
--- a/code/modules/organs/external/standard.dm
+++ b/code/modules/organs/external/standard.dm
@@ -268,8 +268,8 @@
 		cannot_gib = FALSE
 	return ..()
 
-/obj/item/organ/external/head/robotize(company, skip_prosthetics, keep_organs)
-	return ..(company, skip_prosthetics, 1)
+/obj/item/organ/external/head/robotize(company, skip_prosthetics, keep_organs, force)
+	return ..(company, skip_prosthetics, 1, force)
 
 /obj/item/organ/external/head/removed()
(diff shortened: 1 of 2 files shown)


code/modules/organs/external/standard.dm
```diff
@@ -268,8 +268,8 @@
 		cannot_gib = FALSE
 	return ..()
 
-/obj/item/organ/external/head/robotize(company, skip_prosthetics, keep_organs)
-	return ..(company, skip_prosthetics, 1)
+/obj/item/organ/external/head/robotize(company, skip_prosthetics, keep_organs, force)
```
TheLordME, line 271:
```suggestion
/obj/item/organ/external/head/robotize(company, skip_prosthetics, keep_organs = TRUE, force)
```

code/modules/organs/external/standard.dm
```diff
@@ -268,8 +268,8 @@
...
+	return ..(company, skip_prosthetics, 1, force)
```
TheLordME, line 272:
```suggestion
	return ..(company, skip_prosthetics, keep_organs, force)
```

TheLordME (approved):
Wouldnt it make more sense like this?

PR #4765, opened 2022-11-24
fixes beakers eject when hands are full

silicons:


code/game/machinery/_machinery.dm
```diff
@@ -525,7 +525,11 @@
  */
 /obj/machinery/proc/try_put_in_hand(obj/object, mob/living/user)
 	if(!issilicon(user) && in_range(src, user))
-		user.put_in_hands(object)
+		if(!user.put_in_hands(object))
```
TheLordME, line 528: Why not do the same as with the paper deletion?
```suggestion
		user.put_in_hands_or_drop(object))
```

TheLordME (changes requested):

Zandario (approved):

TheLordME (approved):

PR #4771, opened 2022-11-25
[Dependabot] Dependencies Testing/Staging

Zandario:
I may regret this.
## About The Pull Request
Basically added a workflow (`.github/workflows/update-dependabot`) that updates the targets for dependabot.

The default behavior has also changed. Dependabot will now target a `dependencies` branch that will be automatically updated by `.github/workflows/rebase_dependencies.yml`. This will allow us to test dep updates without fucking master up. 

tl;dr
**WEBDEV PIPELINE CARGO-CULT MOMENT**


Zandario:
Not worth the overhead, I'm fine with reviewing changes per pr as I currently do

PR #4778, opened 2022-11-28
Github Workflow Updates

Zandario:
# Size-based PR Labeling!

I made two seperate categories, one for code and one for maps so they don't impact each other!

I also removed `dev-atlas` from the workflow whitelist as that branch hasn't existed in about half a year.


Zandario:
Size-Based Labeling failing can be ignored as new workflows being added in a PR will use read permissions, only getting write access once inside the repo.

silicons (approved):

PR #4815, opened 2022-12-09
new radiation but with a potentially better solver

silicons:
tl;dr

uses a pseudo-raycasting system
as opposed to cones

the cost per tile goes up
but the cost per *radiation pulse* theoretically goes down, because in theory, it's easier to block.


silicons:
works but not fast enough

PR #4831, opened 2022-12-11
emissives

silicons:
tin

tl;dr soft lighting from surfaces

now featuring renderers that can layer with the atom for later usage in humans

(DO NOT USE MY RENDERER SYSTEM; use tg's for basic shit like APCs, you DO NOT NEED something as expensive as that for most basic purposes!)


TheLordME (approved):

TheLordME (approved):
yolo

PR #4837, opened 2022-12-12
renames flags to atom_flags for, well, atoms; transitions some reasonable atom flags to be more, well, obvious

silicons:
basically 
since byond has no concept of namespaces (i hate byond),
we're going to start renaming flags to imply what they're behind so when someone sees flags they don't get confused.


TheLordME (approved):
yolo

PR #4839, opened 2022-12-12
SSmaterials - lookup

silicons:
adds SSmaterials
adds lookup for id/path

name lookup is kept because I'm not converting every file at this time

**Please start using the new lookups so we can transition.**
One of us will transition any remaining name lookup usages sometime in the future.


code/modules/materials/materials_vr.dm
```diff
@@ -1,4 +1,5 @@
 /datum/material/flesh
```
TheLordME, line 1: I would kindly ask you to move these /datum/s to materials.dm in compiance with our "Get rid of _v files" policy

TheLordME (changes requested):

TheLordME (approved):

PR #4841, opened 2022-12-13
Maploader

silicons:
1. maps are now loaded at runtime
- the world is actually compiled with nothing in it; SSmapping initializes a space reservation level first and then resizes the world to the given station's size, and starts the init process.

2. /datum/map, /datum/map_level erased; all map data is now .json's
- map-specific behaviors that cannot be generalized will be kicked into /datum/map_module and /datum/level_module, both having hooks to run on-spawn

3. multiz transitions are now auto-generated rather than requiring mappers do the work
- this supports diagonals! it'll be weird, though

4. level linking (up, down, all 4 cardinals) now occurs automatically;
- you can specify this manually in the .json by referencing level IDs
- you can use the new /world_struct system with level ID for an even better experience that allows things like GPS to grab virtual levels/heights

5. planet datums are now lazyloaded and don't use hardcoded z numbers
- you can specify this manually in the .json on the levels
- you can specify this in the .json for entire world_structs

6. the maploader now heavily caches things; init and submap seeding should speed up by a **lot**

7. /_mapload now only tells the system what map id to load; maps are all in /maps now

8. loading new levels has been highly streamlined with the level config system; maps can now include things like centcom/lavaland in their .json and have them automatically load; circular dependencies ARE allowed.

9. fast traits/attributes system
- traits are for boolean values, e.g. is this level part of the main station, etc
- attributes are for any key-value storage, like initial gas mixtures, baseturfs, etc.


silicons:
go away

silicons:
lohikar talked sense into me, time for attempt 3
this pr will stay open for diffs / tracking

PR #4846, opened 2022-12-14
SSao, Z-Mimic, and O7 Lighting

Zandario:
## About The Pull Request
Z-Mimic is nice.

Currently just SSao atm, why SSao? Filter-based ao doesn't really work with Z-Mimic

O7 Lighting cause I got tired of trying to make ours work with this. But hey, we got directional lights now so that's cool.

:cl: @Zandario & @Lohikar
add: SSzmimic
add: O7 Lighting
add: SSao
del: SSopenspace
tweak: Lights now can shine in angles between 0 and 180
/:cl:


code/controllers/subsystem/zmimic.dm
```diff
@@ -0,0 +1,734 @@
...
+
+		// If we previously were MZ_MIMIC_BASETURF, there might be an orphaned proxy.
+		else if (T.mimic_underlay)
+			QDEL_NULL(T.mimic_underlay)
+
+
+		// Handle space parallax & starlight.
+		if (T.below.z_eventually_space)
+			T.z_eventually_space = TRUE
+			t_target = SPACE_PLANE
```
Lohikar, line 260: I'm not certain that this is applicable to this ZM revision, but this probably should be:
```suggestion
			if ((T.below.mz_flags & MZ_MIMIC_OVERWRITE) || T.below.type == /turf/space)
						t_target = SPACE_PLANE
```

iirc this fixes an issue with glass on top of parallax/space.

Linzolle:
is there any way to add a preference to disable seeing any AO at all, currently the preference only respect the old "fake" AO

Zandario:
> is there any way to add a preference to disable seeing any AO at all, currently the preference only respect the old "fake" AO

Not sure why we would put a preference on that, also that would require me to rewrite ZMimic to use plane_masters

Lohikar:
It's probably better to leave the ACES tonemapping for another PR, I'm still tweaking it.

Zandario:
> It's probably better to leave the ACES tonemapping for another PR, I'm still tweaking it.

Fair enough, I'll just comment that stuff out for now for possible future reference

Zandario:
We love Lohikar

PR #4859, opened 2022-12-17
vore now blocks suit sensors

silicons:
makes my life easier when writing sop/corporate regulations.

server policy doesn't revolve around erp, but it helps if the code doesn't make it relevant to the game loop in the first place.


TheLordME (approved):
The comment is a bit odd

PR #4881, opened 2022-12-22
Ports Magazine Loading from Bullets on Floor from Paradise

granodd:
## About The Pull Request
Ports picking up bullets into a magazine from a very old version of Paradise code I had sitting around. It compiles and it works without any immediately obvious issues to me, though I'm not savvy enough to pick up on where the code could be optimized.

## Why It's Good For The Game
• Makes me not want to scream in agony when I accidentally unload 50 rounds from an LMG magazine and then have to put them all back in by hand.
• Also makes cleaning up after donksoft fights much, much easier (the main reason I made this PR)

## Changelog

:cl:
add: clicking on bullets on the floor with an empty or partially empty magazine will transfer them into the magazine. Does not pick up spent ammunition.
/:cl:


granodd:
As far as I know this code works and doesn't break anything else so I'm removing WIP tag.

PR #4882, opened 2022-12-22
organ runtime

silicons:
while this is frowned upon i will allow it because i don't want to force sanitization math everywhere


Zandario (approved):

Zandario:
Fair enough

PR #4930, opened 2023-01-03
reworks gps

silicons:
:cl:
add: new GPS interface
add: GPSes can now store up to 25 waypoints; attack another to transfer them
add: GPS now shows an arrow pointing to waypoint if you are on the same level as it
/:cl:


TheLordME (approved):
no risk no fun
(this has been tmed for a while now)

PR #4948, opened 2023-01-07
SSoverlays tweaks + audit

Lohikar:
Rips out SSoverlays in favor of the O7 one, which supports priority overlays. This should fix issues with AO vanishing, as well as being generally faster.
Also audited all SSoverlay calls, some were strange, invalid, or pointlessly inefficient.


code/controllers/subsystem/overlays.dm
```diff
@@ -217,35 +306,6 @@ SUBSYSTEM_DEF(overlays)
...
-	. = ..()
-	UPDATE_OO_IF_PRESENT
-
-/turf/compile_overlays()
-	. = ..()
-	if (above)
-		update_above()
+/atom
+	var/tmp/list/our_overlays	//! our local copy of (non-priority) overlays without byond magic. Use procs in SSoverlays to manipulate
+	var/tmp/list/priority_overlays	//! overlays that should remain on top and not normally removed when using cut_overlay functions, like c4.
```
Zandario, line 311: Won't this set this instance as the first def of atom

Zandario (dismissed):

code/controllers/subsystem/overlays.dm
Lohikar: Probably, didn't think to check. Can check tomorrow when I'm back at my desktop.

Lohikar: Checked, it's not changing where go-to-definition takes you for `/atom` -- it's already screwed up.

Captain277 (approved):

Lohikar:
This has a serious bug and should not have been merged.

PR #4962, opened 2023-01-11
Mentors

Zandario:
Porting the Mentor system we have on CitMain


code/modules/mentor/mentor.dm
```diff
@@ -0,0 +1,94 @@
...
+	owner = GLOB.directory[ckey]
+	if(owner)
+		owner.mentor_datum = src
+		owner.add_mentor_verbs()
+		if(!check_rights_for(owner, R_ADMIN,0)) // Don't add admins to mentor list.
+			GLOB.mentors += owner
+
+/datum/mentors/proc/CheckMentorHREF(href, href_list)
+	return TRUE
+	/*			Disabled
```
TheLordME, line 35: do you plan on implementing this later?

interface/stylesheet.dm
```diff
@@ -59,6 +59,11 @@ em						{font-style: normal;font-weight: bold;}
 .admin_channel			{color: #9611D4;	font-weight: bold;}
 .event_channel			{color: #cc3399;	font-weight: bold;}
 
+//* Mentor *//
+.mentorsay				{color: #5E2D79;	font-weight: bold;}
```
TheLordME, line 63: indent fuckup?

TheLordME (approved):

code/modules/mentor/mentor.dm
Zandario: CheckAdminHref does the same right now, this will be done in a different PR

Shadow-Quill:
Currently, `mentor follow` is a verb visible to everyone in the right click menu on mobs. (Though clicking it does nothing.)

TheLordME (changes requested):
Multitude of Fixes still required

silicons (commented):
please check the verbs and make the names properly named for rementor/dementor
and make sure mentor_follow is not on every mob

also you should probably be granting the mentor follow/unfollow verbs when someone's like, a mentor, instead of it being on everyone

PR #4965, opened 2023-01-12
Overlays fixes

Lohikar:
- Fixes an issue where SSoverlays was corrupting humantype layer slots.
- Fixes an issue where SSoverlays was corrupting managed overlays.
- Fixes an issue where `copy_overlays()` runtimed if called on an atom without overlays.

### DIFF
--- a/code/controllers/subsystem/overlays.dm
+++ b/code/controllers/subsystem/overlays.dm
@@ -174,6 +174,10 @@ SUBSYSTEM_DEF(overlays)
 /atom/proc/build_appearance_list(atom/new_overlays)
 	var/static/image/appearance_bro = new
 	if (islist(new_overlays))
+		// A lot of code seems to assume that it's safe to pass long-lived lists to SSoverlays.
+		// That was true with the old overlays SS, but is not true of this one.
+		// So we're copying the list. Fix all the callsites if you want to avoid this copy.
+		new_overlays = new_overlays:Copy()
 		listclearnulls(new_overlays)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -46,14 +46,17 @@ GLOBAL_LIST_EMPTY(damage_icon_parts)
 
 //Add an entry to overlays, assuming it exists
 /mob/living/carbon/human/proc/apply_layer(cache_index)
-	if((. = overlays_standing[cache_index]))
-		add_overlay(.)
+	var/overlay_slot = overlays_standing[cache_index]
+	. = overlay_slot
+	if(overlay_slot)
+		// `add_overlay()` will mutate lists you pass it.
+		add_overlay(islist(overlay_slot) ? overlay_slot:Copy() : overlay_slot)
```
TheLordME, line 53: why overlay_slot:Copy() and not overlay_slot.Copy()?

TheLordME (approved):

code/modules/mob/living/carbon/human/update_icons.dm
Lohikar: Per comment on discord, just because the var is untyped -- the var could be made a list, but the slot code is doing type abuse so either way it's unsafe. Might as well be explicit about the type unsafety.

Zandario (approved):

PR #4975, opened 2023-01-15
zmimic ignores observers

silicons:
we don't need to mirror them


Zandario (approved):

Lohikar:
This is deliberate behavior, are you not using invisibility for observers?

PR #4979, opened 2023-01-16
Z-Mimic: Automatic Plane Mangling

Lohikar:
This is somewhat experimental, but it should mostly eliminate the need to manually set the plane mangling flag on movables.

Also renames the movable level `mz_flags` to `zmm_flags` to fix VV.

~~WIP: I really don't like how many `istype(src, ...)` this code contains.~~


Lohikar:
I don't think it's likely I'm going to fix the ultimately-minor style complaint if the maintainers are OK with it, so unmarking as draft.

TheLordME (approved):

Zandario (approved):

PR #4985, opened 2023-01-17
Cyborg init refactor + other tweaks/fixes

Lohikar:
This is more of a reshuffling of borg init than a refactor of it, but at least modifying modules should be less painful now.

Other changes:
- Fixed a runtime that could occur when a cyborg used certain non-TGUI consoles.
- Dogborg rest no longer has visual lag.
- Borg movement delay now actually works.
- Some borg actions now have `target` set for do_after progbars.

:cl:
tweak: Targeted borg emotes now only require a partial case-insensitive match, i.e. `*bark-gla` to bark at a `Glass Airlock`. This only applies to borgs using emotes.
add: Dogborgs can now bark at objects.
fix: Module resets should no longer render dogborgs invisible.
fix: Module resets should no longer temporarily break borg inventories.
fix: Some borg emotes that were supposed to have sounds now actually play said sounds.
/:cl:

### DIFF
--- a/citadel.dme
+++ b/citadel.dme
@@ -3095,6 +3095,7 @@
 #include "code\modules\mob\living\silicon\robot\drone\swarm_abilities.dm"
 #include "code\modules\mob\living\silicon\robot\drone\swarm_items.dm"
 #include "code\modules\mob\living\silicon\robot\drone\~defines.dm"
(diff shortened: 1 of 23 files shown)


MediHound:
On request of Lohikar, I've tested if I can still select my custom sprite with this PR testmerged.
Result of this is that I can not select my custom sprite with this PR testmerged.

PR #4986, opened 2023-01-17
Installs Prettier for TGUI

Zandario:
## About The Pull Request
It's pretty

based on https://github.com/tgstation/tgstation/pull/67935


Zandario:
So prettierx is VERY different from prettier. That's annoying.

PR #4996, opened 2023-01-24
Whitelist backend changes

TheLordME:
## Why It's Good For The Game
Whitelist are run by name, so renaming a whitelisted species also needs an update to the WL as well.
Removes a shitty gimmick of a mirror most people dont have access to because its a fucking nightmare of conditions we partially removed
and a fix for both runtimes and uninteded stat cheating

## Changelog

:cl:
del: Removed the code behind the raider mirror
fix: fixed the shadekin traits applying to non shadekin
config: changed WL using Species ID rather than name
refactor: changed character species to always have a superspecies_id
/:cl:

### DIFF
--- a/code/__DEFINES/species/species.dm
+++ b/code/__DEFINES/species/species.dm
@@ -1,6 +1,5 @@
 /**
  * Species IDs
- * any time you check species.id or species.uid you should put it through the ckey() proc, at least for Whitelists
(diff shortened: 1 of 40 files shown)


TheLordME:
Test on local reveals that all Whitelists still work as intended.

PR #5008, opened 2023-01-27
Adds a dice roll emote

gurking:
## About The Pull Request

Adds the *roll emote. Using the params, you can roll a die of your choice, add a bonus to your roll and compare it against a difficulty check.

To use this emote specify what die you want to roll in the first line. Prefix it with a "D". You may add a bonus to your roll, simply add a "+" and your number after your chosen die. If you want to incorporate a difficulty check, add a "-DC" at the end. You may use a bonus, a difficulty check or both.

Here is an example: roll-D20+5-DC15. This will roll a twenty-sided die, add five and compare it against a difficulty check of 15.

image
(Typo already fixed, not getting a new image.

## Why It's Good For The Game

It's great for text RP to be able to fairly decide contests.

## Changelog

:cl:
add: *roll
/:cl:


TheLordME (changes requested):
**Bag of gaming dice**: _Am I a joke to you? _

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -974,10 +974,53 @@
 			else
 				message = "makes a light spitting noise, a poor attempt at a whistle."
 
+		if ("roll")
+			if(!param)
+				to_chat(src, "To use this emote specify what die you want to roll in the first line. Prefix it with a \"D\". You may add a bonus to your roll, simply add a \"+\" and your number after your chosen die. If you want to incorporate a difficulty check, add a \"-DC\". You may use a bonus, a difficulty check or both.<br>Here is an example: roll-D20+5-DC20. This will roll a twenty-sided die, add five and compare it against a difficulty check of 15.")
```
Shadow-Quill, line 979: `roll-D20+5-DC20`
Wouldn't this be a Difficulty check of 20, not 15?

gurking:
> **Bag of gaming dice**: _Am I a joke to you?_

This isn't the character physically throwing a die, this is a dice roll to resolve role-play scenarios à la TTRPGs.

code/modules/mob/living/carbon/human/emote.dm
gurking: Typo. Fixed.

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -974,10 +974,53 @@
...
+				if (!t2)
+					dc = null
+				var/die_number = text2num(copytext(die, 2, length(die) + 1))
+				var/bonus_number = text2num(bonus)
+				var/dc_number = text2num(copytext(dc, 3, length(dc) + 1))
+
+				var/die_result = rand(die_number)
+				var/die_total = die_result + bonus_number
+
+				if (die_number)
```
TheLordME, line 998: this should be structured in an else if chain, starting with the "die_number && bonus_number && dc_number" check etc.

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -974,10 +974,53 @@
 			else
 				message = "makes a light spitting noise, a poor attempt at a whistle."
 
+		if ("roll")
+			if(!param)
+				to_chat(src, "To use this emote specify what die you want to roll in the first line. Prefix it with a \"D\". You may add a bonus to your roll, simply add a \"+\" and your number after your chosen die. If you want to incorporate a difficulty check, add a \"-DC\". You may use a bonus, a difficulty check or both.<br>Here is an example: roll-D20+5-DC15. This will roll a twenty-sided die, add five and compare it against a difficulty check of 15.")
```
TheLordME, line 979: this usage text should be a variable you declare before the if(!param) and used here and at the if(!message), to avoid having to adjust it multiple times (its also better style)

code/modules/mob/living/carbon/human/emote.dm
```diff
@@ -974,10 +974,53 @@
...
+				return
+			else
```
TheLordME, line 981: you dont need this else, since you return in the other case anyway

TheLordME (changes requested):

TheLordME:
> As requested, a use case:
> 
> Two people are having a role-play conflict in writing. They are not using mechanics to resolve the conflict. This could be a fight scene for example. Using the dice roll emote, they can fairly roll for outcomes that both players can see. They use the appropriate dice roll, any eventual bonus due to roleplay advantages like species or scenarios, and use an appropriately set difficulty check to determine the outcome of their attempts.
> 
> This is only used between consenting players, of course. But it allows for a transparent and fair system to decide contests in a text-based situation.

I feel unhappy about this being an emote, it should rather be an OOC verb, but I couldnt tell you how id realise that, so this is an okay compromise

gurking:
Changed code as requested, thanks for the style tips.

silicons (approved):

silicons (approved):

TheLordME (approved):

PR #5031, opened 2023-02-03
kills observers (no, not ghosts)

Zandario:
## About The Pull Request
Signals all the way bb

Drafted as I want someone to make sure my tired self didn't screw up a signal.


nevimer:
Opportunity to wrap these procs in the new macros for 515 compat

Zandario:
True

PR #5060, opened 2023-02-13
adds OSS locked "Biotechnician" title for Medical Doctor

silicons:
requested by papalus


code/modules/jobs/job_types/station/medical/medical_doctor.dm
```diff
@@ -24,7 +24,9 @@
 		"Medical Resident" = /datum/prototype/alt_title/doctor/resident,
 		"Medical Intern" = /datum/prototype/alt_title/doctor/intern,
 		"Orderly" = /datum/prototype/alt_title/orderly
+		"Biotechnician" = /datum/prototype/alt_title/biotechnician
 		)
+	strict_titles = TRUE
```
silicons, line 29: this tells the job datum to switch to "strict" checks which is how it restricts OSS contractors from **only** being able to pick that title

PR #5062, opened 2023-02-13
makes background restricted alt titles far less awful

silicons:
read comments for what they do
they're also marginally faster now


TheLordME (approved):
GAMING!

PR #5099, opened 2023-02-22
makes the radiocarbon spectrometer 10x faster

silicons:
## Why It's Good For The Game

xenorach needs a rework
this is just a bandaid

i have never had to suffer a more miserable "minigame" in my life compared to this

i don't know when/who designed this but i'm not sure a quick time action game is particularly fun or engaging when the tool itself is no more than a fluffy roleplay aid with how much actual good this entire system does for the game outside of giving science something to do on boring rounds.

## Changelog

:cl:
tweak: radiocarbon spectrometers now take 8 seconds instead of 80.
/:cl:


code/modules/xenoarcheaology/tools/geosample_scanner.dm
```diff
@@ -20,7 +20,7 @@
 	var/last_process_worldtime = 0
 	//
 	var/scanner_progress = 0
-	var/scanner_rate = 1.25			//80 seconds per scan
+	var/scanner_rate = 12.5			//80 seconds per scan
```
TheLordME, line 23: maybe adjust the comment as well?

TheLordME, line 23:
```suggestion
	var/scanner_rate = 12.5			//8 seconds per scan
```

TheLordME (changes requested):

TheLordME (approved):

PR #5105, opened 2023-02-25
maybe fixes xenobio stunbatons

silicons:
stunbaton refactor when


code/modules/xenobio/items/weapons.dm
```diff
@@ -14,6 +14,8 @@
 /obj/item/melee/baton/slime/attack_mob(mob/target, mob/user, clickchain_flags, list/params, mult, target_zone, intent)
 	var/mob/living/L = target
 	if(istype(L) && status) // Is it on?
+		// do the usuals - might work might not who knows
+		..()
```
TheLordME, line 18: doesnt this just mean it works like a normal stun baton, before acting like the slimebaton???

PR #5109, opened 2023-02-25
fixes slimebatons

silicons:
title


silicons:
as said in dsicord,

the reason we ..() is because that was always the case
ironically i reomved it last update

the stun goes through for normal people ubt the agonyforce is low enough it's ineffective-ish

TheLordME (approved):

PR #5128, opened 2023-03-04
sets the odds of gravity failure and solar storm events to 0

nik707:
## About The Pull Request

These two events not only don't function, but also make no sense on a planet bound station. So they're turned off now.

## Why It's Good For The Game

events which do nothing mean we dont get an event which might do something

## Changelog

:cl:
tweak: tweaked chances of rolling gravloss/solar storm to 0
/:cl:


TheLordME (approved):

MediHound:
Solar storms can actually affect miners in the asteroid belt. But true, it's an event which would affect very few people.

PR #5137, opened 2023-03-07
more cursed movespeed rebalancing (teshari buff), slightly, slightly buffs custom species trait selection temporarily

silicons:
Values are experimental

## Why

Since I anti fun'd two races and am about to anti fun rigs and proteans I might as well fully rebalance the most common sources of movespeed mods to make things more equal.

Teshari's old modifier accounted for their old speed. This more accurately reflects where they're at right *now*.

Major Hardy is next on the chopping block after Haste.

Sorry, this is anti fun, but I'd rather encourage people to work on unique species than continue to have Custom Species be the absolute meta with its value tweaks.

In exchange, I'm also lifting the maximum amount of traits you can take for *now*.

All values are subject to change as we roll out rig / species updates in the next few months.

## Changelog

:cl:
tweak: Teshari are more resistant to slowdown now to match their nerfed speed.
tweak: Major Hardy has been nerfed to 0.25x instead of 0.1x.
/:cl:

### DIFF
--- a/code/__DEFINES/misc.dm
+++ b/code/__DEFINES/misc.dm
@@ -441,7 +441,7 @@ GLOBAL_LIST_EMPTY(##LIST_NAME);\
 #define HERM "herm"
 // For custom species
 #define STARTING_SPECIES_POINTS	1
-#define MAX_SPECIES_TRAITS		5
+#define MAX_SPECIES_TRAITS		7
 
 // Xenochimera thing mostly
 #define REVIVING_NOW		-1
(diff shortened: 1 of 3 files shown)


TheLordME:
So, why are we making tesh the new explo meta?

silicons:
> So, why are we making tesh the new explo meta?

they were hit too hard when i murdered them back in the day for being obnoxious hyperspeed chickens
they're now going to basically be lower health but with major hardy with suit, and 33% faster without a suit
i think they're weak enough with their health / limb fragility 

~~honestly nothing really excuses speed but this is less because I like teshari being super fast and more because I'm literally out of options because I hammered them down without actually fixing other species :)~~

the real fix here is /datum/movespeed_modifier and re-evaluating like. everything.
but that'll have to wait for a few more refactors unfortunately, and for us to start actually rebalancing / reworking species

TheLordME (approved):

PR #5158, opened 2023-03-11
Armor rework

silicons:
tl;dr

armor now has % mit and tier
tier modulates % mit and can do stuff like block sharp effects

this makes protection 2d instead of 1d

rebalances and reworks most aromr.


code/__HELPERS/datastructs/armor.dm
```diff
@@ -1,3 +1,234 @@
...
+			return energy
+		if(ARMOR_BOMB)
+			return bomb
+		if(ARMOR_BIO)
+			return bio
+		if(ARMOR_RAD)
+			return rad
+
+/datum/armor/proc/mitigation(flag, tier = ARMOR_TIER_DEFAULT)
+	switch(flag)
```
TheLordME, line 106: shouldnt there be an else here to cover all other cases?

code/__HELPERS/datastructs/armor.dm
```diff
@@ -1,3 +1,234 @@
...
+			return melee_soak
+		if(ARMOR_BULLET)
+			return bullet_soak
+		if(ARMOR_LASER)
+			return laser_soak
+		else
+			return 0
+
+/datum/armor/proc/resultant_damage(damage, tier, flag)
+	switch(flag)
```
TheLordME, line 137: default case?

code/__HELPERS/datastructs/armor.dm
```diff
@@ -1,3 +1,234 @@
...
+		if(ARMOR_ENERGY)
+			return damage * (1 - energy)
+		if(ARMOR_BOMB)
+			return damage * (1 - bomb)
+		if(ARMOR_BIO)
+			return damage * (1 - bio)
+		if(ARMOR_RAD)
+			return damage * (1 - rad)
+
+/datum/armor/proc/describe_list()
```
TheLordME, line 156: I assume this is pseudo OOC information

code/datums/armor/station/exploration.dm
```diff
@@ -0,0 +1,33 @@
+/datum/armor/exploration/soft
+	melee = 0.35
+	melee = MELEE_TIER_DEFAULT
```
TheLordME, line 3: isnt this an overwrite to 0?

code/game/objects/items.dm
```diff
@@ -755,3 +764,24 @@ modules/mob/living/carbon/human/life.dm if you die, you will be zoomed out.
...
+	damage = fetch_armor().resultant_damage(damage, tier, flag)
+	return args.Copy()
+
+/**
+ * called to be used as mob armor
+ * side effects are allowed
+ * 
+ * @returns copy of args with modified values
+ */
+/obj/item/proc/running_mob_armor(damage, tier, flag, mode, attack_type, datum/weapon, target_zone)
```
TheLordME, line 785: isnt this the exact same result as above when called?

TheLordME (changes requested):
I went through the first 259 Files
This should have been 3 or more PRs
1. Armor Rework
2. refactor projectiles away from being an Item subtype
3. Rename "force" to "damage_force"
So, Atomise your PR

code/datums/armor/station/exploration.dm
silicons: yep
tiers are arbitrary. right now it's 0 but we can basically set it to any value, that's why everything is a #define

TheLordME: Then why do you set melee to 0.35 **in the line above**

code/game/objects/items.dm
silicons: check = no side effects, running = side effects

in rigsuit update absorbed damage goes directly to the armor, but if we're just checking armor we don't want it to take damage

code/__HELPERS/datastructs/armor.dm
silicons: nope, if anything it should probably CRASH if someone is using the improper enums, as the only stuff that should go into the proc are the enums.

code/__HELPERS/datastructs/armor.dm
silicons: nope, if anything it should probably CRASH if someone is using the improper enums, as the only stuff that should go into the proc are the enums.

code/__HELPERS/datastructs/armor.dm
silicons: i don't really like the way i'm handling this but it's the only sane way i could think of
i liked the old system where it has 10 strings for 10 cases, but

- handling tiers would be weird
- handling soak would be weird because we might have armor with shit mitigation and very good soak later

code/datums/armor/station/exploration.dm
silicons: oh FUCK that's a typo

TheLordME (changes requested):

TheLordME (approved):

PR #5186, opened 2023-03-16
Ability System

silicons:
We have 3 ability systems in code.

Let's add the final one.

Now including eventual TGUI panel support.

Also refactors mind transfers a bit because why not.


code/modules/species/species.dm
```diff
@@ -176,13 +169,17 @@
 	/// The likelihood of a speech sound playing.
 	var/list/speech_chance = list()
 
-	//// Age
+	//? Abilities
```
TheLordME, line 172: doesnt this replace inherent spells, and maybe inherent verbs, if yes please move it to that block (~line 320)

TheLordME (changes requested):

TheLordME (approved):

PR #5204, opened 2023-03-18
Creates and adds the Mining Carbine

MediHound:
## About The Pull Request

Introduces the Mining Carbine to items that the Mining Vendor can vend, and also creates the Mining Carbine.

## Why It's Good For The Game

Unless a miner uses explosives to mine, their job lasts pretty long and is quickly repetative.
The Mining Carbine is supposed to be an alternate to the Phoron Bore, which can also be bought at the Mining Vendor, allowing miners to have a little bit more variety.
Different firing modes allow the miner to mine specifically a single ore, or mine a cluster of ores or rocks.
While I was at it, I also massively reduced the amount of damage the Scatter Mining Laser does, since it had no reason to be that high, and also changed the firing sound to something that'd hurt the ears less at continuous use.

Suggestion for the Mining Carbine was made in Discord and was greenlit.

## Changelog

:cl:
add: Created the Mining Carbine
add: Added the Mining Carbine to the Mining Vendor
tweak: Changed excavation laser sound
tweak: Added Mining Carbine to list of items that can be carried on the suit
balance: Drastically reduced the excavation scatter damage
/:cl:

### DIFF
--- a/code/modules/clothing/suits/hooded.dm
+++ b/code/modules/clothing/suits/hooded.dm
@@ -476,6 +476,7 @@
 		/obj/item/gun/energy/kinetic_accelerator,
 		/obj/item/kinetic_crusher,
 		/obj/item/resonator,
+		/obj/item/gun/energy/gun/miningcarbine,
 		/obj/item/gun/magnetic/matfed
 		)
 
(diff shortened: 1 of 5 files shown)


silicons (approved):
acceptable for now but will be changed immediately when mining is reworked.

PR #5208, opened 2023-03-19
Greyscaled Human Torso Touchup

MonaraMir:
The gore was sprited onto the torso, which caused issues with certain limbs (vassilian) which had thinner portions at the root. This resulted in pixels that appeared almost at random. I dont know why gore isnt an overlay or something when a limb is removed, but this works *for now*. Both male and female versions were touched up to remove dangling gore pixels.
before
after


Keekenox:
add gore overlays

PR #5209, opened 2023-03-19
Rad collector to port connection

Dimasw99:
## About The Pull Request

Radiation collectors are now able to be connected to connection ports and share its air contents with the connected network.
I have play tested it fairly extensively and couldn't find any issues.

## Why It's Good For The Game

In a normal scenario radiation collectors don't need to be refueled and if, rarely. But in some cases where you push the boundaries of the engine it can become a chore very quickly of pretty significant time consumption. Of course this boon is not without risk and connecting the collector to a network with too much pressure can cause the tank to rupture. Always read the warning labels!

In the end it would add an extra layer of depth for engine designs and also allow for more adventurous options for rad collectors in the future like temperature related radiation resistance.

## Changelog

:cl:
add: Radiation collectors can now be connected to port connections
/:cl:

### DIFF
--- a/code/modules/power/singularity/collector.dm
+++ b/code/modules/power/singularity/collector.dm
@@ -8,6 +8,8 @@
 	density = TRUE
 	req_access = list(ACCESS_ENGINEERING_ENGINE)
 //	use_power = 0
+	var/datum/gas_mixture/air_contents = new
+	var/obj/machinery/atmospherics/portables_connector/connected_port
 	var/obj/item/tank/phoron/P = null
 	/// stored power in kilojoules
 	var/stored_power = 0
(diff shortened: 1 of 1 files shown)


TheLordME (approved):
I guess, cant wait for people to come screeching because they blew up

silicons (changes requested):
while this works as intended, collectors need to use an internal air mixture.

1. it's extraordinarily brittle to snowflake around pipenetworks like this. we already have a somewhat fragile system that needs to be updated, don't add onto it like this.
2. this allows you to basically make a chain of maxcaps by connecting collectors to an extremely overpressurized network. this shouldn't be possible, it already is but this system completely throws the checks and balances we have for this out the window.

PR #5212, opened 2023-03-19
Blueshield HUD icon fix

MonaraMir:
Accidently capitalized the B in the icon file oopsies


MonaraMir:
This doesnt fix sorting order btw I couldnt figure it out

PR #5232, opened 2023-03-24
Changes Gygax sprite to the 'advance' one

SpaceMan-Jim:
Seems kinda silly this was just sitting in the code and not being used anywhere, especially since it looks much better than the one currently used.

## About The Pull Request

Gives the Gygax a face lift by replacing the icon_state, initial_icon, and wreckage lines with gygax_adv and mecha_wreckage/gygax/advance respectively

## Why It's Good For The Game

Gives the Gygax a new fancy look instead of the currently used one. Just found it sitting in the code unused anywhere else.

## Changelog

:cl:

imageadd: gygax_adv replaced instances of the original Gygax sprite, same with the darkgygax.
imagedel: the original ones are simply swapped with the advance.

/:cl:


Captain277:
I don't think we should reskin the Gygax with the Advanced Gygax skin. I'm pretty hesitant on this.

SpaceMan-Jim:
As I stated the 'advance' sprite is practically not used at all in code, there is no advance gygax. The only thing that comes close is the dark gygax simple mob as there is specifically one that is a advance model, which I assume is a neat lil mini boss thing. But again, the sprite is just the one you'd see on TG or any other server and just looks better, as the one we use is.. terrible. Dimasw is right aswell, I can bet on no one is going to go 'i'ma code a advance gygax for players to make/use', just to use the sprite, I'd think that'd just be unnecessary.

Zandario (approved):
This is fine to me for the meantime

Captain277:
> > Fair enough. Do me a favor though, if you're able, and add a Paint Kit for the Gygax that gives it the "Classic" skin. I think we did that for the Durand.
> 
> That is exactly what I thought went on, and do not worry, I had that in mind.. I just, have to figure that out and actually have time on my hands.

Oh it's actually pretty easy! It's code/game/objects/items/paintkit.dm. 
image
Doesn't take much to code, you'd just set the "new_icon" var to the *old* Gygax state.

silicons (approved):

PR #5233, opened 2023-03-24
Adds the Covert/Overt "Jacket" Type Plate Carrier

Captain277:
## About The Pull Request

1. **Adds the Covert/Overt 'advanced plate carrier'.**
2. **Allows the mounting of cosmetics on plate carriers again.**
3. **Adds a Blueshield variant of the above.**

## Why It's Good For The Game

1. _It's ya girl Ariel here with another hacky code gimmick item. This time it's an armor vest with a retractable helmet that can be hidden under certain accessory jackets and cloaks. I decided that the normal hooded jacket was a fine template for this, and I just gave it the plate carrier's unique traits so I could subtype it later if I wanted to make non-Space CIA variants. For now it's just like, a cool thing for events or whatever. Anyways Kevin's armor rework is gonna go through soon and I need to update this after he does that. Plus I might experiment with subtyping this as a voidsuit instead (but I think the jacket version presented here is the right way to go so probably not)._
2. _I feel like cloaks used to be allowed on carriers. Now they should be again. Accessory jackets too._
3. _These carriers are meant to be for like, bodyguards or secret agents. So I figure Blueshields should get one. It's got very basic armor, so it's not going to do much more than serve as a fluff item, I expect._

## Changelog

:cl:
add: Adds the Covert/Overt advanced plate carrier.
add: Adds Blueshield variant of above.
tweak: Allows mounting of accessories on plate carriers.
/:cl:


code/modules/clothing/suits/hooded.dm
TheLordME, line 1: dont these also need armor datums like the hoods?

Captain277: No, because the armor datum is handled by the plate inside the armor. They're basically hoodies spoofing modular armor code, which felt like the most expedient way to achieve this effect.

code/modules/clothing/suits/armor.dm
```diff
@@ -568,14 +568,16 @@
 	item_icons = list(SLOT_ID_SUIT = 'icons/mob/clothing/modular_armor.dmi')
 	icon_state = "pcarrier"
 	valid_accessory_slots = (\
```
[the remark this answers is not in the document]
Captain277: Oh I see! I'll update that.

Zandario (approved):
Whatever, sue me

PR #5238, opened 2023-03-25
makes painkillers not stack on themselves

TheLordME:
## Why It's Good For The Game
its funny how you can take a bit of tramadol and undergo surgery without feeling all that much, this PR remedies is so, that only the strongest active painkiller actually has an effect


yourdoom9898:
Well, at least I don't have to mix ParaTram anymore I guess

PR #5241, opened 2023-03-26
protean changes + synth repair / temperature buffs

silicons:
## About The Pull Request

- proteans now have regular synth vulnerabilities to heat / cold, including being too cold due to vacuum.
- proteans however, won't die from it as their blob forms are still immune
- proteans have been given major hardy in exchange for everything else
- removes **all** material consumption buffs from proteans, other than, obviously, steel.
- adds nanite chambers that can revive a protean using materials, depending on if their brain / orchestrator / refactory are inside.
- synths now take way less material to fix + incinerate slower when not using heat sinks
- fixes some serious movement issues

## Why It's Good For The Game

the proper synth + protean rework comes later
but since we wanted to play the game of "everything should be op because proteans are"
i decided i'll start playing the game too.

## Changelog

:cl:
del: protean immunity to heat - blobform immunity stays, so it is not a full kill
add: nanite chambers - can regenerate and revive proteans
balance: synths are stronger and easier to repair now
/:cl:


Kraseo:
- aronai snowflake race code :+1:

Dimasw99:
> including being too cold due to vacuum.
TG moment. Should be over heating.

TheLordME (approved):

PR #5257, opened 2023-03-30
imrove buildmode turf heuristics

silicons:


code/modules/admin/verbs/buildmode.dm
```diff
@@ -371,8 +371,15 @@ GLOBAL_LIST_EMPTY(buildholders)
 			if(pa.Find("left") && !pa.Find("ctrl"))
 				var/turf/TC = get_turf(object)
 				if(ispath(holder.buildmode.objholder,/turf))
+					// warning: this is bad heuristics, but for the time being, we don't have another choice / i'm too lazy to think this through.
 					var/turf/T = get_turf(object)
-					T.ChangeTurf(holder.buildmode.objholder)
+					var/turf/making_path = holder.buildmode.objholder
+					if(T.density) // T is dense, we likely want to replace it as it's likely a wall
+						T.ChangeTurf(holder.buildmode.objholder)
+					else if(!T.density && initial(making_path.density)) // thing we're placing is dense but not existing turf, place on top
```
Zandario, line 379: is `!T.density` really needed here? This would only be checked if it's already false?

silicons: nope, hang on

Zandario (approved):

PR #5263, opened 2023-04-01
Adds Craftable Plasma Weapons

Captain277:
## About The Pull Request

1. **Adds a craftable plasma rifle and pistol.**

## Why It's Good For The Game

1. _Some might say it isn't, but I love fancy weapons, and diversity, and explosions. This gun contributes to all three. It functions along the standard rules for this kind of gun. It will eventually be refined into a different overheat mechanic, but for now it's a diceroll system when fired on high power._

## Changelog

:cl:
add: Adds a plasma rifle and pistol.
/:cl:


silicons (dismissed):
1. no r&d design for this should happen. i'm already at wit's end dealing with science balance and i'm not going to greenlight novel ideas right now.
2. this is literally a craftable gyrojet.

Captain277:
> 1. no r&d design for this should happen. i'm already at wit's end dealing with science balance and i'm not going to greenlight novel ideas right now.
> 2. this is literally a craftable gyrojet.

Yeah the initial code putting it in R&D stayed in after I made it craftable. I caught that issue earlier today and resolved it. As for point two, I've made the weapon prohibitively expensive to produce. I'm happy to make it even more expensive/complex to build if necessary.

code/game/objects/items/weapons/other.dm
```diff
@@ -901,3 +901,25 @@
...
+	icon = 'icons/obj/items_vr.dmi'
+	icon_state = "caseplasmarifle"
+	w_class = ITEMSIZE_NORMAL
+
+/obj/item/plasma_pistol_casing
+	name = "casing (plasma pistol)"
+	desc = "This hammered plasteel casing has been shaped and cut into the upper frame of a plasma pistol."
+	icon = 'icons/obj/items_vr.dmi'
+	icon_state = "caseplasmapistol"
+	w_class = ITEMSIZE_NORMAL
```
lolman360, line 925: why do these exist if they're just used for like... 1 recipe each, 2 for the coil? just add the copper/plasteel costs to the recipes and increase the time to craft on both

TheLordME (approved):

PR #5264, opened 2023-04-01
Multiple Map Fixes

MediHound:
## About The Pull Request

This fixes three mapping issues that were found ingame or reported on Discord.

## Why It's Good For The Game

1) RnD needs Fabrication access on the ID to be entered, not general Science access. (This means that jobs with general Science access, like Explorer/Pilot/Blueshield can no longer walk into RnD and print things.)
2) Fixes a wall in the north west of Surface 3 blocking a stair, thus allowing people to walk up but not down.
3) Fixes the Chemistry area, making it so that teleporting to Chemistry teleports one into Chemistry, not on a wall outside of it.

## Changelog

:cl:
tweak: Access on RnD doors
tweak: Wall placement above stairs on NW Surface 3
tweak: Re-adjusted Chemistry area
/:cl:


MonaraMir:
Allow blueshields access to RND, Robotics, and Circuitry. They need to reach these spots to perform their duties for the RD. The "general" area for research is basically a T junction which makes up a small small fraction of time spent walking through by the RD.

MediHound:
> Allow blueshields access to RND, Robotics, and Circuitry. They need to reach these spots to perform their duties for the RD. The "general" area for research is basically a T junction which makes up a small small fraction of time spent walking through by the RD.

Negative.
I have asked about it, as seen here. Blueshields do not get RnD access.
https://discordapp.com/channels/161574200581029888/1071527124718076024/1091549302540738650
Neither do CE, HOP, CMO, and so on. Only the Research department has RnD access.
If an RD is on duty and the Blueshield is with them, the RD can open the door for them.
No one but Research gets access there by default on their card.

I fully understand that some Blueshield players like to use the way it currently is to walk into RnD and print things (I've seen it happen twice in the last week), but they get Common Department access.
This has been further confirmed by a Head Admin.

Furthermore, if the Blueshield role needs access to specific areas, then they should get the access on their ID card.
Right now they unintentionally have access to RnD not because they mistakingly were given that permission on their ID card, but because the door leading into RnD is not asking for RnD access.
By finally setting the doors up correctly, aka making the RnD doors ask for RnD access, we also make it so other roles which intentionally get only Common Science access and not RnD (Explorer) can not walk in there by themselves.

Last but not least, it is not up to me to decide what areas which role gets access to. If you want to add RnD access to a role's ID card, then bring it up with the Maintainer(s) and if they agree, give them the access.
What I am doing with my PR is not locking roles out of RnD, but fixing the door so it asks for RnD access.

silicons (approved):

PR #5279, opened 2023-04-03
Buckets Now Reflect When They're (Mostly) Full of Liquid

Captain277:
## About The Pull Request

1. **Enables Overlay that Shows When a Bucket Has Liquid In It.**

## Why It's Good For The Game

1. _Just a nice little QOL thing. Or really just an aesthetic thing, really._

## Changelog

:cl:
tweak: Enables check for water overlay on buckets.
/:cl:

### DIFF
--- a/code/modules/reagents/reagent_containers/glass.dm
+++ b/code/modules/reagents/reagent_containers/glass.dm
@@ -299,11 +299,18 @@
 	else
 		return ..()
 
+/obj/item/reagent_containers/glass/bucket/afterattack()
+	.=..()
+	update_icon()
+
 /obj/item/reagent_containers/glass/bucket/update_icon()
 	cut_overlays()
 	if (!is_open_container())
 		var/image/lid = image(icon, src, "lid_[initial(icon_state)]")
 		add_overlay(lid)
+	if (is_open_container() && reagents.total_volume > 100)
+		var/image/water = image(icon, src, "water_[initial(icon_state)]")
+		add_overlay(water)
 
 /obj/item/reagent_containers/glass/bucket/wood
 	desc = "An old wooden bucket."


code/modules/reagents/reagent_containers/glass.dm
```diff
@@ -299,11 +299,18 @@
...
+	update_icon()
+
 /obj/item/reagent_containers/glass/bucket/update_icon()
 	cut_overlays()
 	if (!is_open_container())
 		var/image/lid = image(icon, src, "lid_[initial(icon_state)]")
 		add_overlay(lid)
+	if (is_open_container() && reagents.total_volume > 100)
+		var/image/water = image(icon, src, "water_[initial(icon_state)]")
+		add_overlay(water)
```
TheLordME, line 313: does this overlay work with the wooden version below?

TheLordME: if not, please either make it work, or make sure its not applied

TheLordME (changes requested):

code/modules/reagents/reagent_containers/glass.dm
Captain277: Yeah it should work globally.

Captain277: Being that the wood and standstone buckets are still both subtypes of glass/bucket.

Captain277: Whoops I stand corrected. Apparently it's not asserting. Easy fix.

TheLordME: my main concern is that the overlay does not fit visually into the bucket

TheLordME (approved):

PR #5280, opened 2023-04-03
hotfix for simplemob armor + stairs

silicons:
we need to rewrite movement backend later


code/modules/multiz/structures.dm
TheLordME, line 1: dont comment out code, delete it

TheLordME (changes requested):

TheLordME (approved):

PR #5294, opened 2023-04-05
fixes armor examine formatting

silicons:
br not \n


lolman360:
can we use a different delimiting character instead fo @?? | maybe? #? anything but @

TheLordME (approved):

PR #5297, opened 2023-04-05
"Temporary" fix for fighters being unable to move

MonaraMir:
""""Temp"""" fix for fighters not being able to move due to a movement update that fucks 64x64 binding boxes.

No I dont plan to fix 64x64, but I know it'll be months on end before these are functional otherwise


silicons:
w

PR #5299, opened 2023-04-05
adds a delay to translocators

silicons:
telescience update will deal with the rest
unfortunately people won't stop gaming with this so

:cl:
balance: translocators now have a 5 second activation delay when teleporting others
/:cl:

### DIFF
--- a/code/modules/vore/fluffstuff/custom_items.dm
+++ b/code/modules/vore/fluffstuff/custom_items.dm
@@ -986,7 +986,7 @@
 	if(!teleport_checks(target,user))
 		return //The checks proc can send them a message if it wants.
 	
-	if(!do_after(user, 5 SECONDS, target))
+	if(user != target && !do_after(user, 5 SECONDS, target))
 		return
 
 	ready = FALSE
--- a/code/modules/vore/fluffstuff/custom_items.dm
+++ b/code/modules/vore/fluffstuff/custom_items.dm
@@ -985,6 +985,9 @@
 
 	if(!teleport_checks(target,user))
 		return //The checks proc can send them a message if it wants.
+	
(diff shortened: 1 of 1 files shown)


TheLordME (approved):

MediHound:
Requesting a change to this PR:
The primary roles who have consistent access to Translocators, are Explorers and Miners. The intended use is as a retreat option when wounded.
By adding a 5 second timer, this primary use is nullified. A 5 second delay implies that the Miner or Explorer has been able to get far enough away from the threat to be able to stand around for 5 seconds doing nothing. However, in most cases when one gets in a situation where one would need to use a Translocator to use escape danger, one is under immediate danger and can not wait for 5 seconds. Waiting 5 seconds while a mob is attacking you, leads to nothing but the person going into crit and thus dropping the translocator, thus making the attempt to use it a death trap.

The change I request as such is the following:
When the Translocator is used on anyone OTHER than the uses, there is a 5-10 second timer. When the user uses it on himself, there is no timer.

MonaraMir:
This is intentional. It's specifically to stop instant escapes and instant captures.

MediHound:
Then this is effectively removing the actually intended use, rolling back balance about 4 months where one lucky hit (broken foot/hand) could effectively instantly decide the battle with no chance of escape. Not speaking about PVP but normal PVE.
And it is removing the actually intended use of the Mini Translocator without openly announcing or making it visible that it is removing said intended use.
The announcement that was given was "Miners, watch out, Translocators now need 5 seconds to activate" and not "Miners, you can no longer use the Translocator to escape death because we intentionally do not want you to be able to escape."

Captain277 (approved):

PR #5301, opened 2023-04-06
Z-Mimic (and AO): Minor fixes

Lohikar:
- Fixes a runtime with ZM and move-up.
- Some types that don't make sense to mimic have been marked IGNORE.
- Mimics can no longer get wet from the pool.
- AO should no longer look weird around diagonal shuttle walls.


TheLordME (approved):
I guess

PR #5312, opened 2023-04-06
Glass Flooring

Zandario:
## About The Pull Request
I've been lazy about adding this and then Lohikar said Kapu might beat me to it on Daedalus, so of course I did this as fast as I could.

Do note that this doesn't add it anywhere on the map, I'll let mappers find good uses.

## Preview
06-Thursday-dreamseeker_4337


Lohikar:
One note to mappers: it should be legal to put glass turfs on z-levels that don't have a z below, they'll copy space (parallax) instead.

Captain277 (approved):

PR #5321, opened 2023-04-08
Adds Lawn Gnomes (You Just Got Gnomed)

Captain277:
## About The Pull Request

1. **Adds two flavors of craftable lawn gnome.**

## Why It's Good For The Game

1. _I was hired to create gnomes, and I have thus created gnomes. These two lawn gnomes (normal and giant), are craftable with marble. A limited supply may be purchased from toy vendors. One has also been added to the botany premium menu as well. The server just got gnomed._

## Changelog

:cl:
add: Adds lawn gnomes.
/:cl:


code/datums/components/crafting/recipes/recipes_misc.dm
```diff
@@ -495,6 +488,29 @@
 	category = CAT_MISCELLANEOUS
 */
```
TheLordME, line 489: throw out the remaining comments above, please?

TheLordME (changes requested):

code/datums/components/crafting/recipes/recipes_misc.dm
TheLordME: I meant all of them...

Captain277: I like leaving in things that haven't been added yet, so we have a reference to work off of if we ever add them. The extendo hand, for instance. I think there's value in retaining it as a concept/recipe for the future. I often benefit from these kinds of notes.

TheLordME (approved):

PR #5334, opened 2023-04-09
massively buffs rcd decon

silicons:
taking 10 metal to deconstruct a floor that took 1 metal is stupid
let's change it to 2.


lolman360:
mfw can now decon airlocks for basically free :D

Captain277 (approved):

PR #5389, opened 2023-04-17
Fixes Several Bugs With Traps and Increases Lethality

Captain277:
## About The Pull Request

1. **Increases Trap Lethality/Maiming Ability.**
2. **Fixes Bugs With Traps.**

## Why It's Good For The Game

1. _Certain traps, namely pop-up pillar and buzz-saw traps, now dismember people struck by them._
2. _Leg breaker traps now actually break legs, traps that tripped even when covered have the correct checks put in place now, and the falling log can be cut by any sharp object/actually updates its icon when broken._

## Changelog

:cl:
add: Adds dismemberment to blade traps.
fix: Fixes bugs in trap code.
/:cl:

### DIFF
--- a/code/game/objects/effects/traps.dm
+++ b/code/game/objects/effects/traps.dm
@@ -174,7 +174,11 @@ Add those other swinging traps you mentioned above!
 	. = ..()
 	if(AM.is_incorporeal())
 		return
-	if (istype(AM, /mob/living))
+
+	if(!tripped)
+		return
+
+	else if(istype(AM, /mob/living))
 		var/mob/living/M = AM
 		var/damage = rand(min_damage, max_damage)
 		M.apply_damage(damage, BRUTE)
@@ -209,8 +213,12 @@ Add those other swinging traps you mentioned above!
 	. = ..()
 	if(AM.is_incorporeal())
 		return
-	if (istype(AM, /mob/living))
-		break_legs()
+
+	if(!tripped)
+		return
+
+	else if(istype(AM, /mob/living))
+		break_legs(AM)
 		AM.visible_message("<span class='danger'>[AM] falls into the path of the piston!</span>", \
 						"<span class='userdanger'>Your leg is crushed by the piston!</span>")
 
@@ -264,7 +272,11 @@ Add those other swinging traps you mentioned above!
 	. = ..()
 	if(AM.is_incorporeal())
 		return
-	if (istype(AM, /mob/living))
+
+	if(!tripped)
+		return
+
+	else if(istype(AM, /mob/living))
 		var/mob/living/M = AM
 		var/damage = rand(min_damage, max_damage)
 		M.apply_damage(damage, TOX)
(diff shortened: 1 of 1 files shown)


code/game/objects/effects/traps.dm
```diff
@@ -174,7 +174,11 @@ Add those other swinging traps you mentioned above!
 	. = ..()
 	if(AM.is_incorporeal())
 		return
-	if (istype(AM, /mob/living))
+
+	if(!tripped)
```
TheLordME, line 178: unless tripped is labeled wrong, this means the damage is only applied when the trap has been tripped already

code/game/objects/effects/traps.dm
```diff
@@ -209,8 +213,12 @@ Add those other swinging traps you mentioned above!
 	. = ..()
 	if(AM.is_incorporeal())
 		return
-	if (istype(AM, /mob/living))
-		break_legs()
+
+	if(!tripped)
```
TheLordME, line 217: same as above

code/game/objects/effects/traps.dm
```diff
@@ -264,7 +272,11 @@ Add those other swinging traps you mentioned above!
 	. = ..()
 	if(AM.is_incorporeal())
 		return
-	if (istype(AM, /mob/living))
+
+	if(!tripped)
```
TheLordME, line 276: same as above
and you should probably add a proc that is "apply_damage(/mob/living/victim)" that is changed per trap, instead of overwriting Crossed() every time

code/game/objects/effects/traps.dm
```diff
@@ -490,10 +504,13 @@ Add those other swinging traps you mentioned above!
 	. = ..()
 	if(AM.is_incorporeal())
 		return
-	else if(broken)
+
+	if(broken)
 		return
-	else if (!tripped)
+
+	if(!tripped)
```
TheLordME, line 511: same as above

code/game/objects/effects/traps.dm
```diff
@@ -521,14 +538,21 @@ Add those other swinging traps you mentioned above!
...
```
TheLordME, line 545: same as above

code/game/objects/effects/traps.dm
```diff
@@ -601,10 +632,13 @@ if (istype(AM, /mob/living))
...
```
TheLordME, line 639: same as above

code/game/objects/effects/traps.dm
```diff
@@ -633,10 +667,13 @@ if (istype(AM, /mob/living))
 	. = ..()
 	if(AM.is_incorporeal())
 		return
-	else if(broken)
+
+	if(broken)
 		return
-	else if (!tripped)
+
+	if (!tripped)
```
TheLordME, line 674: I keep repeating myself

TheLordME (changes requested):

code/game/objects/effects/traps.dm
Captain277: Tripped means the trap is indeed triggered. Traps are triggered by pressure plates. Damage shouldn't be applied unless the trap is activated by a pressure plate, ergo, if the trap isn't active, it shouldn't do damage.

silicons (approved):

PR #5391, opened 2023-04-17
Rift Arrival spawn fix + classm dorms

MonaraMir:
Placed shuttle dock spawns instead of arrival shuttle spawns. Moved shuttle dock, placed arrival spawns where they used to be.

Class M dorms had linked door bolts on 1/3rd of dorm rooms. Fixed here.


TheLordME (dismissed):

MonaraMir:
Shitty broken system that doesn't actually select spawns anymore for who knows how long. Only placing normal shuttle dock spawns at the shuttle dock to force everyone to spawn there. Someone else needs to come figure out why this system doesn't work, and likely hasn't been working for months on end now. Preferably before shipmap, since we need to have 2 different latejoin locations (cryo,robo)

silicons (approved):

PR #5399, opened 2023-04-20
aggressive memory optimizations

silicons:
- removes geosample: no one ever used this and it needs a way better system than this
- atom colours are only applied to obj and mobs by default now, there's no point tracking for further up stuff, as they probably have their own systems
- turns a bunch of lists into lazylists

### DIFF
--- a/code/game/atoms/atom.dm
+++ b/code/game/atoms/atom.dm
@@ -248,10 +248,6 @@
 	if(loc)
 		SEND_SIGNAL(loc, COMSIG_ATOM_INITIALIZED_ON, src) /// Sends a signal that the new atom `src`, has been created at `loc`
 
-	//atom color stuff
-	if(!isnull(color) && atom_colouration_system)
-		add_atom_colour(color, FIXED_COLOUR_PRIORITY)
-
 	if(light_power && light_range)
 		update_light()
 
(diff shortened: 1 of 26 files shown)


code/game/atoms/atom.dm
```diff
@@ -276,7 +264,7 @@
  *
  * To have your LateIntialize proc be called, your atoms [Initalization][/atom/proc/Initialize]
  *  proc must return the hint
- * [INITIALIZE_HINT_LATELOAD] otherwise you will never be called.
+ * [INITIALIZE_HINT_LATELOAD] otherwise you will never be called.0
```
TheLordME, line 267: why is there a 0 at the end of this line?

TheLordME (approved):

code/game/atoms/atom.dm
```diff
@@ -256,10 +248,6 @@
 	if(loc)
 		SEND_SIGNAL(loc, COMSIG_ATOM_INITIALIZED_ON, src) /// Sends a signal that the new atom `src`, has been created at `loc`
 
-	//atom color stuff
-	if(color)
-		add_atom_colour(color, FIXED_COLOUR_PRIORITY)
-
```
Kapu1178, line 262: this wont use much memory at all.

TheLordME (approved):

silicons:
for nayone reading tihs later:

yes, i know this is "trivial" and doesn't do much, but i'm just doing this because it's 30 minutes of effort to claw a few tens of MB back and on god citadel rp is running into the fucking Ceiling right now (near 3GB alloc!)

PR #5403, opened 2023-04-21
Adjusts the vatborn hunger_rate

gurking:
## About The Pull Request

Lowers the vatborn default hunger rate from 700% to 50% higher than the baseline human.

## Why It's Good For The Game

The comment makes me believe they were originally intended to have a *lower* hunger rate than humans. I feel like it's fine to keep them higher, this is the same value as grav-adpated humans have, without being ridiculous with how much you have to eat.

## Changelog

:cl:
balance: Balances the vatgrown hunger rate
/:cl:

### DIFF
--- a/code/modules/species/station/standard/human_subspecies.dm
+++ b/code/modules/species/station/standard/human_subspecies.dm
@@ -117,7 +117,7 @@
 	toxins_mod = 1.05
 
 	total_health = 115
-	hunger_factor =  0.075//50% more hungry
+	hunger_factor =  0.035
 
 	species_spawn_flags = SPECIES_SPAWN_CHARACTER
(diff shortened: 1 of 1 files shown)


TheLordME (changes requested):
The comment was correct, the original intended value was 0.035,seems like i misplaced a zero
I am not seeing any reason why vatgrowns would need to have an increased hunger

TheLordME:
Oh, and for the changelog:
Please use full sentences, like 
balance: Balances the vatgrown hunger rate
(These are the entries you see when opening the changelog in game)

gurking:
Yeah okay, I changed it to 0.035. The exact value doesn't really matter, I just didn't want to go overboard since they took more food for the longest time. Also, I'm pretty sure the changelog is deprecated.

TheLordME (approved):
changelog is just broken, dont worry about it to much

PR #5411, opened 2023-04-23
Fighter Fabrication (Pinnace)

MonaraMir:
Adds a way to fabricate, assemble, and construct fighters in the same way mechs are built. Pinnace only until I add more NT fighters, and get "landing gear mode" for baron/space only fighters.

LIMITED BUILD STATE ICONS!!
Refer to chat to know what step you're on.


MonaraMir:
Attempt at build state icons does not function. Not sure why, but it doesnt matter much and it was only 2 states (cockpit controls from red, to green, to grey) so I'm just leaving it alone for now. I might fix it when I do the other fighters.

TheLordME (approved):

Captain277:
> Attempt at build state icons does not function. Not sure why, but it doesnt matter much and it was only 2 states (cockpit controls from red, to green, to grey) so I'm just leaving it alone for now. I might fix it when I do the other fighters.

I'd be interested in getting this working before the fullmerge. Because once it's in, these missing steps will be lost to memory.

PR #5416, opened 2023-04-24
Impromptu Living Statue Rework

Zandario:
## About The Pull Request

Just was supposed to be a port of tgstation/tgstation#73630, but I felt like fixing the other things.

- Their spells function and are added.
- Their main gimmick is now a component, so that can be reused for other objects/mobs.

## Changelog

:cl:
refactor: Living Statues properly function.
/:cl:


TheLordME (commented):
technically approved, but out of fear cap might just merge it when they see the approved review

Captain277 (approved):

PR #5446, opened 2023-04-29
Effective dose works for alcohol subtypes' effects, + alcohol intolerance

Putnam3145:
## About The Pull Request

Whoops I accidentally did a balance change with my please-don't-trigger-me PR.

Adds alcohol intolerance, a trait that prevents proper metabolization of most alcohol ingredients; instead of effects simulating drunkenness, you get _immediate_ liver failure, toxin damage, pain and vomiting. Yes, I actually do prefer that personally.

Certain drinks now have their effects modified by the *effective dose* calculated in the main ethanol affect_ingest and affect_blood procs, rather than just sort of guessing at it or doing completely wrong things (deathbell was doing `if(dose * strength > strength)`, which is to say, `if(dose > 1)`, for example).

## Why It's Good For The Game

honestly it's mostly just a personal thing, don't want my character to get drunk. The doses thing should be good, though, since it actually makes effects of various drinks, like, sane.

## Changelog

:cl:
add: Alcohol intolerance (no drunkenness, just horrific amounts of damage)
tweak: Various alcoholic drinks now actually take strength mod of species into account for secondary/additional effects
/:cl:

### DIFF
--- a/citadel.dme
+++ b/citadel.dme
@@ -170,7 +170,6 @@
 #include "code\__DEFINES\dcs\signals\signals_item\signals_item_economy.dm"
 #include "code\__DEFINES\dcs\signals\signals_item\signals_item_inventory.dm"
 #include "code\__DEFINES\dcs\signals\signals_mob\signals_mob_appearance.dm"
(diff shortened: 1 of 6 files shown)


TheLordME (approved):
i guess

PR #5480, opened 2023-05-08
minor mapping issues in sec

MonaraMir:
Fixes minor mapping issues in sec

Added a bolt button to sec bathroom

Fixed EVA storage lighting

Removed the maint door that led to a flat wall

Added emergency shutters to a few sec doors around the break room


TheLordME (approved):
Minor mapping issues you say....

PR #5486, opened 2023-05-10
Pinnace costs less plasteel

MonaraMir:
Makes everything half or more as cheap in plasteel, without changing other materials

Too much plasteel for how weak the ship is.


TheLordME (approved):
I guess

PR #5491, opened 2023-05-12
Qpad Precursor Tech Requirement

MonaraMir:
Makes Qpads require precursor tech to be created.

#5489 fixes the bridge teleporter and gives a one-way option to command instead of a two-way option to literally anyone science hands Qpad machinery to. (exploration)

### DIFF
--- a/code/modules/research/designs/circuits/circuits_vr.dm
+++ b/code/modules/research/designs/circuits/circuits_vr.dm
@@ -22,7 +22,7 @@
 /datum/design/circuit/quantum_pad
 	name = "Quantum Pad"
 	id = "quantum_pad"
-	req_tech = list(TECH_ENGINEERING = 4, TECH_POWER = 4, TECH_BLUESPACE = 4)
+	req_tech = list(TECH_ENGINEERING = 4, TECH_POWER = 4, TECH_BLUESPACE = 4, TECH_PRECURSOR = 1)
 	build_path = /obj/item/circuitboard/quantumpad
 	sort_string = "HABAH"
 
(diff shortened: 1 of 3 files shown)


TheLordME (changes requested):
precursor tech isnt admin spawn only, you can obtain precursor tech pretty easily if there is mining, or you have luck with archeology

MonaraMir:
Yeah intended, it's supposed to be harder and far less frequent mostly. If it still becomes an issue I'll make it admin spawn only.

TheLordME (approved):

pixelkitty286:
The precursor tech level should be higher since engineering can just give science an XYZ thing which is precursor.

yourdoom9898:
I liked going to Gaia, guess that's never happening again

IrkallaEpsilon:
Might as well remove it, Xenoarch locking is effectively removing it from the game. Nobody ever does that (for VERY good reason, its mind numbing and pointless most of the time). Engineering rarely gets that one thing from the vendor, so why soft remove it. Remove it proper next time.

MonaraMir:
The goal is to effectively remove Qpads from most rounds. Special areas like gaia can have a teleporter of their own mapped in, a gateway (we'd need a change so you can select the destinations on gateways, so gaia doesnt completely block it), or a Qpad that leads between gaia and the station. (stationside qpad would 100% get disassembled, analyzed, and reused elsewhere now that the parts contain precursor tech.)

If you wanna go to gaia, the option's available via exploration, but I can map a teleporter if it's highly requested.

The issue with two-way teleportation manifested in roughly half of the events I ran. It removes tension by allowing for instant travel between safe, and not safe zones. I'd rather have one-way teleport to supply explorers, security, and other departments. Hopefully encouraging triage stations and resupply requests via Holocall instead of just teleporting back and doing it yourself. planetside (and soon shipside) engagement with external crew should increase from this change.

MonaraMir:
Note that exploration can still get precursor tech sometimes, and there's other methods for getting the tech. It just needs to appear way less, and have a precursor explanation behind it, since it doesn't even really fit with the direction of our lore (as lore maints told me, I dont know the details)

PR #5495, opened 2023-05-14
abilitiese Sonar pulse

TheLordME:
:cl:
tweak: sonar pulse is now an ability, rather than a verb
/:cl:


lolman360:
can you still macro it

TheLordME:
I think so, but you need to ask @silicons , they made the ability system, I think there is a way to bind abilities to keys
but I dont know how

silicons (approved):

PR #5503, opened 2023-05-16
Preference Defaults and Keybinding support/defaults

MonaraMir:
Sets default FPS to 40 for new players instead of 0(server default, which is 20 currently)
Creates/Sets default rest keybinding to V
Sets default Move upwards keybinding to shiftC
Sets default Move downwards keybinding to ctrlC
Creates/Sets default subtler anti-ghost keybinding to 7

### DIFF
--- a/code/modules/keybindings/keybind/living.dm
+++ b/code/modules/keybindings/keybind/living.dm
@@ -24,5 +24,5 @@
 
 /datum/keybinding/living/rest/down(client/user)
 	var/mob/living/L = user.mob
-	L.resist_a_rest()
+	L.toggle_intentionally_resting()
 	return TRUE
--- a/code/modules/keybindings/keybind/living.dm
+++ b/code/modules/keybindings/keybind/living.dm
@@ -11,6 +11,12 @@
 	full_name = "Resist"
 	description = "Break free of your current state. Handcuffed? on fire? Resist!"
 
+/datum/keybinding/living/rest
+	hotkey_keys = list("V")
+	name = "rest"
+	full_name = "Rest"
+	description = "Toggle between lying down or standing up."
+
 /datum/keybinding/living/resist/down(client/user)
 	var/mob/living/L = user.mob
 	L.resist()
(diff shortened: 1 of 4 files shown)


code/modules/keybindings/keybind/living.dm
```diff
@@ -11,6 +11,12 @@
 	full_name = "Resist"
 	description = "Break free of your current state. Handcuffed? on fire? Resist!"
 
+/datum/keybinding/living/rest
+	hotkey_keys = list("V")
+	name = "rest"
+	full_name = "Rest"
+	description = "Toggle between lying down or standing up."
+
```
silicons, line 19: need /down() behavior

code/modules/keybindings/keybind/mob.dm
```diff
@@ -184,6 +184,12 @@
 	full_name = "Subtle"
 	description = "Does a subtle emote."
 
+/datum/keybinding/mob/subtler
+	hotkey_keys = list("7")
+	name = "Subtler Anti Ghost"
+	full_name = "Subtler Anti-Ghost"
+	description = "Does a subtle emote that's invisible to ghosts."
+
```
silicons, line 192: need /down behavior

code/modules/keybindings/keybind/living.dm
MonaraMir: Make sure I called the right proc on this one.

code/modules/keybindings/keybind/movement.dm
TheLordME, line 1: bay uses , and . for this

MonaraMir: I feel like ctrl/shift C is better for fighter combat, which is where it'll likely have the highest need for it to be accessible without moving the hand across the keyboard.

TheLordME: fair enough, just thought you'd like to know

TheLordME (approved):

PR #5504, opened 2023-05-16
reworks staggering effects

silicons:
## About The Pull Request

staggering now takes the highest amount instead of stacking slowdowns 

## Why It's Good For The Game

this makes staggering more predictable and therefore more useful to us as a whole, since we can use them without worrying about stacking with other staggering effects

## Changelog

:cl:
balance: slight changes to flash/emp staggers
/:cl:


TheLordME (approved):
its a silicons PR

PR #5540, opened 2023-05-29
Adds Some Monotiles

AlphaM01:
## About The Pull Request

Adds monotile versions of dark, techmaint and techfloor grid tiles.
They've a normal simulated turf and break if removed with a crowbar (exactly like normal monotile)

Screenshot_6260

## Why It's Good For The Game

Mappers noticed the absence of some monotiles. That shortcoming is now fixed, except for normal techfloor due to it's sprite requiring an even number of pixels for detailing.

## Changelog

:cl:
add: Added dark monotile turf.
add: Added techmaint monotile turf.
add: Added techfloor grid monotile turf.
/:cl:


TheLordME:
are you adding ways to obtain these or are they mappers only?

AlphaM01:
> are you adding ways to obtain these or are they mappers only?

Mappers only, which, as far as I can tell is also the case for the existing monotiles.

silicons (approved):

PR #5542, opened 2023-05-29
Modular Computer Revival

Mazianni:
## About The Pull Request

Re-enables the old modular computer TGUI modules - this was mostly an exercise in patience and fixing different variables, etc, between Polaris and CitRP. Most stuff :tm: works at the moment.

Camera consoles don't for reasons I don't understand (yet), so there is still work to be done here.

## Why It's Good For The Game

Modular computers are neat.

## Changelog

:cl:
fix: Re-enables modular computers and TGUI modules
tweak: ports communications and downloader TGUI module from Polaris 

/:cl:


code/modules/modular_computers/computers/modular_computer/core.dm
```diff
@@ -1,14 +1,13 @@
...
+/obj/item/modular_computer/process()
 	if(!enabled) // The computer is turned off
 		last_power_usage = 0
-		return FALSE
+		return 0
 
 	if(damage > broken_damage)
 		shutdown_computer()
-		return FALSE
+		return 0
```
silicons, line 8: regressions? more in the file too; you replaced true/false with 1/0

TheLordME (approved):

PR #5549, opened 2023-05-31
Chore: Converting all text("[]", foo) to "[foo]"

Zandario:
## About The Pull Request

I did this because some of the old text() uses would cause BYOND Beta version 515.1607 to no longer compile. (Which is a bug but still an excuse to finally deal with this)

## Beware

I hastily did this.

## Changelog

:cl:
code: Converted all text("[]", foo) to "[foo]"
/:cl:


code/_globals/text_macros.dm
Zandario, line 1: This was the absolute dumbest use of GLOB ever.

silicons (approved):

PR #5561, opened 2023-06-03
ghost follow selection

TheLordME:
## Changelog

:cl:
qol: No longer do you have to scroll through a list of 800 simple mobs and 3 player characters to find a player character
qol: "Call Communicator" now has a nicer list selection
/:cl:


silicons (approved):
mm

PR #5576, opened 2023-06-06
removes admins being able to see subtlers 

silicons:
## Why It's Good For The Game

i've had enough of the debate around this.

main server methodology of individual logging is best, but since no one has the time / energy / care to do that, i am going to just remove it instead until someone does.

until then, this is already logged, in the actual game logs. abuse can be quickly handled.

## Changelog

nothing player-facing.


ktoma36 (approved):

nik707:
*sigh

PR #5579, opened 2023-06-08
Hammerhead and Fighter Transit Tunnel

MonaraMir:
## About The Pull Request
Rebuilds hammerhead patrol barge with 6 spots for fighters (only 2 filled with barons at round start)
New tunnel for moving fighters from the construction bay to the Hammerhead, and vice versa.
Fixes all Elevators on the rift. You can now fall down them again.
Fixes offset lights, does not fix light duplication as the elevator picks up each floor's lights and moves them.
Various small fixes, typos, ect to clean up code that I saw while working.

## Why It's Good For The Game
Hammerhead with more carrying capacity aka more opportunities for robotics.

Fighters can now be constructed in the fighter bay (intended location) and flown/wheeled over to the Hammerhead via landing gear. (Yes, even space-only fighters can safely be transported with their "stabilization mode" toggled off, which turns on their landing gear.)

Hammerhead needed a touchup on some oversights, and a bit more wiggle room for upgrades. It's functionally the same as the previous iteration, but with more max speed (but with noticeably worse fuel efficiency)

Elevators need some love in terms of bugfixing. I did the mapping portion of it, now someone else can get glass to function without runtiming, and fix the lights spawning on every floor without the elevator present.


ktoma36:
Only thing that catches my eye with the map changes is how easy it would be for people to escape perma by just digging to the fighter tunnel. Otherwise just some misc things are the open turfs that go down to the base z level, isnt it supposed to be like 3 km deep or something? 
image

maps/map_files/rift/rift-03-underground1.dmm
ktoma36, line 1: I keep forgetting I can just leave comments trait on files but yeah, honestly we can probably either box in that lower spot of perma or something cause its just kind of a free escape as it is right now

ktoma36 (dismissed):

maps/map_files/rift/rift-03-underground1.dmm
MonaraMir: Those pickaxes SERIOUSLY suck. They're like excavation picks, multiple hits for one rock. Escape before this took around an hour and a half. This is now more like 45 minutes, and that's *if* security doesn't remove the pickaxes.

Also, once you're "free", you're still kinda stuck in the lower levels for a while, while you're also freezing to death. It's not easy to escape and survive.

PR #5589, opened 2023-06-10
persistent pictures framework

silicons:
no player facing changes

makes pictures persistent. fixes up the character saving system too.


TheLordME (approved):
l.ooks good from what I have seen
(I have no idea what it does in the end)

PR #5597, opened 2023-06-12
Adds resleeving and resurrection sickness

TheLordME:
## Changelog

:cl:
add: Added resurrection sickness, after being revived, lasting for 20 to 30 minutes
add: Added resleeving sickness, after being resleeve, lasting the rest of the round. 
/:cl:


code/datums/status_effects/status_effect.dm
```diff
@@ -227,6 +230,14 @@
 	var/id = initial(path.identifier)
 	return LAZYACCESS(status_effects, id)
 
+/mob/proc/has_alerting_status_effect()
+	var/list_length = LAZYLEN(status_effects)
+	for(var/i = list_length;i>=0; i--)
+		var/datum/status_effect/S = LAZYACCESS(status_effects, i)
+		if(S.alert_admins_when_leaving)
+			return S
```
Lohikar, line 238: This is a strange way to write this loop, it should probably be:
```suggestion
	for (var/datum/status_effect/S in status_effects)
		if (S.alert_admins_when_leaving)
			return S
```

TheLordME: My fears what the for will do when the list is null, since that is the default value for the variable, not empty list

Lohikar: `for (var/thing in null)` is valid, it's a no-op.

PR #5601, opened 2023-06-13
Take 2: Some fixes and QoL

TheObserver-sys:
## About The Pull Request
Would you believe me if I hadn't updated my git in about 400 years, and had to blow the old version of my repo up?
Yes? No? It doesn't matter.

Anyways! Meat and potatoes of this:
Allows players to make gene and plant discs freely in the protolathe. Since we do not have a dedicated genetics, this will help the pains of actually doing genetics by giving us storage solutions for genes.

Fixes a problem with brass also creating slag when compressing, by setting the copper alloy flag to 1.

And finally: Allows you to upgrade the braces! If your brace has T3 or better, a single brace can hold an entire drill. All credit goes to Hatterhat for this one, as I pretty much wholesale ripped it from his buff of the big drill™ on Virgo.

## Why It's Good For The Game
Not making slag is ALWAYS good. It saves on material, too.
Having more discs for a cheap cost is also good, it means you can reduce headaches while scoping out for genes, because there are many, and the ability to track them are currently few.
And honestly, the less lugging a person has to do with the mining drill, the more likely people might stop blowing up an already unstable planet with miniature hydrogen bombs.

## Changelog

:cl: The0bserver
add: Discs are able to be produced in the protolathe now. Go nuts, or don't. I'm not your guardian.
balance: Mining Drills can finally be operated with just one brace with the requisite parts. Thank you, Hatterhat!
fix: Copper no longer smelts slag when set to "Alloying."
/:cl:

### DIFF
--- a/code/game/objects/items/circuitboards/machinery/mining_drill.dm
+++ b/code/game/objects/items/circuitboards/machinery/mining_drill.dm
@@ -19,4 +19,7 @@
 	build_path = /obj/machinery/mining/brace
 	board_type = new /datum/frame/frame_types/machine
 	origin_tech = list(TECH_DATA = 1, TECH_ENGINEERING = 1)
-	req_components = list()
+	req_components = list(
+		/obj/item/stock_parts/manipulator = 1
+		)
+
--- a/code/modules/mining/drilling/drill.dm
+++ b/code/modules/mining/drilling/drill.dm
@@ -247,7 +247,7 @@
 		if(supports.len >= braces_needed)
 			supported = 1
 		else for(var/obj/machinery/mining/brace/check in supports)
-			if(check.brace_tier > 3)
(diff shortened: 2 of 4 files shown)


TheLordME:
you also never contributed here....
grafik

TheObserver-sys:
> you also never contributed here.... grafik

Yes, like I said. My old branch was like.. 2-3 years old and was horrendously out of date. Updating everything caused a hilarious issue -- so I went down the path of least resistance, blew everything up, and started again. Now, making fun aside, I don't know why it's throwing a runtime on the test when it runs without complaint on my end. Some insight would be nice on that end so we can move this along.

TheObserver-sys:
Think I got it. Stand by while I test again.

silicons (approved):

PR #5619, opened 2023-06-15
updates euthenia to use new map format

silicons:
@ktoma36


silicons:
Also throws out the two unused levels

ktoma36 (approved):

PR #5620, opened 2023-06-15
legacy map ports & maploader adjustments

silicons:
- all legacy maps (.dmms) are shoved into the right folders and brought up to speed. they are now linted for errors and can be webmap'd (soon as I bother aa07)
- maploader now supports mapping_helpers to hook themselves to fire before atom init. this is ideal for stuff like map generators like engine templates
- speaking of engine templates, they're now generic and can be map-specific

now the caveats

- i made shuttles into a dumpster fire because i'm not going to rework about 40 shuttles overnight. they *work* - just not well.
- i made some other things a dumpster fire because, yes, people kept deciding it's a good idea to copy/paste map defs.
- i nuked a lot of icon varedits and other unnecessary stuff in old maps. this isn't ideal but it's better to get those old maps in play than to be perfect since there's no way in hell i'm going to fix them "properly" in a timely manner.


TheLordME (dismissed):

code/controllers/subsystem/mapping/levels.dm
```diff
@@ -195,3 +226,34 @@
...
 	while(. in random_fluff_level_hashes)
 	random_fluff_level_hashes += .
+
+/**
+ * hooks us to SSmapping initializations; this should be called during New() for atoms.
+ *
+ * if no maploading can be hooked, we init immediately
+ * if Initialize() is in SSatoms, this crashes for safety as that should not happen.
+ */
+/datum/proc/hook_map_initializations()
```
Lohikar, line 236: Why is this so low level? This is very niche, and only one type seems to use it.

code/controllers/subsystem/mapping/levels.dm
```diff
@@ -195,3 +226,34 @@
...
+	if(isnull(SSmapping.map_initialization_hooked))
+		// postpone to after init
+		if(SSatoms.initialized == INITIALIZATION_INSSATOMS)
+			CRASH("undefined behavior: initialization is currently in SSatoms but we tried to hook map init.")
+		message_admins("a datum with map initializations was created. if this was you, you are in charge of invoking map_initializations() on it. this is not called by default outside of mapload as many things using the hook are highly destructive.")
+		// addtimer(CALLBACK(src, TYPE_PROC_REF(/datum, __immediate_map_initializations)), 0)
+	else
+		SSmapping.map_initialization_hooked += src
+
+/datum/proc/__immediate_map_initializations()
```
Lohikar, line 246: Why?

Lohikar:
This PR should really have a lot more description given how many things it's changing.

TheLordME (approved):

PR #5621, opened 2023-06-15
Nightshift Color/Brightness

MonaraMir:
## About The Pull Request

brightens nightshift lighting from 0.4 to 0.6 (dayshift is 0.8)

## Why It's Good For The Game

Nightshift was too dark, and blue adds atmosphere (It overwhelmingly won the vote too)

## Changelog

:cl:
tweak: Changes nightshift color to blueish
tweak: Changes nightshift brightness from 0.4 to 0.6
/:cl:

### DIFF
--- a/code/__DEFINES/color/lights.dm
+++ b/code/__DEFINES/color/lights.dm
@@ -34,4 +34,4 @@
 #define LIGHT_COLOR_INCANDESCENT_TUBE       "#E0EFF0" // rgb(224, 239, 240) Slightly blueish white.
 #define LIGHT_COLOR_INCANDESCENT_BULB       "#FFFEB8" // rgb(255, 254, 184) Slightly yellowish white.
 #define LIGHT_COLOR_INCANDESCENT_FLASHLIGHT "#FFCC66" // rgb(255, 204, 102) Slightly yellowish white.
-#define LIGHT_COLOR_NIGHTSHIFT              "#EFCC86" // rgb(239, 204, 134) Slightly yellowish white.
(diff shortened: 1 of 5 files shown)


code/__DEFINES/color/lights.dm
```diff
@@ -34,4 +34,4 @@
 #define LIGHT_COLOR_INCANDESCENT_TUBE       "#E0EFF0" // rgb(224, 239, 240) Slightly blueish white.
 #define LIGHT_COLOR_INCANDESCENT_BULB       "#FFFEB8" // rgb(255, 254, 184) Slightly yellowish white.
 #define LIGHT_COLOR_INCANDESCENT_FLASHLIGHT "#FFCC66" // rgb(255, 204, 102) Slightly yellowish white.
-#define LIGHT_COLOR_NIGHTSHIFT              "#EFCC86" // rgb(239, 204, 134) Slightly yellowish white.
+#define LIGHT_COLOR_NIGHTSHIFT              "#616191" // rgb(134, 162, 239) Blueish white.
```
deathride58, line 37: The RGB value in the comment does not match the color here at all.
Additionally, the sheer darkness of this color affects the resulting light output in ways that might not be intentional; for instance, this is around 50% value, which means that the resulting light is, in fact, not increased to 0.6 power, but rather, effectively reduced to 0.3 (!!!) power. Light mixing might also be an issue with a light color this dark, though admittedly I can't recall which lighting backends play well with dark light colors

MonaraMir: Seems to function well when it comes to the testmerge as well as locally. Comment/RGB was fixed

ktoma36 (approved):

PR #5632, opened 2023-06-18
removes the wood recipe for the (bone) horse post

timothyteakettle:
## About The Pull Request
it's a horse post made out of bone and sinew found in the crafting menu intended for ashlanders to make use of
it gets deconstructed into bone still which is what caused it to get reported in bug-triage in the first place

## Why It's Good For The Game
it's um made of bone so you should probably not be able to make it from wood (but someone making a wooden one would be neat!)

## Changelog
:cl:
del: removes the recipe for the horse post that requires wood, because it is made out of bone and sinew
/:cl:


timothyteakettle:
requesting a review from Captain as their PR added it just in-case there is some reason for the recipe to exist

Captain277:
Oh what a huge oversight. Thank you for catching this!

Captain277 (approved):

PR #5640, opened 2023-06-20
You can now handcuff a lying mob

Shadowtail117:
## About The Pull Request

You can now apply handcuffs on a mob that is lying down. This can be because they are intentionally lying down, because they collapsed from being magdumped by shitsec, etc.etc.

## Why It's Good For The Game

This should ideally make it easier for security to arrest someone after the fight is over. Currently, even when someone is lying down and in painshock they can still mash resist to stop you from getting a grab on them, which means you have to keep tasing them until they pass out. That is probably not ideal, so this makes it much easier to handcuff a downed suspect (and also makes high-risk arrests a little easier to pull off).

## Changelog

:cl:
add: You can now handcuff someone who is lying down without needing to grab them first.
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/handcuffs.dm
+++ b/code/game/objects/items/weapons/handcuffs.dm
@@ -46,6 +46,8 @@
 /obj/item/handcuffs/proc/can_place(var/mob/target, var/mob/user)
 	if(user == target)
 		return 1
+	if(target.lying) //Mobs that are lying down can be handcuffed without needing to grab them, to make arrests a little easier
+		return 1
 	if(istype(user, /mob/living/silicon/robot))
 		if(user.Adjacent(target))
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/handcuffs.dm
```diff
@@ -46,6 +46,8 @@
 /obj/item/handcuffs/proc/can_place(var/mob/target, var/mob/user)
 	if(user == target)
 		return 1
+	if(target.lying) //Mobs that are lying down can be handcuffed without needing to grab them, to make arrests a little easier
```
silicons, line 49: this should be target.resting, not target.lying

lying is strictly for transform and can be 0, 90, -90, etc

PR #5644, opened 2023-06-22
updates rust_g; adds rustg-based icon_states; fixes unit tests

silicons:
~~so back when i did inventory slot refactor i added this thing called a render state cache
it basically allows for lazy people to default when needed and don't default when not needed~~

~~problem is
to facilitate this we need to icon_states
which seems to be using just an exorbitant amount of memory.~~

byond icon_states bad
replaced with rustg icon states
there's a massive footgun in that it's obviously going to be slow due to the fcopy() but for the use case of inventory it should be fine.

frees about 400MB.

also fixes us just having bad shuttle paths

### DIFF
--- a/citadel.dme
+++ b/citadel.dme
@@ -338,6 +338,7 @@
 #include "code\__HELPERS\icons\download.dm"
 #include "code\__HELPERS\icons\flatten.dm"
 #include "code\__HELPERS\icons\hologram.dm"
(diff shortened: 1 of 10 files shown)


timothyteakettle (approved):
(this is an extremely dangerous PR)

ktoma36 (approved):

PR #5655, opened 2023-06-25
bumps projectile speed up

Mazianni:
## About The Pull Request

tweaks projectiles to not be so slow

## Why It's Good For The Game

we are not all neo.

## Changelog
fast boolet.


timothyteakettle:
personally for faster projectiles but this probably needs discussing / testmerging to play with the value a bit to get it right

silicons (approved):

PR #5673, opened 2023-07-04
Fixes saddlebags

DiscordWizard:
## About The Pull Request

Saddlebags apparently could be worn by non taurs and had no sprite once worn, this fixes both those issues and uncomments them from the loadout.

Works fine on a test branch, non-taur received the error message and couldn't wear it while drake taur and other taurs used to test it with could wear it with an icon showing up for every saddlebag.

The one saddlebag removed was pretty redundant and older than the rest of the saddlebag code, it's basically saddlebag_common but specifically for horse.

## Why It's Good For The Game

Well the things were broken and unusable and now they are working and usable, I'd say it's a net positive.

## Changelog

:cl:
fix: fixed saddlebags
/:cl:


timothyteakettle (changes requested):
this does not compile so you have not tested it from the latest commit (check the path names between loadout and the saddlebag item)

timothyteakettle:
going to close this as no changes made in the last week after requesting changes and no reply after pinging them on the discord

PR #5678, opened 2023-07-06
fixes a tgui cardmod issue

silicons:
i fucked up, this results in dupes.


TheLordME (approved):
Why are cardmon critical prio? Does anyone even use them?

PR #5686, opened 2023-07-08
readds rd/sar rigs, adds a rig to the CMO, repaths modules/rigsuits to modules/hardsuits in preparation for the rig update

silicons:
restores stealth removals of old rigs
gives the CMO a rig because why not
and gets the files out of the way so I can implement new rigs


TheLordME (approved):
yolo

PR #5702, opened 2023-07-15
fixes closets being wonky

Mazianni:
## About The Pull Request
fixes closets not rendering in the map right / some having no icon


TheLordME (dismissed):

TheLordME (approved):

timothyteakettle:
hold of on merging this until it's properly tested just to avoid the potential issue of all lockers suddenly being invisible

timothyteakettle (changes requested):
bodybags do not use closet_appearance and as a result do not have a sprite with this change
edit: this was fixed in "Baywall Bugfixing #5685" so this is redundant everything is fine here

timothyteakettle (approved):

PR #5713, opened 2023-07-16
TRIUMPH SECURITY EQUIPMENT STORAGE, ARMORY, AND FIGHTER BAY REWORK

Nut2:
## About The Pull Request
Puts plate carriers, tags, plates, and helmets, as well as helmet covers into the sec equipment storage
revamps the armory with the same gear as the atlas
replaces the shitty shuttles within the security fighter bay with the barons from the atlas

## Why It's Good For The Game
it sucked before :(


TheLordME:
A couple things:

1. grafik
2. dont add items to preset lockers (the security officer lockers in this case) via mapping. add them in the definition instead. You can find the definition of contents in file `code\game\objects\structures\crates_lockers\closets\secure\security.dm` the list you want to edit starts at line 322, add a comma at the last entry and add the warrant projector or what ever you are adding to the lockers after that, make sure you are still in the parenthesis. I could live with you just removing the items from the lockers in this PR, and making a seperate PR to add them to the definition.
3. decals on floors and walls where they dont make sense
grafik
4. Firealarm was removed (not an actual issue, but maybe add a new one somewhere
grafik

PR #5722, opened 2023-07-17
message range update to visible_action_feedback, and makes subtles make a, well, subtly-does-something message for people within range instead of being completely silent

silicons:
## About The Pull Request

you will now see "x does something subtly" in small font when someone subtles onscreen but you can't see it.

## Why It's Good For The Game

we had a discussion about subtles

we are obviously not removing subtles
that said, subtles, unlike whispers, have no tell.
this adds a tell without invading too much privacy.

## Changelog

:cl:
add: subtles now have a subtle message to people who can see it, akin to "someone whispers something".
/:cl:

### DIFF
--- a/citadel.dme
+++ b/citadel.dme
@@ -951,6 +951,7 @@
 #include "code\game\area\Space Station 13 areas.dm"
 #include "code\game\area\ss13_deprecated_areas.dm"
 #include "code\game\area\Tether_areas.dm"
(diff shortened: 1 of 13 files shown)


silicons:
DNM until vore-only subtle gets added so this doesn't hit that part of the population.

TheLordME (approved):

PR #5727, opened 2023-07-17
pAI clothing adjustments

timothyteakettle:
## About The Pull Request
1. clothing name now has (pAI) appended to it because i really don't want someone to name their pAI after an item of clothing and be indistinguishable, that's not okay
2. when you use clothing on a pAI, it notifies the person who did so if it was successful, the pAI gets a similar message
3. copied clothing now copies the colour of the item as well

## Why It's Good For The Game
2/3 are some good qol stuff, 1 is because we have a clear rule of "don't be a creep", it's a ghost role, and is easily open for abuse which i would rather not happen

## Changelog
:cl:
tweak: pAI clothing has (pAI) appended to its name, copies colour when copying other clothing items, and notifies the pAI and user when you scan a piece of clothing on a pAI
/:cl:


timothyteakettle:
i'll make this not a draft once i've tested the changes

silicons (approved):

PR #5742, opened 2023-07-20
fixes using items on lathes like rpeds

silicons:
atleast if they don't allow insert item

todo: make insert item return TRUE / FALSE / NULL and only allow it on null.


TheLordME (approved):
woooo

PR #5745, opened 2023-07-20
adds maint access to most jobs and reorganizes + shuffles some access

silicons:
should result in no live server changes
organizes and reshuffles access to make more semantic sense in context of additional vs minimal access
everyone now has maint access so we can start putting maint access on maint airlocks on the map.


TheLordME (approved):
woooo

PR #5796, opened 2023-07-28
Important Update

Zeldazackman:
## About The Pull Request

Just a small but important update that was missing for whatever reason.

Swiped from Polaris and Verks

## Why It's Good For The Game

Fixes clowns being able to constantly spam their horns sometimes while at the same time adding in the ability for admins to change the honk sounds on a bike horn.

Yes, you can finally have your ASMR deep-fried garbage on the go.

## Changelog

Fixes honk spam and adds in the variable for honk_sound allowing bikehorn.ogg to be edited by admins.


Zeldazackman:
I'll try this again later when I'm not exhausted.

PR #5817, opened 2023-08-02
Surt Event Wrap Up

CharlesWedge:
## The Xenos have been successfully repelled from Surt
And the galaxy keeps on moving. This small PR adds more news stories, three in total. Also the Scorians have a new suit of relic armor as a sign of victory over the invading demons from the stars above. Not much else to say really only now I need to think up more fun events for the gang.

:cl:
add: a single suit of armor for the Scori, placed in the temple where a relic like it belongs.
add: Three new news stories. Maybe something is going wrong with the economy?
/:cl:


BlueWildrose (changes requested):
A lot of spelling and typos to be corrected. If you're using an IDE like Visual Studio Code, consider getting a spellchecker plugin possibly.

TheLordME (approved):

PR #5835, opened 2023-08-05
map rotation

silicons:
fixes map rotation bugs ;)


code/modules/mapping/map.dm
```diff
@@ -90,7 +90,7 @@
 			levels += is_it_a_path
 			continue
 		// json
-		if(serialized_level[1] == "{}")
+		if(serialized_level[1] == "{")
```
TheLordME, line 93: is this intended?

TheLordME (changes requested):

PR #5876, opened 2023-08-15
several pAI fixes, ability to scan and display things as a hologram, add accessories to clothing list, ups pAI downtime when carded

timothyteakettle:
## About The Pull Request
alt click -> scan object (60 second cooldown because its an expensive operation)
adds the image to a list, if it surpasses 10 items it deletes the first (oldest) item in the list

image
image
image

also adds accessories to clothing options not sure why i didn't include that originally

fixes a bug with transforms on rest for pAIs

fixes a bug with wiping pAIs

ups downtime of pAI from 6 seconds to 20 seconds when their shell is attacked

fixes wide sprites like alraune and taurs causing the pAI to be offset incorrectly

fixes weird case where you would rest, fold, unfold, and have resting = 0 but intentionally_resting = 1

## Why It's Good For The Game
i promise to atomize the pr the next time i work on pAIs
also this is 90% bug fixes

## Changelog
:cl:
add: allows pAIs to scan objects and mobs then display them as a hologram
add: adds accessories to the pAI clothing options
fix: fixes your size being reset as a pAI when resting in holo form
balance: pAIs now get deactivated for 20 seconds up from 6, when taking enough damage to be downed, this was way too low a downtime for suffering a pseudo-death
fix: wiping a pAI no longer creates a corpse that doesn't exist (garbage collection was NOT happy)
fix: pAIs using a character holoform for things like alraune/taurs will now be offset correctly to appear like other characters, in the middle, and not offset to the right
fix: resting, folding, then unfolding, now doesn't treat you like you are standing and resting at the same time
/:cl:


timothyteakettle:
ignore linters they're crying over something stupid that i'll change before this is full merged

timothyteakettle:
linters hate me updating icons in initialize apparently there's a call to stoplag being made somewhere

timothyteakettle:
probably ready now

timothyteakettle:
i promise future PRs will be more atomized than this but there was still quite a lot of old bugs in play since before i started work on pAIs so it was easier to just stick everything i was working on under one PR for pAIs

TheLordME (approved):

PR #5883, opened 2023-08-17
character serialisation SQL changes

timothyteakettle:
## About The Pull Request
throwing a draft up for the planned changes and so we have somewhere to document them:

### **new character lookup table with the columns:**
**charactername**: a character's name, ckey() is applied for formatting
**playerid**: the user's ckey
**characterid**: the identifier for a character, automatically generated from charactername and playerid

### what this means:
characters with the same name will by default have the same character id, which we can use for persistence
if you have sleeves that are different names, we can manually alter the character id's to be the same, so they share persistent data between them

### other stuff
migration will be done to update this table for existing characters
playerid, character name, character type is the primary key

~~character id will get automatically updated if you change a character's name (either changed, if no key exists for your ckey and the new character id, or removed if it does, because it's a primary key constraint)~~
kevin said don't do this so i didn't, nothing happens if you edit your character's name now which is, a bit strange i guess??


timothyteakettle:
everything is working except renaming characters

PR #5906, opened 2023-08-23
better language filtering checks

silicons:
fixes runtimes coming from someone changing ids


code/__DEFINES/languages/ids.dm
TheLordME, line 1: As if anyone reads your comments

TheLordME (approved):

PR #5911, opened 2023-08-24
Completely changes how DCS types works

silicons:
DCS now registers based on its registered type, as opposed to at every subtype.
Dupe mode is no longer considered realistically usable unless if-and-only-if all var pre-setting behavior is considered - which is not the case on the majority of citrp-made components outside of /datum/component/radioactive

## Why?

Components are a generic way to attach datums to things.
The common practices included:
- No getcomponent
- Use signals for all interaction
- Don't subtype, set vars on initialization

This resulted in components, while being quite stable, being also quite rigid, in my opinion. Given we don't use components for the same things, and probably never will, I think it's better for us to get the optimizations from not supporting what's honestly behavior that we shouldn't rely on in the first place - if something needs to be applied multiple times (e.g. radiation) it should rely on the old behavior and this new system still allows it. If it doesn't, the author should either not be adding the component multiple times, or **making** their component functional under this system. InheritComponent still has all the tools necessary to properly do component inheritance, it's just now we default to all components only being considered the same if it's the exact subtype - which from what I could see, is also the previous case (as if you add a subtype, uh, bad shit's going to happen if the base type is registered but not the subtype anyways!)

### DIFF
--- a/code/datums/components/_component.dm
+++ b/code/datums/components/_component.dm
@@ -14,6 +14,9 @@
 	 * Defines how duplicate existing components are handled when added to a datum
 	 *
 	 * See [COMPONENT_DUPE_*][COMPONENT_DUPE_ALLOWED] definitions for available options
+	 *
+	 * Dupe detection operates on registered_type. If you use components with subtypes,
+	 * you shouldn't be relying on dupe_mode or argument passing at all.
 	 */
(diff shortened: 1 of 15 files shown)


TheLordME (approved):
I guess

PR #5924, opened 2023-08-26
Restores Old Biohazard Bin

Captain277:
## About The Pull Request

1. **Restores Old Biohazard Bin sprite.**

## Why It's Good For The Game

1. _I really don't like the new biohazard bin. It has none of the depth and spirit of the original. It looks like a plain laundry cart. So I've added the old biohazard crate back in._

## Changelog

:cl:
add: Adds old Biohazard bin.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/crates.dm
+++ b/code/game/objects/structures/crates_lockers/crates.dm
@@ -713,3 +713,12 @@
 /obj/structure/closet/crate/secure/zenghu
 	desc = "A secure sterile crate marked with the logo of Zeng-Hu Pharmaceuticals."
 	closet_appearance = /singleton/closet_appearance/crate/branded/zhenghu/secure
+
+//Brings back the old Biohazard Bin.
+/obj/structure/closet/crate/biohazard
+	name = "biohazard bin"
+	desc = "A large bin for medical and other personnel that deal with bio-grade waste."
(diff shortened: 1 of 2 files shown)


code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -713,3 +713,14 @@
...
+//Brings back the old Biohazard Bin.
+/obj/structure/closet/crate/biohazard
+	name = "biohazard bin"
+	desc = "A large bin for medical and other personnel that deal with bio-grade waste."
+	icon = 'icons/obj/storage.dmi'
+	icon_state = "biohazard"
+	climb_allowed = TRUE
+	points_per_crate = 5
+
+/obj/structure/closet/crate/biohazard/LateInitialize()
```
silicons, line 726: why does htis need to lateinitialize?

silicons (changes requested):
...?

code/game/objects/structures/crates_lockers/crates.dm
Captain277: Due to how crates are now coded, if I don't cancel LateInit on this bin it gets converted into one of the new tiny crates, which is what this whole PR is trying to stop.

silicons (approved):

PR #5939, opened 2023-09-01
Old Tether

silicons:
Restores old tether.

This is on `old_tether` branch on the repository, so PRs to it is possible.


silicons:
tell me when we reach a decision on this

PR #5940, opened 2023-09-02
The Hive Awakens

CharlesWedge:
## Oh No More Robots
There is actually less paths for the hivebots. They are actually some of the most primitive mobs on the codebase. So it was high time they were given a facelift. As I said with my previous mob update robots are a good alternative as mobs compared to humanoids, and with the hivebots we can present a threat of hostile machine intelligence to round out the existing threats of pirates, mercs, aliens beasts and the supernatural. Once more these robots are also far mroe generalist then the existing robot varieties and as most types of them are not very dangerous they can be released on civilian crew without fear of them causing extreme damage,

## Changelog

:cl:
add: A couple new varieties of both melee and ranged hivebots
removed: redundant hivebot varieties
tweak: siegebots now have sniper range fitting their name, their attack has been nerfed (holy fuck the one shoot explode on contact grenades with a base attack of 10... that's 1 frag grenade a second!!!)
fix: hivebots now use their various cataloguer entiries
sprites: hivebot types are now more visually distinct
/:cl:


Captain277:
You've got an error on /DrugDen. There's a hivebot mapped in there that no longer exists.

Captain277 (approved):

PR #5958, opened 2023-09-08
[IDB IGNORE] [MDB IGNORE] Euthenia final stretch

ktoma36:
## About The Pull Request

(Hopefully) The final Stretch before I can start doing proper playtest and eventually the final merge of Euthenia


ktoma36:
My brain is goo, apologies for slow progress

code/modules/recycling/conveyor2.dm
```diff
@@ -29,6 +29,9 @@
 /obj/machinery/conveyor/centcom_auto
 	id = "round_end_belt"
 
+
+/obj/machinery/conveyor/on
```
TheLordME, line 33: ????

code/modules/recycling/conveyor2.dm
```diff
@@ -218,6 +221,8 @@
 	var/list/conveyors		// the list of converyors that are controlled by this switch
 	anchored = 1
 
+/obj/machinery/conveyor_switch/two_way_on
+	position = 1
```
TheLordME, line 225: I am not sure why we would want default on conveyors/conveyor switches

code/modules/recycling/disposal.dm
```diff
@@ -493,6 +493,38 @@
 	else
 		return ..(mover, target)
 
+
+
+/obj/machinery/disposal/wall
+	name = "inset disposal unit"
```
TheLordME, line 499: did you test if this works? I am not sure if the bump proc is called when you are on a wall tile

TheLordME (commented):
cant really look at the map with MDB ingnore on

code/modules/recycling/disposal.dm
ktoma36: They work quite well, ive not noticed anything with bumping into them though. Is the bump proc tripping into one?

code/modules/recycling/conveyor2.dm
ktoma36: Mostly so I didnt have to make a var edit version, I wanted one that would be default on

code/modules/recycling/conveyor2.dm
ktoma36: Ah forgot about that, removing it now

TheLordME (approved):

PR #5967, opened 2023-09-11
makes sonar pulses cooldown faster

silicons:
remember that sonar only persists for like a half second
if we boost the image persist time we should make this slower but uh, it's kind of hard to use without


Dimasw99:
I always felt like the sonar fades away too fast to actually figure out the things around you. Encouraging to use the ability several times in a row

silicons:
@Dimasw99 it was actually supposed to, but i'm not against just upping the fade duration if people think it's too short as i recognize that most people don't really like the super fast flicks

ktoma36 (approved):

TheLordME (approved):

PR #5969, opened 2023-09-12
blobform reworks

silicons:
for both protean and promethean

they no longer drop items but can be stripped
furthermore, radio works now


code/modules/species/protean/protean.dm
```diff
@@ -339,5 +339,25 @@
 			src.forceMove(prig)
 			return
 
+/mob/living/carbon/human/verb/rig_self()
+	set name = "Deploy Nanosuit To Self"
+	set desc = "Deploy a light nanocluster RIGsuit around yourself."
+	set category = "Abilities"
+
+	if(istype(back, /obj/item/hardsuit/protean))
```
TheLordME, line 347: what if you are wearing another protean?

TheLordME (approved):

TheLordME (approved):

PR #5974, opened 2023-09-13
shuttles: turf grids

silicons:
refactors shuttle movement
adds support for shuttle rotation


silicons:
replaced by #shuttlesv2

PR #6012, opened 2023-09-22
RPED 2: You mean I can fast replace cells and beakers with this?

TheObserver-sys:
## About The Pull Request
Once this is completed, we will be able to slot cells and beakers into all part replacers, and replace them with better versions.

The tiering system may seem weird, but I personally believe that the slime cell is our T5 for machines, as it auto-recharges, as opposed to the hyper which does not.

Beakers will have bluespace at T3/T4, once it's bothered to be done.

## Why It's Good For The Game
Our RPEDS need this upgrade, and it's about time we give it to them. Eventually, we are gonna wanna figure out how to make them more efficient, especially if we ever wanna make the real SLIM--  BSRPED 

## Changelog

:cl:
add: Your Local RPED has been upgraded! Congratulations! You can now replace cells and beakers as you go about. Really, it's about time.
/:cl:


code/modules/power/cell.dm
```diff
@@ -44,6 +45,9 @@
 		STOP_PROCESSING(SSobj, src)
 	return ..()
 
+/obj/item/cell/get_rating()
+	return rating
```
TheLordME, line 49: wouldnt it make sense to just use the maxcharge var instead of an extra new var?

code/modules/reagents/reagent_containers/glass.dm
TheLordME, line 1: you are not even giving the most important beakers a rating

TheLordME (approved):

PR #6018, opened 2023-09-24
Makes bananium, silencium actually spawnable - retypes chitin to be a material

BlueWildrose:
## About The Pull Request

- Gives bananium and silencium material properties, thus allowing them to actually spawn, whether by in-game or admin means. They couldn't before.
- Retypes insect chitin to become a material - because material chitin seemed to have existed already, which imply this was in plans probably. Spiders actually did have "chitin" as their bone type when you butchered them, but it was a material chitin that didn't work until now. Scorians rejoice, you can make chitin foldable chairs for whatever reason.
- This also changes most materials to have their IDs use their defines for code consistency.

Open to suggestions regarding the material's properties if they could be changed at all.

## Why It's Good For The Game

Fix good.

## Changelog

:cl:
fix: Bananium and silencium is actually spawnable now as they've been given material properties.
fix: Chitin has become materialized. This also means spiders will now yield chitin when butchered too.
code: some material IDs now using defines, though it's not all of it
/:cl:

### DIFF
--- a/citadel.dme
+++ b/citadel.dme
@@ -3177,6 +3177,7 @@
 #include "code\modules\materials\definitions\misc\snow.dm"
 #include "code\modules\materials\definitions\misc\wax.dm"
 #include "code\modules\materials\definitions\organic\bone.dm"
+#include "code\modules\materials\definitions\organic\chitin.dm"
 #include "code\modules\materials\definitions\organic\cotton.dm"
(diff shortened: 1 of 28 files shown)


TheObserver-sys:
thank you <3

TheLordME (approved):

PR #6023, opened 2023-09-25
life crystals are no longer unlimited range

TheLordME:
## Why It's Good For The Game
why life crystals should be nerfed:

1. Funny crystal rubbing and its able to detect your death from planets away is stupid
2. So far the lore (ingame description) suggests that its used to inform a thrid party(veymed) about the death of their customer, it should be worn on the customer, not left somewhere.
3. Currently its a loadout item for 1(one) loadout point, while practically an infinit range death alarm, death alarm implants are, opposed to that, a supply crate, that costs 40 points for 7, and death alarms work only while someone is in radio range of you

## Changelog

:cl:
tweak: tweaked life crystals working over unlimited range
/:cl:

### DIFF
--- a/code/modules/vore/fluffstuff/custom_items.dm
+++ b/code/modules/vore/fluffstuff/custom_items.dm
@@ -463,7 +463,8 @@
 	//He's dead, jim
 	if(state != 1)
 		return
-	if(owner && (in_range_of(src, owner,7)) && owner.stat == DEAD)
+	var/list/connected_z_levels = GetConnectedZlevels(get_z(src))
+	if(owner && (owner.z in connected_z_levels) && owner.stat == DEAD)
 		update_state(2)
 		audible_message("<span class='warning'>The [name] begins flashing red.</span>")
 		sleep(30)
--- a/code/modules/vore/fluffstuff/custom_items.dm
+++ b/code/modules/vore/fluffstuff/custom_items.dm
@@ -463,7 +463,7 @@
 	//He's dead, jim
 	if(state != 1)
 		return
-	if(owner && (owner in src.view) && owner.stat == DEAD)
+	if(owner && (in_range_of(src, owner,7)) && owner.stat == DEAD)
 		update_state(2)
 		audible_message("<span class='warning'>The [name] begins flashing red.</span>")
 		sleep(30)
--- a/code/modules/vore/fluffstuff/custom_items.dm
+++ b/code/modules/vore/fluffstuff/custom_items.dm
@@ -461,7 +461,9 @@
 
 /obj/item/clothing/accessory/collar/vmcrystal/proc/check_owner()
 	//He's dead, jim
-	if((state == 1) && owner && (owner.stat == DEAD))
+	if(state != 1)
+		return
+	if(owner && (owner in src.view) && owner.stat == DEAD)
 		update_state(2)
 		audible_message("<span class='warning'>The [name] begins flashing red.</span>")
 		sleep(30)


timothyteakettle (changes requested):
if you're dead set on having it be nerfed so it's not as strong as a death alarm, make it so it works on the same Z level or on station at least, instead of 7 tiles of range

the speed of which this went from an ooc discussion to a pr makes it feel incredibly kneejerk but whatever

TheLordME:
> if you're dead set on having it be nerfed so it's not as strong as a death alarm, make it so it works on-station instead of 7 tiles of range
> 
> the speed of which this went from an ooc discussion to a pr makes it feel incredibly kneejerk but whatever

its what every Product manager dreams off, Time to market < 1 day

BlueWildrose:
> if you're dead set on having it be nerfed so it's not as strong as a death alarm, make it so it works on-station instead of 7 tiles of range
> 
> the speed of which this went from an ooc discussion to a pr makes it feel incredibly kneejerk but whatever

Funnily enough, the death alarm isn't even all that powerful. It only works if there's comms present.

In other words, if you're at Surt or some off-site area of some sort - in one of the places where you would reasonably _want_ a death alarm, it wouldn't really work.

This PR basically makes it so that if you died at lavaland and there was nobody else present but ERPers, you're pretty much fucked.

BlueWildrose:
> But then the question becomes, should it be a loadout item?

Considering that ultimately, Cit-RP describes itself as "slice of life" on the hub, and antags aren't really a thing outside of events, why should we make it harder for dead bodies to be discovered to begin with?

TheLordME:
I would rather fix the transcore than have the life crystal play an interplanetary instant transmission trinket

BlueWildrose:
> I am personally of the opinion that if explo (or off station mining) dies, they deserve to be dead and taken out of the round.
> I have said before that resleeving takes a lot of the risk out of "high risk" jobs, with the mirror update fixing it abit.
> 
> plus, if you go out alone you clearly dont want to be revived fast,
> but what if there is nobody to accompany you? who is gonna rescue you? the guy that isnt there to go with you to begin with?

- Most of the time explorers are never going alone unless they're being an idiot. There's a high chance that they aren't getting left behind.
- Mining jobs almost _require_ that you go out of comms range almost all of the time. This is 100% of the time on Triumph - as you're on a ship map and thus your only options is either Surt or belt mining. To say that miners deserve to stay dead if they screw up at all - for example, **lagging into lava** seems a bit harsh. Considering how rarely mining is played for mysterious reasons, they more often than not will end up going alone because of such too. Even if on a planet map, in order to gather all the materials for science, you'll need to dig deep to get metallic hydrogen - the one mineral that can't be obtained outside of Surt or the vastly forgettable mineral rich planet.
- Resleeving is an issue with the revival process. It has nothing to do with the process of discovering a body.

> I would rather fix the transcore than have the life crystal play an interplanetary instant transmission trinket
Then fix it. I'd rather we create new solutions to these problems before we decide to nerf something. For instance, the Surt mining base could have any form of communication to inform those in the facility that something bad happened.

pixelkitty286:
I'm against this PR for the reason of: It causes a very high risk of characters being PKed due to their bodies not be discovered. The mechanical point of wearing the life crystal is so it notifies off sight cloning of your demise when you are off sight, not to notify the station.

TheLordME:
the point of this PR is to unmake the life crystal a magic "rub your finger against me and I know if you die through space and time no matter where"
#6024 adds alternative means to inform the station about the untimely demise, you just need a mirror installed for it to work

TheLordME:
#6025 is another alternative

yourdoom9898:
> I'm against this PR for the reason of: It causes a very high risk of characters being PKed due to their bodies not be discovered. The mechanical point of wearing the life crystal is so it notifies off site cloning of your demise when you are off site, not to notify the station.

PKs (outside of admin intervention) are entirely a player choice. Just dying out as a miner/explo doesn't mean your character permanently dies, that's why you have to file a DNR to keep NanoTrasen from **_cloning you_**. The reason this change was made is because people were using the fluff item for notifying off site cloning as a ghetto death alarm because the item was shitcoded by VORE, and therefore didn't actually check if it was being worn by it's wearer when it breaks if the wearer dies, so people could leave them in the lobby of Medical and if they shattered, Medical knew they died.

silicons:
gamers stop testmerging prs that are under discussion

PR #6035, opened 2023-10-01
Makes air alarms tolerate tiny amounts of phoron

SingingSpock:
## About The Pull Request

Sets the max1 threshold for phoron to 0.25, making air alarms not freak out at the tiniest amount of the gas

## Why It's Good For The Game

Vox breathing and solved air crises won't permanently make a room orange atmos alert

## Changelog

:cl:
fix: Air Alarms no longer freak out at infinitesimally small amounts of phoron
/:cl:

### DIFF
--- a/code/modules/atmospherics/gasmixtures/gas_types/core.dm
+++ b/code/modules/atmospherics/gasmixtures/gas_types/core.dm
@@ -79,4 +79,4 @@
 	gas_flags = GAS_FLAG_FUEL | GAS_FLAG_FUSION_FUEL | GAS_FLAG_CONTAMINANT | GAS_FLAG_FILTERABLE | GAS_FLAG_CORE
 	gas_groups = GAS_GROUP_CORE
 
-	default_tlv = list(0, 0, 0, 0.5)
+	default_tlv = list(0, 0, 0.25, 0.5)


SingingSpock:
Anyone know why it keeps adding old commits when I make new branches?

Putnam3145:
So, personally, the way I would do it is not to give it a phoron-specific 0.25 thing but make it so that there's some epsilon of tolerance--i.e. if it says "0", it actually means "allow up to 0.001 moles", which should generally be fine. More universal than just the phoron thing, too.

silicons (approved):

SingingSpock:
> So, personally, the way I would do it is not to give it a phoron-specific 0.25 thing but make it so that there's some epsilon of tolerance--i.e. if it says "0", it actually means "allow up to 0.001 moles", which should generally be fine. More universal than just the phoron thing, too.

That'd be nice, but I don't know how to do that. Phoron is currently the only gas that's set to 0 for its min1

SingingSpock:
Could we get this un-TMed so that we can make sure #6037 works for our intended purposes without it?

TheLordME:
#6037 was tested locally, but I also took this one from the TM list, so you are free to test it on the server next round (due to the values I used you need to scrubb the atmosphere for a bit, to get below the threshold) but it should get out of warning level within less than 10 minutes when using a portable scrubber)

SingingSpock:
> #6037 was tested locally, but I also took this one from the TM list, so you are free to test it on the server next round (due to the values I used you need to scrubb the atmosphere for a bit, to get below the threshold) but it should get out of warning level within less than 10 minutes when using a portable scrubber)

Having to wait 10 minutes for a room to not freak out after a Vox crew member exists in it isn't really going to solve the issue this is intended to solve, then.

PR #6044, opened 2023-10-02
Makes radiation wave decay inverse square instead of exponential

Putnam3145:
## About The Pull Request

As title. Right now it decays exponentially instead of through an inverse square law.

## Why It's Good For The Game

It... makes radiation stronger, actually. Not sure this is better. But engineers should be wearing rad suits when there's radiation anyway, so, y'know. Probably good.

## Changelog

:cl:
fix: "Inverse square factor" for radiation now actually inverse square instead of exponential decay
/:cl:

### DIFF
--- a/code/datums/radiation_wave.dm
+++ b/code/datums/radiation_wave.dm
@@ -140,7 +140,7 @@
 	var/dir
 	var/spread
 	var/existing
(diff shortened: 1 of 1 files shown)


silicons:
this is a drastic change that must be locally or live tested

TheLordME (approved):
Its been TMed for a week now

PR #6048, opened 2023-10-02
Fixes Openable Snack Traits Passing to Fruit

Captain277:
## About The Pull Request

1. **Adjusts Openable Snack Traits.**

## Why It's Good For The Game

1. _Somehow, Palpatine came back. I mean, uh, somehow fruits and grown snacks were being affected by the openable trait. I'm not exactly sure how that happened, since I didn't subtype them at all. However upon review and discussion with Kev, we decided not to tie Opening into the open container flag for foods at all. So I've reworked the proc to just check for a sealed variable._

## Changelog

:cl:
fix: Fixes Openable Snack Traits Passing to Fruit
/:cl:

### DIFF
--- a/code/modules/food/food/snacks.dm
+++ b/code/modules/food/food/snacks.dm
@@ -16,6 +16,7 @@
 	var/nutriment_amt = 0
 	var/list/nutriment_desc = list("food" = 1)
 	var/datum/reagent/nutriment/coating/coating = null
+	var/sealed = 0
 	var/custom_open_sound
 	var/open_message = "You peel open the can! It looks ready to eat!"
 	var/opened_icon = 0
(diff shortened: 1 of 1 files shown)


code/modules/food/food/snacks.dm
```diff
@@ -16,6 +16,7 @@
 	var/nutriment_amt = 0
 	var/list/nutriment_desc = list("food" = 1)
 	var/datum/reagent/nutriment/coating/coating = null
+	var/sealed = 0
```
silicons, line 19: this should be FALSE; use defines to denote booleans. rest of the file should be TRUE instead of 1

Captain277: Done!

TheLordME (approved):

PR #6068, opened 2023-10-13
Woah! Nice Clock(work tools)!

TheObserver-sys:
## About The Pull Request
image

While Clockwork content is currently still InDev™, this doesn't mean we can't add a few crumbs from them early. This adds a new loot pool, new tools, and a new crafting recipe.

Clockwork tools: Equal in power (stronger in the welder's case) to the Alien Versions, with different use sounds, and a description for each.  In addition: The clockwork multitool has wirevision, equal to the alien multi, and thus, the code has been changed accordingly.

The Replica Clockwork Welder: Looks like the real clockwork welder, but has half the capacity, and .4 use speed, with no eye protection. It regenerates on its own, though, and requires an experimental welder to make. A nice minor upgrade, and not a complete hazard to use.
 

## Why It's Good For The Game
These things were bound to make it into the game one way or another, so why not get that small piece of the bandaid done and over with. We can corrupt them later when Clockwork Content shifts from InDev, to something more tangible. 

## Changelog

:cl:
add: Adds clockwork tools, replica clockwork welders, and the clockwork tool spawning pool. Currently, it's hooked up to the rouge miner generation so they can appear in game, at prob(1)
/:cl:


TheLordME (commented):
please make macros for the <span>s

TheObserver-sys:
> please make macros for the <span>s

They have been made -- however, I have no clue how you call them in the middle of a description, nor as an entire description for a tool.

lolman360:
> > please make macros for the s
> 
> They have been made -- however, I have no clue how you call them in the middle of a description, nor as an entire description for a tool.

`"string with normal text [SPAN_FUNNY("string with the funny text")]"` works:tm: (i tested it in discord with !dm and it works fine)

TheObserver-sys:
> > > please make macros for the s
> > 
> > 
> > They have been made -- however, I have no clue how you call them in the middle of a description, nor as an entire description for a tool.
> 
> `"string with normal text [SPAN_FUNNY("string with the funny text")]"` works™️ (i tested it in discord with !dm and it works fine)

Yeah, funny you should mention that.
image
image

TheLordME:
> Yeah, funny you should mention that. image image

You can use the update_desc() proc (it allows you to use Runtime assignment to desc), 
alternatively you can just leave the span part out and append it to an examine(which also would allow you to add a proximity requirement)
`/obj/item/multitool/clockwork/examine(mob/user, dist)
	. = ..()
	. += SPAN_NZCRENTR("funny span massage")`

silicons (approved):
you generally don't want a space between the = and class and ', because that space is hard to read + gets sent verbatim so that's 2 characters of wasted space, but, yea, good enough for corporate :D

PR #6072, opened 2023-10-13
Adjusts Silencer Code and Adds 9mm Compact Double Stack

Captain277:
## About The Pull Request

1. *Rearranges and Redistributes Suppressor Variables.*
2. *Tweaks Update Icon for Gyrojet Pistol.*
3. *Adds 9mm Double Stack magazine.*

## Why It's Good For The Game

1. _The suppressor system was a mess. So much of a mess that I didn't notice the Konigin couldn't be suppressed until today, and I added that gun like, forever ago. So I dug into it and the procs and checks were really clunky and conflicting with each other. I've fixed that._
2. _I planned on making the icon just run the same check on everything, but there's a complex system at play here that isn't worth rewriting just yet._
3. _It's pretty strong, so I'm not making it craftable or anything, but I would like if it existed._

## Changelog

:cl:
tweak: Rearranges and Redistributes Suppressor Variables
tweak: Tweaks Update Icon for Gyrojet Pistol
add: Adds 9mm Double Stack magazine
/:cl:


TheLordME:
Gun code :|

TheLordME (commented):
why are you renaming the unloaded state from "-empty" to "-e" that is a reduction in readability of the code

Captain277:
> why are you renaming the unloaded state from "-empty" to "-e" that is a reduction in readability of the code

Pistols used -e but rifles used -empty. When I changed the update icon proc to check silencers - where before every gun had to define it on their own - I was working with pistols so I used that standard. When the scope of the PR expanded, I just matched what I'd already written.

code/modules/projectiles/guns/ballistic.dm
```diff
@@ -44,6 +44,19 @@
...
+		if(silenced)
+			icon_state = "[silenced_icon]"
+		else
+			icon_state = "[initial(icon_state)]"
+	else if(magazine_type && !ammo_magazine)
+		if(silenced)
+			icon_state = "[silenced_icon]-empty"
+		else
+			icon_state = "[initial(icon_state)]-empty"
+
```
timothyteakettle, line 59: I feel like this entire conditional section could be written a bit cleaner in a form something like this:

```
if(magazine_type)
    var/silenced_state = silenced ? silenced_icon : initial(icon_state)
	var/magazine_state = ammo_magazine ? "-empty" : ""
	icon_state = "[silenced_state][magazine_state]"
```

code/modules/projectiles/guns/ballistic.dm
```diff
@@ -236,6 +249,30 @@
...
+			if(!user.attempt_insert_item_for_installation(A, src))
+				return CLICKCHAIN_DO_NOT_PROPAGATE
+			to_chat(user, "<span class='notice'>You screw [A] onto [src].</span>")
+			silenced = TRUE
+			w_class = ITEMSIZE_NORMAL
+			update_icon()
+			return CLICKCHAIN_DO_NOT_PROPAGATE
+		else if(istype(A, /obj/item/tool/wrench))
+			var/obj/item/silencer/S = new (get_turf(user))
+			if(silenced)
```
timothyteakettle, line 266: you're making a new silencer everytime someone uses a wrench on a gun that can be suppressed

code/modules/projectiles/guns/ballistic.dm
```diff
@@ -236,6 +249,30 @@
...
+				to_chat(user, "<span class='notice'>You unscrew [S]] from [src].</span>")
+				user.put_in_hands(S)
+				silenced = FALSE
+				w_class = ITEMSIZE_SMALL
+				update_icon()
+				return
+			else
+				return
```
timothyteakettle, line 274: there's no need to put an else return here

code/modules/projectiles/guns/ballistic.dm
```diff
@@ -236,6 +249,30 @@
...
+			return CLICKCHAIN_DO_NOT_PROPAGATE
+		else if(istype(A, /obj/item/tool/wrench))
+			var/obj/item/silencer/S = new (get_turf(user))
+			if(silenced)
+				to_chat(user, "<span class='notice'>You unscrew [S]] from [src].</span>")
+				user.put_in_hands(S)
+				silenced = FALSE
+				w_class = ITEMSIZE_SMALL
+				update_icon()
+				return
```
timothyteakettle, line 272: there's no need for a return here

silicons (approved):

PR #6115, opened 2023-10-31
Makes the Tcomms SMES start with input on

SingingSpock:
## About The Pull Request

Exactly what it says on the tin. Makes the tcomms SMES start with input turned on so that it can utilize secondary power sources without engineering being present. This will increase the time before tcomms runs out of power but decrease the time for the rest of the station (since it's pulling in a consistent 100kW)

## Why It's Good For The Game

Apparently folks on shifts without engineering were having power in tcomms run out before the rest of the station because the SMES wasn't charging and tcomms was segregated from the rest of the grid.

## Changelog

:cl:
tweak: tcomms SMESes now start with input toggled on
/:cl:

### DIFF
--- a/code/modules/power/smes/smes.dm
+++ b/code/modules/power/smes/smes.dm
@@ -560,6 +560,7 @@ GLOBAL_LIST_EMPTY(smeses)
 	name = "telecomms smes"
 	desc = "A high-capacity superconducting magnetic energy storage (SMES) unit. This is the one dedicated to telecommunications."
 	charge = KWH_TO_KWM(SMES_COIL_STORAGE_BASIC * 1 * 0.25)
+	input_attempt = 1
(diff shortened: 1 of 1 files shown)


ktoma36 (approved):

ktoma36 (approved):

ktoma36:
Good change, will prolly need to check maps to make sure they're not using var edited versions if this doesn't solve the issue across the board. :)

SingingSpock:
> Good change, will prolly need to check maps to make sure they're not using var edited versions if this doesn't solve the issue across the board. :)

I personally made this subtype and mapped both of them in, they shouldn't be varedited unless someone other than me touched them in the last couple weeks

Captain277 (approved):

PR #6133, opened 2023-11-10
[TESTMERGE ONLY] duct tapes a 515 bug that dcs people on going to lobby

silicons:
inspect the code and you'll see why this is testmerge only ☠️


silicons:
shoo, bot

PR #6145, opened 2023-11-14
runtime maploading 2/3: multiz update

silicons:
3/3 is world structs and the fun that comes with that

this pr has:

- transitions
- new multiz system
- lookup optimizations
- misc updates


TheLordME (approved):
Its a silicons PR, nobody else knows whats going on anyways

PR #6158, opened 2023-11-22
Atlas Tram Update (Attempt 2) [DNM]

CharlesWedge:
## Map Progression!
Nanotrasen is building a colony on Lythios and as a result people will no evac from a tram rather then with a shuttle. This adds a tram station on a rearranged surface 3 complete with tunnel, expands CC with its own tram station and finally adds a new roach dungeon to surface 3 to replace the void that was created by the half removal of the spider caves. 

## Why It's Good For The Game
The creation of the colony on Atlas will for variety in station objectives as Atlas will now have an offscreen colony they will be tasked with supporting. It will also lay the ability to export power to the colony which I plan to make a regular station objective of sorts with time.

## Changelog

:cl:
add: Tram Station, New Dungeon, New CC
tweak: Layout of Surface 3
fix: also ranged roaches now no longer shoot like machine guns
/:cl:


CharlesWedge:
Note This is passing checks but for some reason the escape shuttle/tram does not work. Do not Test Merge until we can fix this.

PR #6183, opened 2023-12-04
smacks the CI suite with the if: always()

silicons:
well
turns out having jobs stop because the previous one exits is ruining the point

this is honestly a snowflakey patch but we'll deal with it proper later.


Zandario:
I hate this.

silicons:
"tg did it"

PR #6193, opened 2023-12-08
[TESTMERGE ONLY] Revert "Makes radiation wave decay inverse square instead of exponential"

timothyteakettle:
FUCK


silicons (approved):
so yeah

this is theoretically good
and i actually want this

but

it's just too laggy
so
we're going to revert it

PR #6212, opened 2023-12-21
tgui embedding hooks

silicons:
tl;dr tgui modules got an upgrade to be slightly better usual-case

still horrible.


TheLordME (approved):
What ever this dark magic is

PR #6220, opened 2023-12-26
Adds a new sound for custom votes, to prevent people from thinking they are transfer votes

TheLordME:
## Changelog

:cl:
admin: custom votes now have their own sound rather than the same as transfer votes
/:cl:


timothyteakettle (approved):

Keekenox:
LETS GO, ADD IT ADD IT, EVEN IF IT IS TEMPORARY

PR #6244, opened 2024-01-02
Adds a confirmation prompt for hiding items in food

Putnam3145:
## About The Pull Request

As title. Right now, if you use a small-enough item on any slicable food, you hide the item in the food, gone until it's sliced or eaten. This feature basically only exists to be annoying and, I assume, for "slip the prisoner breakout tools" meme. The former's bad and the latter's very, very rare. This adds a confirmation to it.

## Why It's Good For The Game

Making a super annoying feature much, much less annoying

## Changelog

:cl:
qol: There is now a prompt before you try to hide something in food
/:cl:

### DIFF
--- a/code/modules/food/food/snacks.dm
+++ b/code/modules/food/food/snacks.dm
@@ -217,6 +217,9 @@
 		var/hide_item = !has_edge(W) || !can_slice_here
 
 		if (hide_item)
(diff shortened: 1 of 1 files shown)


BlueWildrose (approved):
Personally I would've just made item hiding a do_after so we can always resort to the classic "move away" nonsense instead but this also works.

TheLordME (approved):

PR #6256, opened 2024-01-07
Makes phoronlocks go to 20C instead of 22

Putnam3145:
## About The Pull Request

Title.

## Why It's Good For The Game

22C is too hot for comfort for some species and the phoronlock heaters are so incredibly powerful that they inevitably heat the entire station to their temperature, and it really doesn't take long to do so. This is above the comfort point for some species, so this is an annoyance.

Plus, it being 22C doesn't make it work faster or better or anything. It's pointless. It runs at max power until it reaches the target temperature anyway. Turning up the thermostat to make it heat faster doesn't work in real life or in-game, sorry.

## Changelog

:cl:
tweak: Phoronlock heaters go to 20 instead of 22
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock/phoronlocks.dm
+++ b/code/game/machinery/doors/airlock/phoronlocks.dm
@@ -75,7 +75,7 @@
 /obj/machinery/portable_atmospherics/powered/scrubber/huge/stationary/phoronlock		//Special scrubber with bonus inbuilt heater
 	active_power_usage = 2000
 	efficiency_multiplier = 4
-	var/target_temp = T20C + 2
+	var/target_temp = T20C
 	var/heating_power = 150000
 
 /obj/machinery/portable_atmospherics/powered/scrubber/huge/stationary/phoronlock/heater //Variant for use on rift


TheLordME:
When updating phoronlocks i ran into the issue of the locks being stuck at 19.984°C
I am have started work on reworking airlocks as a whole, and will add a proper delta check to the programs.
Until then, this PR would break all Phoron and Frostlocks on atlas/tether

Putnam3145:
> When updating phoronlocks i ran into the issue of the locks being stuck at 19.984°C I am have started work on reworking airlocks as a whole, and will add a proper delta check to the programs. Until then, this PR would break all Phoron and Frostlocks on atlas/tether

They're still broken, just broken at a different spot. Now they go up to 21.984°C, which heats up the whole station. Having them go to 19.984°C, which very slightly cools the whole station, is not "breaking them", it's making them slightly *less* broken.

BlueWildrose (approved):
since the airlocks are already broken, might as well see how this influences things. gonna TM.

TheLordME:
the difference is, the phoronlock cycles in or gets stuck trying to rewarm the chamber

PR #6258, opened 2024-01-07
superheated epic gas cave now has a warning sign and also lava instead of water

lolman360:
tin


timothyteakettle:
me when i put one foot in the cool hot spring
my entire body evaporating:

BlueWildrose (approved):
good game design :^)

PR #6260, opened 2024-01-08
TGUIifies Material Processor Console

BlueWildrose:
## About The Pull Request

Before:

After:

Not shown in the screenshot: Mousing over the unlabeled buttons will show tooltips.

Also, I've added something to the exosuit fabricator to make bananium and silencium show up correctly as icons.

## Why It's Good For The Game

The raw HTML was shit. I made it TGUI instead. Need I say more?

Do note that this is my first time making a TGUI interface for SS13, so anyone who has critique for it is welcome to give it.

## Changelog

:cl:
code: The Material Processor UI now uses TGUI.
/:cl:


yourdoom9898:
I'm not entirely a fan of the settings being icon based from a readability point, even if it does look nice. As a new player, the icons wouldn't be intuitive to what setting is what.

TheLordME (approved):

BlueWildrose:
> I'm not entirely a fan of the settings being icon based from a readability point, even if it does look nice. As a new player, the icons wouldn't be intuitive to what setting is what.

Mousing over them has tooltips. It'll become more apparent what they are when they're hovered over.

PR #6271, opened 2024-01-11
Reworks airlock programs

TheLordME:
## About The Pull Request
In the old version Airlock controllers used a badly implemented system for determining what to do.
This PR reworks that system to use a set of defined states, in which the airlock can be.
There is two kinds of states, stable and transforming. 
Stable states stay the same unless an order is given.
Transforming states are actively altering the contents of the airlock chamber, like pressurizing.
This system allows for for more reliable and predictable behaviour of airlocks.

## Why It's Good For The Game
A key system for Space stations should be reliable and behave predicatable. The system that is replaced in this PR was not predicatable.

## Changelog

:cl:
code: Airlocks now use properly implemented statemachines
/:cl:


TheLordME:
Potential todos:
- [ ] Add a proper Heater for use by Airlock controllers
- [ ] Replace portable stationary scrubbers in airlocks with floor mounted ones
- [ ] Add a UI tab that allows for in game configuring

silicons (approved):

PR #6290, opened 2024-01-22
Fixes airlock overlays not appearing

Mazianni:
## About The Pull Request

Technically a regression. Smart overlays do not play nice with doors for reasons unknown.

## Why It's Good For The Game

some overlays actually show up now.


TheLordME (dismissed):

Lohikar:
This is not a good idea, new usages of the old system should not be introduced.

Mazianni:
> This is not a good idea, new usages of the old system should not be introduced.

Addressed.

Captain277 (approved):

PR #6298, opened 2024-01-25
Pilot office and Planetary sensors

TheLordME:
## About The Pull Request
Adding a Pilot Office on Surface 3 of the NSB Atlas. is in compliance with the development roadmap for pilots.
Adding planetary sensors

## Why It's Good For The Game
The pilot office is in the mentality of the future development plans for the job, making them less dependant on exploration (since they have their own rooms now)
Without planetary sensors it you have to send a shuttle into orbit to see your local space, now you can do that from the comfort of your bridge/pilot's office.

## Changelog

:cl:
add: Added Pilot office to surface 3
add: Added planetary sensors to lythios43c
code: Moves Exploration dep. areas into their own file, instead of the huge messy file they were in before
/:cl:


Captain277:
I think the Pilot's office here needs an entire redraft. We were kicking the topic around in Discord earlier, and I played with the layout some and came up with this. It gives the Pilots a public-facing meeting area without needing to redistribute their equipment locker area. They can share Exploration's break room as well, so that zone is extraneous. Most importantly this retains the Maintenance walkway and keeps the substation accessible from multiple hallways.
image

TheLordME:
> removes a coin spawn, place one elsewhere in maint why the combination of suit storage units and suit cyclers? why do they require their own break room when exploration has one + they're a short walk from the bar? that area is better used for maint lacking decals up to par for its surrounding areas

The equipment is moved out of explo together with the office, since pilots are not part of explo, as you can see by the fact that they dont get explo IDs, report to the HoP rather than the RD or Pathfinder
The suit storages are there because they look better


> I think the Pilot's office here needs an entire redraft. We were kicking the topic around in Discord earlier, and I played with the layout some and came up with this. It gives the Pilots a public-facing meeting area without needing to redistribute their equipment locker area. They can share Exploration's break room as well, so that zone is extraneous. Most importantly this retains the Maintenance walkway and keeps the substation accessible from multiple hallways. image

While I would have prefered to put the office somewhere else, Surface 3 makes the most sense, since thats where the majority of shuttles are

TheLordME:
> good new location
> 
> see most recent diffs, needs a new coin spawn to replace the old on, exploration has lost its roundstart spawn landmarks and one charger
> 
> also what are the SMES units in the old pilot prep room for?

they seperate the two shuttles from the main grid to reduce take off lag (the entire main grid is rebuild, which means every machine hooked to it, either by bypass or directly is updated

silicons:
i hate that locate in world but it's fine for now

PR #6328, opened 2024-03-03
Fixes Debris Field POI Spawning

CharlesWedge:
## It Was a Three Line Fix
And it took me this long to look into it. Anyway The Debris Field should now probably spawn more POIs then the Clown Ops ship. 

## Changelog

:cl:
fix: Debris Field POIs spawn properly now.
/:cl:


ktoma36 (approved):
Good to go though we should keep an eye on it if we swap to other maps

PR #6335, opened 2024-03-11
maybe fixes travis

silicons:
please stop pickign euthenia at random.

### DIFF
--- a/code/controllers/subsystem/mapping/maps.dm
+++ b/code/controllers/subsystem/mapping/maps.dm
@@ -67,7 +67,7 @@
 
 /datum/controller/subsystem/mapping/proc/read_next_map()
 	var/datum/map/station/next_map
-	var/datum/map/station/default = keyed_maps[keyed_maps[1]]
+	var/datum/map/station/default = get_default_map()
 	if(isnull(default))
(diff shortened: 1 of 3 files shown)


code/controllers/subsystem/mapping/maps.dm
```diff
@@ -67,7 +67,7 @@
 
 /datum/controller/subsystem/mapping/proc/read_next_map()
```
LetterN, line 68: do a ci define check then return forced map

LetterN (changes requested):

code/modules/mapping/map.dm
```diff
@@ -158,6 +158,10 @@
 	/// force world to be bigger height
 	var/world_height
 
+	/// allow random picking if no map set
+	/// used to exclude indev maps
```
TheLordME, line 162: use means set to false???

TheLordME (approved):

code/modules/mapping/map.dm
silicons: basically, 'set to false to prevent travis'

this is so people don't forget to turn stuff on; random failures are bad but so is not having any map coverage

PR #6371, opened 2024-04-10
Z-Mimic Season Pass 1

Lohikar:
This is largely just synchronizing ZM with the latest upstream version, though I also renamed SSzmimic to make updates like this easier.

Changes:
- It is no longer possible to outrun Z-Mimic, except now without the cost of doubling ZM init time.
- `ZMM_LOOKBESIDE` is now supported to handle wide mobs as well as long mobs. Use `ZMM_WIDE_LOAD` for both wide and long, such as scaled up atoms.
- It is now possible for a turf to override its Z-Mimic appearance.
- `MZ_TERMINATOR` can now be specified to cause a turf to act as the bottom of a ZM stack. 
- `MZ_NO_SHADOW` can be specified to opt out of Z-M's darkening effect, though lighting copy and AO are still active.
- Some bugs in layering caused by atoms moving up or down z-levels have been fixed.
- Z-Mimic now has a basic performance profiler.
- ZM appearance fixup no longer breaks underlays.


code/__DEFINES/rendering/zmimic.dm
Zandario, line 1: Should probably rename this file as well

Lohikar:
This replaces #5587.

Zandario (approved):

silicons:
testmerging :>

silicons (approved):

PR #6421, opened 2024-04-30
New features, New wrappers for cell_slot system.

Sorrowfulwinds:
## About The Pull Request
Adds wrappers for cells to the cell_slot system for cleaner calling. Adds a function to remove a cell and hand it directly to a mob for saner handling outside the file. Adds a bunch of comments from what I've divined in the hopes of granting others the same wisdom.
Swaps the single inducer line that can use the new wrapper procs.

## Why It's Good For The Game
Will hopefully allow others to look upon silicons work and understand it easier. Allows me to write prettier and safer code for the vehicle refactor.

## Changelog
-Comments a bunch of the functions because my brain is smooth
-Adds wrappers for cell.dm functions to allow for cleaner code lines when dealing with this system
-Adds mob_remove_cell() for easier cell removal in vehicles update.
-"Updates" inducers to use the new wrappers (a single proc)

:cl: Shadowcat
code: Adds cell.dm function wrappers to cell_slot.dm
code: Adds a function to remove a cell and hand it directly to a move in cell_slot.dm
refactor: Comments a lot of the functions in cell_slot.dm
/:cl:


code/game/objects/systems/cell_slot.dm
```diff
@@ -48,6 +54,21 @@
 	cell = null
 	parent.object_cell_slot_removed(., src)
 
+/**
+ * removes cell from the system and drops it in users hands
+ */
+/datum/object_system/cell_slot/proc/mob_remove_cell(mob/user)
+	if(isnull(cell))
+		return
```
TheLordME, line 62:
```suggestion
		return FALSE
```

TheLordME: technically its already doing this, but the readability is better with this

code/game/objects/systems/cell_slot.dm
```diff
@@ -101,3 +134,83 @@
...
+/**
+ * cell function wrapper - returns the rating of the cell inside or null
+ */
+/datum/object_system/cell_slot/proc/get_rating()
+	return cell?.get_rating()
+
+/**
+ * cell function wrapper - returns the cell src or null
+ */
+/datum/object_system/cell_slot/proc/get_cell(inducer)
```
TheLordME, line 150:
```suggestion
/datum/object_system/cell_slot/proc/get_cell(var/inducer)
```
While not required, once again improves readability, and clarifies what type inducer is (I am assuming bool in this case)

code/game/objects/systems/cell_slot.dm
```diff
@@ -101,3 +134,83 @@
...
+ */
+/datum/object_system/cell_slot/proc/use(var/amount)
+	return cell?.use(amount)
+
+/**
+ * cell function wrapper - checks if the specified amount can be provided. If it can, it removes the amount from the cell and returns TRUE otherwise does nothing and returns FALSE
+ * returns null if no cell
+ */
+/datum/object_system/cell_slot/proc/checked_use(var/amount)
+	return cell?.checked_use(amount)
```
TheLordME, line 197: If you want to either return true or false and no null
```suggestion
	return cell?.checked_use(amount) ? TRUE : FALSE
```

TheLordME (changes requested):

code/game/objects/systems/cell_slot.dm
TheLordME: this one is optional and just FYI, the other two I think should be implemented

code/game/objects/systems/cell_slot.dm
silicons: we specifically do not use var/ anywhere in parameters because it's implicit and therefore unnecessary

code/game/objects/systems/cell_slot.dm
silicons: there's two ways to go here.

1. do TRUE / FALSE if it's success / failure to match cell.checked_use
2. add null to allow the caller to check if there's a cell with this proc too using something like `switch(cell_slot.checked_use())`

i'm kind of on the camp of 1 tbh; if something cares about if the cell is there they should be checking `isnull(cell_slot.cell) anyways`

silicons: same goes for use() and others; if you have a use case for checking for null this way do it but otherwise just || 0 in use() and similar to coalesce to 0 if it's null.

code/game/objects/systems/cell_slot.dm
```diff
@@ -101,3 +134,83 @@
...
+/datum/object_system/cell_slot/proc/get_cell(inducer)
+	return cell?.get_cell(inducer)
+
+/**
+ * cell function passthrough - consumes energy using *universal units*
+ * returns amount used or null
+ */
+/datum/object_system/cell_slot/proc/drain_energy(datum/actor, amount, flags)
+	return cell?.drain_energy(actor, amount, flags)
+
```
silicons, line 159: stuff like this should have the @params section of drain_energy

this way dmdoc still functions

code/game/objects/systems/cell_slot.dm
```diff
@@ -56,9 +77,21 @@
...
 	return parent.object_cell_slot_mutable(user, src)
 
+/**
+ * returns TRUE if the slot has a cell
+ */
+/datum/object_system/cell_slot/proc/has_cell()
+	if (!isnull(cell))
+		return TRUE
+	else
+		return FALSE
```
silicons, line 93: return !isnull(cell)

code/game/objects/systems/cell_slot.dm
```diff
@@ -48,6 +54,21 @@
...
+ * removes cell from the system and drops it in users hands
+ */
+/datum/object_system/cell_slot/proc/mob_remove_cell(mob/user)
+	if(isnull(cell))
+		return
+	. = cell
+	user.put_in_hands_or_drop(cell)
+	cell = null
+	parent.object_cell_slot_removed(., src)
+	return TRUE
```
silicons, line 67: this entire thing should be using remove_cell() instead of doing it a second time
duplicate code is less maintainable

also decide if it'll either
1. return TRUE / FALSE for success / failure
2. better, return cell if success, null if failure

and document as such please

silicons (changes requested):
just a few things

code/game/objects/systems/cell_slot.dm
Sorrowfulwinds: Adjusted all the returns for wrappers to return 0 or FALSE depending on context of the function. Moved away from any NULL returns.

code/game/objects/systems/cell_slot.dm
Sorrowfulwinds: Replaced with silicons code

silicons (approved):

PR #6422, opened 2024-04-30
Blowing your head no longer runtimes the server

Sorrowfulwinds:
## About The Pull Request
Exploding a person will delete their head before their hair causing a null access runtime.
Removes a spawn(1) that makes a check stale.
Maybe Testmerge this first. I don't think this should cause any problems but I know some icon ops need a delay to work?

## Why It's Good For The Game
Blowing heads without error is a critical citadel feature.

## Changelog
-Deletes apparently pointless and buggy spawn(1) delay

:cl: Shadowcat
fix: fixes a runtime in the cleanup code for exploding carbon mobs
/:cl:

### DIFF
--- a/code/modules/organs/external/subtypes/standard.dm
+++ b/code/modules/organs/external/subtypes/standard.dm
@@ -276,7 +276,7 @@
 	if(owner)
 		if(iscarbon(owner))
 			name = "[owner.real_name]'s head"
-			//spawn(1) //This appears to do nothing and causes the checks above it to become stale. If this needs to be added back in consider moving it above if(owner). ~Shadowcat
+			
 				owner.update_hair()
 	get_icon()
 	..()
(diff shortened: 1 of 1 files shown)


code/modules/organs/external/subtypes/standard.dm
```diff
@@ -276,7 +276,7 @@
 	if(owner)
 		if(iscarbon(owner))
 			name = "[owner.real_name]'s head"
-			spawn(1)
+			//spawn(1) //This appears to do nothing and causes the checks above it to become stale. If this needs to be added back in consider moving it above if(owner). ~Shadowcat
```
TheLordME, line 279:
```suggestion
			
```

TheLordME (changes requested):

code/modules/organs/external/subtypes/standard.dm
```diff
@@ -276,7 +276,7 @@
 	if(owner)
 		if(iscarbon(owner))
 			name = "[owner.real_name]'s head"
-			spawn(1)
+			
 				owner.update_hair()
```
TheLordME, line 280:
```suggestion
		owner.update_hair()
```

TheLordME (approved):

PR #6442, opened 2024-05-18
adds trees to atlas

bearrrrrrrr:
## About The Pull Request

removes lack of trees to atlas.

obviously can't take screenshots of the whole thing but i will throw some here

image

image

image

in the western planes, also fixes an invisible door leading down to a bunker area and fixes the weird white walls in the bomb site

image

## Why It's Good For The Game

greater visual fidelity is good, probably.

## Changelog

:cl:
add: trees to atlas
/:cl:


TheLordME:
you could also just add trees and bushes and the like to the random debris spawned around the map

bearrrrrrrr:
> you could also just add trees and bushes and the like to the random debris spawned around the map

I don't like this. I find that it's a lot better to do these things by hand. There are certain details that end up looking far better.

It works fine for procedural planets like the class D moon. I might do it for the west planes in the future if it doesn't convey the message well? Unsure.

PR #6451, opened 2024-05-23
moving MRE vendors-ATLAS version (WIP)

KellyTheFenneCat:
moving the MRE vendor to not be a middle finger to the chef (I was told I couldn't just delete it, otherwise this would just be a removal.)

## About The Pull Request

Removes the MRE vendor in the bar area, shoves it where 2 tables where on surface 1 in a side room so it's still available for those that use it.

## Why It's Good For The Game

I feel like just removing it would be better, but I was told to keep it. Anyhow for moving it out of the bar is done purely because it being right next to the kitchen is basically the biggest middle finger you can give to any chef players. They're already fighting the almighty power that is the fitmax vendor, adding another reason to not interact with them 2 feet away from them is just salt in the wound. If you're going to the bar/kitchen for food items, you are going to interact with service staff. If you don't wanna interact with service staff, or there are none, that's what the 50 thousand vendors everywhere else on the map are for.

## Changelog

:cl:
tweak: Removes MRE vendor in the bar on surface 3
tweak: replaces 2 tables in the side room on surface 1 at the end of the hall with the MRE vendor
/:cl:

### DIFF
--- a/maps/rift/levels/rift-04-surface1.dmm
+++ b/maps/rift/levels/rift-04-surface1.dmm
@@ -8439,7 +8439,6 @@
 "fBn" = (
 /obj/effect/floor_decal/borderfloor,
 /obj/effect/floor_decal/corner/lightgrey/border,
-/obj/structure/table/reinforced,
 /turf/simulated/floor/tiled/steel,
 /area/tether/station/public_meeting_room)
 "fCu" = (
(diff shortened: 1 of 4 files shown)


TheLordME:
I would change the changelog to a single message like
🆑
tweak: Moves MRE vendor in the bar to the public meeting room on the ground floor
/:cl:

TheLordME (dismissed):
You could have left the table in the corner and it would still have been accessible but its fine

silicons (approved):

PR #6455, opened 2024-05-24
adds a mask that warms breath as you take it

TheLordME:
## About The Pull Request
As it says in the title, the heating power of 1kW is picked arbitrary to allow the mask to actually help on atlas atmos, it might be adjusted later to use a more realistic system

## Why It's Good For The Game
We have gas masks that filter hostile gases, why not have one that warms your breath. Now that we spend most of our time on a snowball planet?

## Changelog

:cl:
add: Added new mask for warming breath
code: removes a redundant check in breath code
refactor: renamed the filter_gas() proc on masks to process_air()
/:cl:

### DIFF
--- a/code/modules/clothing/masks/breath_warmer.dm
+++ b/code/modules/clothing/masks/breath_warmer.dm
@@ -1,15 +1,12 @@
 /obj/item/clothing/mask/warmer
 	name = "warming mask"
-	desc = "A sterile mask designed to help prevent the spread of diseases."
+	desc = "A mask that warms your breath, making the cold more bearable."
 	icon = 'icons/obj/clothing/ties.dmi'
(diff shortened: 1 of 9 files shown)


code/modules/clothing/masks/breath_warmer.dm
```diff
@@ -0,0 +1,21 @@
+/obj/item/clothing/mask/warmer
+	name = "warming mask"
+	desc = "A sterile mask designed to help prevent the spread of diseases."
```
BlueWildrose, line 3: Shouldn't this description be different?

TheLordME: yes

silicons (approved):

PR #6459, opened 2024-05-26
refactors asset subsystem

silicons:
refactors the asset subsystem to make it much more managed

adds support for 'deferred' assets for tgui, which will be later used for things like tgui character setup and admin panels.
deferred means that it'll be 'sent in a timely manner after the ui opens', as opposed to 'sent before the ui opens'


code/modules/asset_cache/asset_pack.dm
```diff
@@ -0,0 +1,201 @@
...
+
+/datum/asset_pack/proc/get_url_mappings()
+	return loaded_urls
+
+/// Returns a cached tgui message of URL mappings
+/datum/asset_pack/proc/get_serialized_url_mappings()
+	if (isnull(cached_serialized_url_mappings) || cached_serialized_url_mappings_transport_type != SSassets.transport.type)
+		cached_serialized_url_mappings = TGUI_CREATE_MESSAGE("asset/mappings", get_url_mappings())
+		cached_serialized_url_mappings_transport_type = SSassets.transport.type
+
```
silicons, line 200: this basically tells tgui to call tgui.html's loadcss/loadjs methods

as far as i can tell it's just a Better Way To Dynamically Load Things 
at some point i'll probably be touching up on tgui and how its assets are done when i do the 516 tgui refresh

code/modules/tgui/tgui.dm
```diff
@@ -142,6 +149,9 @@
 	src_object.on_ui_open(user, src)
 	for(var/datum/module as anything in modules_registered)
 		module.on_ui_open(user, src, TRUE)
+	// now send deferred asets
+	for(var/datum/asset_pack/assetlike in assets_deferred)
+		window.send_asset(assetlike)
```
silicons, line 154: window.send_asset is blocking

this should probably be addtimer/spawn at some point but i'm lazy + for now it's fine
i also haven't analyzed the impact of making this non-blocking on the large scale.

BlueWildrose (approved):

code/modules/client/client_procs.dm
```diff
@@ -219,11 +225,15 @@
...
+	spawn(1)
+		statpanel_boot()
 	// Instantiate tgui panel
 	tgui_panel = new(src, "browseroutput")
 	// Instantiate cutscene system
-	addtimer(CALLBACK(src, PROC_REF(init_cutscene_system)), 0)
+	spawn(1)
+		init_cutscene_system()
+	// instantiate tooltips
+	tooltips = new(src)
```
silicons, line 236: tl;dr:

- timers don't fire fast enough because we don't have MC init stages yet
- spawn(0) is too fast
- so spawn(1) is kind of the best for these procs
- realistically all of these should be datums but i'll leavei t for the TGUI refresh PR.

code/modules/client/client_procs.dm
```diff
@@ -347,8 +357,8 @@
 	// 	if(config_legacy.aggressive_changelog)
 	// 		changelog_async()
 
-	// ensure asset cache is there
-	INVOKE_ASYNC(src, PROC_REF(warn_if_no_asset_cache_browser))
+	// run post-init 'lint'-like checks
+	on_new_hook_stability_checks()
```
silicons, line 361: this is literally just everything that goes "if things aren't loaded yet load them"

BlueWildrose (approved):

PR #6461, opened 2024-05-27
Adds dry monster tamer, some sodas to mining vendor

Putnam3145:
## About The Pull Request

1. Adds an alternative to monster tamer with the same effects, made with root beer instead of whiskey. This is almost exactly as easy to obtain; both are in dispensers, the real "hard part" is grinding up meat.
2. Adds sarsaparilla, root beer and hard root beer to the mining vendor. This is partially for flavor reasons, but mostly because whiskey is also in mining vendors, so miners can make their own monster tamer; they should be able to do it with the new version, too.
3. Corrected the 20% ABV monster tamer not to be called "vaguely alcoholic".

## Why It's Good For The Game

1. An alcohol intolerant xenochimera would be a "soft grief" of a sort, like taking radiation weakness on an engineer character or whatever. It's possible to play around, but mostly it just causes grief for _other people_, if it goes wrong.

2. Monster tamer's an _extremely_ goofy part of the setting. It'd be fine if any ol' drink worked, or if it had to be pure ethanol, but no, it has to be _specifically_ whiskey. This handwaves it by making it some complex long-chain carbohydrate thing.

## Changelog

:cl:
add: Dry monster tamer, made with root beer and animal protein
tweak: Mining vendor gets root beer/sarsaparilla/hard root beer
spellcheck: 20% alcohol-by-volume monster tamer is no longer described as "vaguely alcoholic"
/:cl:


TheLordME (commented):
can't you base one of the tamers on the other? to avoid doubleing the ingest proc?
or make the actual code into a seperate proc and just call it from either

BlueWildrose (approved):
yeah, looks good

PR #6489, opened 2024-06-02
Alraune internals fix and some buffs

BlueWildrose:
## About The Pull Request

- Alraune can use internals again. The reason why they couldn't was because their snowflake code was trying to check to see if the alraune was able to take from the environment first before making them depend on internals. I've decided to wipe that code - if the alraune want to use internals, let them just use internals. Who cares if it might be wasteful?
- Boosted alraune and diona physiology again. They're stronger now, to compensate for that slowdown.
- Alraune burn modifier changed from 1.5 to 1.1

I may also do something to increase their pain tolerance, or just negate their pain altogether, but I'll need to ponder on that.

## Why It's Good For The Game

Fix good.

Also, alraunes too slow. Let's not make alraunes slower.

## Changelog

:cl:
balance: Alraune physiology is boosted again to carry even more without being slowed down.
fix: Alraunes can use internals again.
/:cl:

### DIFF
--- a/code/modules/species/station/alraune.dm
+++ b/code/modules/species/station/alraune.dm
@@ -32,10 +32,10 @@
 	metabolic_rate = 0.75 // slow metabolism
 
 	brute_mod     = 1    //nothing special
-	burn_mod      = 1.5  //plants don't like fire
+	burn_mod      = 1.1  //plants don't like fire
 	radiation_mod = 0.7  //cit change: plants seem to be pretty resilient. shouldn't come up much.
 
-	item_slowdown_mod = 0.25 //while they start slow, they don't get much slower
+	item_slowdown_mod = 0.1 //while they start slow, they don't get much slower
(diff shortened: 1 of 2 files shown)


code/modules/species/station/alraune.dm
```diff
@@ -141,21 +141,8 @@
...
-		var/envpressure = H.loc.return_pressure()
-		if(envpressure >= hazard_low_pressure)
-			environmentalair = TRUE
-
-	if(fullysealed || !environmentalair)
-		breath = H.get_breath_from_internal()
+	//just fuck off with this snowflake bullshit about checking if they're sealed off and just test for internals. too complicated.
+	var/datum/gas_mixture/breath = H.get_breath_from_internal()
 
 	if(!breath) //No breath from internals so let's try to get air from our location
```
TheLordME, line 147: this check and the following code is also redundant, and already processed by the get_breath from internals

TheLordME (changes requested):

TheLordME (approved):

PR #6509, opened 2024-06-10
Adds Cargo-only permission to the entrance into Mining Refinery on Atlas

washikarasu:
## About The Pull Request
This adds the cargo permission to the maintenance airlock into the mining refinery on Riftmap

## Why It's Good For The Game
Because we don't want people going in and just swiping everything from the refinery, or stealing unclaimed points.

## Changelog
Can no longer enter refinery from the back as anything but cargo.

:cl:
fix: fixed maintenance airlock permissions for mining refinery
/:cl:

### DIFF
--- a/maps/rift/levels/rift-02-underground2.dmm
+++ b/maps/rift/levels/rift-02-underground2.dmm
@@ -6452,9 +6452,7 @@
 /obj/structure/cable/green{
 	icon_state = "4-8"
 	},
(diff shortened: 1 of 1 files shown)


TheLordME (changes requested):
even if access is set on the airlocks by default, please use an access helper
Type path is /obj/map_helper/access_helper/airlock/station/

TheLordME (approved):

PR #6521, opened 2024-06-15
Spatial Grids

silicons:
adds a fast spatial lookup system able to grab things within x tiles 

theoretically faster than range()

inspired but not copied from /tg/'s spatial grids

### what's not done yet

- ours doesn't have cells being datums, and therefore we can't have easy registration of enter/exit events
- therefore we'll need to add /datum/spatial_grid_listener or something that acts like a 'managed' callback/signal that redirects enter/exit events from a certain cell to something else.


code/__HELPERS/_logging.dm
```diff
@@ -224,9 +224,9 @@ GLOBAL_LIST_INIT(testing_global_profiler, list("_PROFILE_NAME" = "Global"))
 	WRITE_LOG(GLOB.world_game_log, "SECURITY: [text]")
 
 /proc/log_subsystem(subsystem, text)
-	#ifdef CITESTING
-	log_world("[subsystem]: [text]")
-	#endif
+// 	#ifdef CITESTING
```
TheLordME, line 227: either delete it or keep it in, comments arent for storing old code

TheLordME (dismissed):

TheLordME (approved):

PR #6536, opened 2024-06-21
turf reservations - borders

silicons:
adds the ability to specify a special border around turf reservations

by default,

1. if the border is specified but no initializer is specified, it's a selflooping border
2. if the border isn't specified, **you can fly out of reservations.**

yes, we have no safety guards anymore by default.

this is because turf reservations are a very low level concept and the user of the API is meant to deal with potential reservation escapes.

reservations are still aligned to chunks; though the actual playable area is no longer chunk-aligned.
this may change in the future. this is the case now because we don't want reservations to take up significantly more chunks than they otherwise would have.

### DIFF
--- a/code/datums/components/turfs/transition_border.dm
+++ b/code/datums/components/turfs/transition_border.dm
@@ -9,6 +9,8 @@
  *
  * This way, we have *near* perfect native-like simulation with moves without
  * having to do anything too special.
+ *
+ * todo: multi-tile object support
  */
 /datum/component/transition_border
 	var/atom/movable/mirage_border/holder1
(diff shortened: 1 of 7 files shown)


code/datums/components/turfs/transition_border.dm
```diff
@@ -9,6 +9,8 @@
  *
  * This way, we have *near* perfect native-like simulation with moves without
  * having to do anything too special.
+ *
+ * todo: multi-tile object support
```
TheLordME, line 13: Therapist: What do we do when we feel overwhelmed?

Me: // todo:

Therapist: no

TheLordME (approved):

code/datums/components/turfs/transition_border.dm
silicons: ok tbf this is like an actually real issue

(it only works one way or the other and we need it noted down!!)

PR #6550, opened 2024-06-29
overmaps 1 - generation & spatial

silicons:
here we go gamers


silicons:
athena wants to run an event

that's basically a lore blocker

we'll be ready after next friday

BlueWildrose (approved):

PR #6555, opened 2024-06-30
medibeam stream projectors

silicons:
the replacement for old medibeams is here

they're basically nanite stream projectors

shot nanites are 'clingy' and stay on target, allowing for soft-prebuffing.

old mediguns aren't replaced yet; this is just an adminspawn item for now, while it's iterated upon.

also includes a /datum/beam refactor. new beams can operate in stretch mode as well as segmented, and are more efficient + supports more fancy rendering.


TheLordME (approved):
its one hour past midnight

PR #6589, opened 2024-07-13
Fix hovering fire alarms, refactor DMM rotation code

nyeogmi:
## About The Pull Request

Previously:

- A lot of unnecessary code to thread and repeatedly add orientation-related variables to the preloader.
- dmm_parsed previously modified its own args, then used arglist(args) This is bad energy -- as far as I can tell, this was was relying on the fact that _load_impl didn't use the modified values.
- Fire alarms were hardcoded to deal with east/west incorrectly

Now:

- These variables now live in the dmm_context.
- _load_impl takes fewer unnecessary args, and all its args are explicit
- A few variables have been removed or had their scope limited.
- Fire alarms aren't hardcoded to do that anymore.

Bugs I didn't fix:

- When a map is rotated EAST or WEST, signs spawn in an overlapping fashion that's really ugly.

### Correctness

Fire alarms were spawning sideways when rotated EAST or WEST and probably shouldn't be. I've fixed this behavior now.

### Style

Threading a lot of variables in DM is pretty rough -- DM doesn't actually warn if you do this wrong. Silicons specifically flagged this as a rough spot for code quality, which is why I attacked it.

These variables never change during the load process, so threading them as arguments creates the false implication that they can be different per map object.

### Performance

My goal wasn't to make performance better, just to make the code more readable without making it worse. Brief profiling implies it might be faster, but the difference is marginal and possibly explained by bad test conditions:

- map loading took 24.0, 24.3, 24.6 seconds before the change
- map loading took 23.0, 23.6, 23.8 seconds after this change

All tests were done with rotated Triumph. (I haven't profiled non-rotated yet.)

Back of the envelope calculation implies the number of variable-on-object lookups in the hot loop phase of the map load should be pretty close to the original number -- maybe a little lower, since I've turned some into locals.

The one lookup that looks new (where we get the context from the preloader in preloader_load) was actually one that happened unconditionally at the end of the procedure in the previous code.

## Images

Sign bug: I didn't fix this or introduce this.

image

Fire alarm bug living its best life. I fixed this:

image

Fire alarm is Tethered. (Post-fix):

image

## Changelog

:cl: Pyrex
refactor: DMM rotation code is a little tidier
fix: fire alarms have a 0% chance of hovering
/:cl:

### DIFF
--- a/code/modules/mapping/dmm_suite/dmm_parsed.dm
+++ b/code/modules/mapping/dmm_suite/dmm_parsed.dm
@@ -279,11 +279,12 @@
 	loading = TRUE
 	Master.StartLoadingMap()
 
-	context.loaded_orientation = orientation
-	context.loaded_dmm = src
+	_populate_orientation(context, orientation)
+
+	ASSERT(!isnull(context))
 
 	global.dmm_preloader.loading_context = context
-	var/list/loaded_bounds = _load_impl(arglist(args))
+	var/list/loaded_bounds = _load_impl(x, y, z, x_lower, x_upper, y_lower, y_upper, z_lower, z_upper, no_changeturf, place_on_top, area_cache, context)
 	global.dmm_preloader.loading_orientation = null
 
 	context.loaded_bounds = loaded_bounds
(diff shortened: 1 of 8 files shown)


code/modules/mapping/dmm_suite/dmm_parsed.dm
```diff
@@ -292,24 +291,31 @@
...
-	var/yi = orientation_pattern.yi
-	var/turn_angle = round(SIMPLIFY_DEGREES(orientation_pattern.turn_angle), 90)
-	var/delta_swap = x - y
-
+	var/invert_x = context.loaded_orientation_invert_x
+	var/invert_y = context.loaded_orientation_invert_y
+	var/swap_xy = context.loaded_orientation_swap_xy
+	var/xi = context.loaded_orientation_xi
+	var/yi = context.loaded_orientation_yi
+	var delta_swap = x - y
```
GDNgit, line 318:
```suggestion
	var/delta_swap = x - y
```

code/modules/mapping/dmm_suite/dmm_preloader.dm
```diff
@@ -65,22 +56,26 @@ GLOBAL_REAL_VAR(dmm_preloader_target)
 
 	// handle post processing, so things like directions on subtypes don't break.
 	// only do everything if necessary.
-	if(preloader_local.turn_angle != 0 && what.preloading_dir(preloader_local))
+	var/datum/dmm_context/loading_context_local = preloader_local.loading_context
+	var/turn_angle = loading_context_local.loaded_orientation_turn_angle
+	if(turn_angle != 0 && what.preloading_dir(preloader_local))
```
GDNgit, line 61:
```suggestion
	if(turn_angle && what.preloading_dir(preloader_local))
```
Should just be able to check for truthyness here

GDNgit (commented):
I have now ruined my github recommended

code/modules/mapping/dmm_suite/dmm_parsed.dm
nyeogmi: OH NO! Apparently I have been writing JavaScript lately.

code/modules/mapping/dmm_suite/dmm_preloader.dm
nyeogmi: Yup, you're right!

silicons (approved):

PR #6601, opened 2024-07-20
MC: subsystem time dilation tracking, scheduling tweaks

silicons:
so

it turns out after kicking our server to 40 fps that an interesting thing happened

the MC is no longer overrunning all the time!
YAY!--

No. Things are not that simple.

Time dilation is a measure of how much we're overrunning the tick, and causing the MC, and **all** subsystems to lag.
This is most visible by players moving slower than they should.

That's cool and all, but, what we're experiencing is **subsystem** dilation.
Subsystems yield to not use too much CPU. This is good.

But we need to know exactly how much slower they're running than usual.
"tick used" is already a thing, but that doesn't tell me what I need to know. Using more tick than you should is not the same as _taking longer in absolute time to finish than you should._

The subsystem in question I'm debugging is projectiles.
Projectiles is non-ticker (thanks lohikar,,,,), 40 FPS.

It does not overrun its tick.
It is actually _very_ reasonable about its tick usage, and is always cooperative.

The problem is less cooperative subsystems (or non-subsystem tasks, or even ticker subsystems) are taking all the tick, and it's not getting enough allotment.

This results in things slowing down.
Thus, I need to see how bad it is.

# What the issue actually is

after further looking into it, it turns out the MC doesn't re-schedule subsystems every tick unless they're ticker, and even then it's not every tick.

This is bad because even if tick allotment is correct, things that should fire every tick, _don't_.

# What this actually does

- Add time dilation tracking to subsystems (aka how far we're under-running the real time we should be firing)
- tweaks the MC to support "fire every tick" non-ticker SS
- tweaks the MC a bit in general with var names and documents some of it better
- the MC now checks every subsystem that needs to fire every tick instead of waiting; we should probably have a better system for this but this is the only way to have 0-wait work for now.


code/controllers/subsystem/airflow.dm
```diff
@@ -12,7 +12,7 @@
 SUBSYSTEM_DEF(airflow)
 	name = "Airflow"
 	wait = 2
-	subsystem_flags = SS_NO_INIT
+	subsystem_flags = SS_NO_INIT | SS_BACKGROUND
```
Lohikar, line 15: This is probably not a good idea given that background can be totally starved of CPU time during things like explosions or ZAS events.

code/controllers/subsystem/asset_loading.dm
```diff
@@ -7,7 +7,7 @@ SUBSYSTEM_DEF(asset_loading)
 	name = "Asset Loading"
 	priority = FIRE_PRIORITY_ASSET_LOADING
 	// todo: hibernation
-	subsystem_flags = SS_NO_INIT
+	subsystem_flags = SS_NO_INIT | SS_BACKGROUND
```
Lohikar, line 10: Same applies here.

silicons: asset loading is a lazy thing / not actually urgent (i just named it asest loading but it's more like 'lazyload stuff people haven't asked for yet', if someone asks it's loaded instantly)

TheLordME (approved):

PR #6606, opened 2024-07-21
vehicle mechs - preliminary

silicons:
this basically duct tapes mechs to the vehicle system, while marginally improving some things.
we can start the mech rework after this.

# What this does

- no more MMI mechs; those were shitcode
- mechs are now /obj/vehicle/sealed/mecha
- some minor rebalancing to armor and speed


code/modules/maps/overmap/space/alienship.dm
```diff
@@ -116,62 +116,62 @@
 	. = ..()
 	dump_area = locate(/area/tether_away/alienship/equip_dump)
 
-/area/shuttle/excursion/away_alienship/shuttle_arrived()
```
silicons, line 119: all of this were actually removed in overmaps update years ago but is still sitting around.

ktoma36: I was curious about this, seen it a while ago. Makes sense it's just legacy stuff

ktoma36 (approved):
/vehicles already emulates door bumps so we don't need the special door bump handling that mech code specifically has. Shouldn't matter atm but will need cleared up later.

PR #6619, opened 2024-07-23
Revert some IC balance from Virgo

Dimasw99:
## About The Pull Request
**Solar cell power output has been changed from 1W back to 30W of its original design.**
It being reduced down to 1W made it practically useless for any real design. A lot of even basic components use roughly 30W per second.
**Change back sizes from tesla chargers to small and medium respectivly**
There never really was a need for it and tesla power is rare outside the station.
**Change the list deletor to 0.1 seconds**
A fairly lightweight circuit for what it does and it ensures it can be used together with at List running at the same speed.

### DIFF
--- a/code/modules/integrated_electronics/passive/power.dm
+++ b/code/modules/integrated_electronics/passive/power.dm
@@ -13,12 +13,12 @@
 /obj/item/integrated_circuit/passive/power/solar_cell
 	name = "tiny photovoltaic cell"
 	desc = "It's a very tiny solar cell, generally used in calculators."
-	extended_desc = "The cell generates 1W of energy per second in optimal lighting conditions.  Less light will result in less power being generated."
+	extended_desc = "The cell generates 30W of energy per second in optimal lighting conditions.  Less light will result in less power being generated."
 	icon_state = "solar_cell"
 	complexity = 8
 	origin_tech = list(TECH_POWER = 3, TECH_ENGINEERING = 3, TECH_DATA = 2)
 	spawn_flags = IC_SPAWN_DEFAULT|IC_SPAWN_RESEARCH
-	var/max_power = 1
+	var/max_power = 30
 
 /obj/item/integrated_circuit/passive/power/solar_cell/make_energy()
 	var/turf/T = get_turf(src)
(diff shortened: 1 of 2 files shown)


code/modules/integrated_electronics/passive/power.dm
```diff
@@ -13,12 +13,12 @@
 /obj/item/integrated_circuit/passive/power/solar_cell
 	name = "tiny photovoltaic cell"
 	desc = "It's a very tiny solar cell, generally used in calculators."
-	extended_desc = "The cell generates 1W of energy per second in optimal lighting conditions.  Less light will result in less power being generated."
+	extended_desc = "The cell generates 30W of energy per second in optimal lighting conditions.  Less light will result in less power being generated."
```
TheLordME, line 16:
```suggestion
	extended_desc = "The cell provides 30W in optimal lighting conditions. Less light will result in less power being generated."
```

code/modules/integrated_electronics/passive/power.dm
```diff
@@ -94,7 +94,7 @@
 /obj/item/integrated_circuit/passive/power/relay
 	name = "tesla power relay"
 	desc = "A seemingly enigmatic device which connects to nearby APCs wirelessly and draws power from them."
-	w_class = WEIGHT_CLASS_NORMAL
+	w_class = WEIGHT_CLASS_SMALL
 	extended_desc = "The siphon generates 250W of energy, so long as an APC is in the same room, with a cell that has energy.  It will always drain \
```
TheLordME, line 98:
```suggestion
	extended_desc = "The siphon draws 250W of electrical energy, as long as an APC is in the same room, with a cell that has energy. It will always drain \
```

code/modules/integrated_electronics/passive/power.dm
```diff
@@ -156,7 +156,7 @@
 /obj/item/integrated_circuit/passive/power/relay/large
 	name = "large tesla power relay"
 	desc = "A seemingly enigmatic device which connects to nearby APCs wirelessly and draws power from them, now in industiral size!"
-	w_class = WEIGHT_CLASS_BULKY
+	w_class = WEIGHT_CLASS_NORMAL
 	extended_desc = "The siphon generates 2 kW of energy, so long as an APC is in the same room, with a cell that has energy.  It will always drain \
```
TheLordME, line 160:
```suggestion
	extended_desc = "The siphon draws 2 kW of electrical energy, as long as an APC is in the same room, with a cell that has energy.  It will always drain \
```

TheLordME (dismissed):
the unit Watt is Joules per second, joules is the unit for energy, the right expression, when talking about power draw (or generation) is either watts, or joules per second
watts per second only make sense when you talk about the change of energy draw
(I know you didnt write it, but while we are here we can fix it B) )

deathride58 (approved):

PR #6634, opened 2024-07-29
Backspaced out old dna lock system from guns

silicons:
it's obsolete in the face of pins, which we now use as our auth mechanism for guns.
It's also shit code and badly balanced.

m8ko093os7a61


deathride58 (approved):
Will testmerge this to make sure nothing breaks

TheLordME (approved):

PR #6646, opened 2024-08-05
Scori-The Bronzening

CharlesWedge:
## From Bone to Bronze
Many Scorian items were borrowed from Main's Ashlander assets. The borrowed assets did not do a good job of showing the status of the Scorian civilization as a bronze age civilization as the majority of their tools ended up being bone, to say nothing of the borrowed tools that used assets from mobs that do not exist RP (Legion powered welder and cautery). As such I have replaced the bone tools with bronze ones, in addition, the recipes of the affected items have also been adjusted. Scori will need to acquire more copper to make their tools. Bones are still used for some tools and as components on other recipes.

Many tools have been altered for fluff purposes. Many Scorian Recipes now require tools of some kind (As we can no longer assume they can assemble complex items from raw materials with their bare hands). This may be a slight nerf for the Scori as the bone material may be overpowered. Please don't full merge until we can be sure the balance isn't overly altered by these changes.

## Changelog

:cl:
add: Bronze Tools and Weapons (May be slightly worse than bone due to material fuckery)
tweak: many many recipes
imageadd: icons for the new tools, inhands icons for the tools

/:cl:


code/modules/mining/mine_items.dm
```diff
@@ -307,11 +321,11 @@
 	throw_force = 7.0
 	w_class = WEIGHT_CLASS_SMALL
 
-/obj/item/shovel/spade/bone
```
TheLordME, line 310: Why are we getting rid of the bone spade completely?

code/datums/outfits/ghostrole.dm
TheLordME, line 1: Wouldnt it make sense to start with bone and work for bronze tools?

TheLordME (commented):
Did you test the "new" scori workflow with your changes?

code/modules/mining/mine_items.dm
CharlesWedge: Fixed this was removed by accident

code/datums/outfits/ghostrole.dm
CharlesWedge: I wanted to replace bone tools for a couple of reasons.
1. One as it turns out there is not that big a gap between bone and station tools to begin with so adding a 'bronze' tier would be a step too small and I wasn't going to make bronze just another steel or make bone just awful to use.
2. My intentions for Scori do not include doing tool upgrades every round. M
3.  I want the presence of bronze tools to communicate to even a new player the development level of the Scori civilization. They have left primitive bone behind for reliable copper alloys.
4.  I also wanted to get away from lavaland sprites and tools as much as possible to help distance the Scori from their roots as Drow Ashies (I scream beca. 
5. I may make the system more complex involved in the future.
These are my reasons for not doing the bone stuff turned to bronze stuff and just replacing the bone tools outright.

BlueWildrose (approved):

PR #6579, opened 2024-07-06
combat code overhaul + projectiles v8: impact handling

silicons:
refactors hit handling to be a lot more sane, basically.

this'll be a bit bigger than #6544 unfortunately.

- adds new shieldcall system
- completely overhauls projectile hit handling
- adds new armorcall system
- updates size labelling to work with large prs again
- misc fixes

closes #6679
closes #6667

PR #6667, opened 2024-08-17
Kills off Baymiss and RNG misses making you look like a doofus at random similar to how ANY other bay derivative has done.

IrkallaEpsilon:
A percent chance to miss shots and throws even if you click on the guy or lead your shots are just not good gameplay. 

I have missed at near point blank ranges already.

## About The Pull Request

Removes baymiss. This should have been done a long time ago to be honest. Shit "mechanic". I think removal was attempted in one of the gun refactors but theres still residue in human_defense etc. that still makes me miss with an AEG at point blank range while two handing.

I think I hit every single instance of baymiss. Evasion stat for mobs **should** (TM) still work. Removal is very similar to baycode baymiss removal. I think Chomp did a similar thing a year ago. 

		var/distance = get_dist(TT.initial_turf, loc)
		miss_chance = max(5 * (distance - 2), 0) 

Let me get this straight: Its 20% at 6 tiles distance. 15% at 5. So basically even if you as the player have god aim and god clicks you still are bound to miss relatively often no matter the gun you use, no matter the character. Amazing.

## Why It's Good For The Game

Actually having people being good at clicking matter. Actually having an actual ammo count instead of having 50 rounds P90 mag out of which 7.5 rounds are bound to miss.

## Changelog

:cl:
del: Baymiss wont be missed.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -372,10 +372,13 @@ meteor_act
 			zone = ran_zone(BP_TORSO,75)	//Hits a random part of the body, geared towards the chest
 
 		//check if we hit
+		/*
 		var/miss_chance = 15
 		var/distance = get_dist(TT.initial_turf, loc)
 		miss_chance = max(5 * (distance - 2), 0)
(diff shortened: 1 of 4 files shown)


silicons:
#6579

silicons:
testmerge it tho

PR #6713, opened 2024-09-01
removes simplemob shadekins

silicons:
they're a proper server species now

there's no need for simplemobs

unlike mercenaries and other human(oid) simplemobs i cannot fathom a use for this (other than bloating the code and getting in the way of refactors) because we're not exactly going to just spawn vore monsters to vore the exploration team on citadel rp now are we


silicons:
no vorexploration

TheLordME (approved):

PR #6720, opened 2024-09-01
refactors attack hand proc header

silicons:
needed for #5677 

this is to make that PR slightly easier to review.


code/game/click/other_mobs.dm
```diff
@@ -25,12 +25,16 @@
...
-/atom/proc/attack_hand(mob/user, list/params)
-	var/datum/event_args/actor/clickchain/e_args = new(user, target = src, intent = user.a_intent, params = params)
+//  todo: better desc
+//  todo: e_args is not specified all the time, yet.
+/atom/proc/attack_hand(mob/user, datum/event_args/actor/clickchain/e_args)
+	// todo: remove
+	if(isnull(e_args))
+		e_args = new(user)
+	// end
 	if(on_attack_hand(e_args))
```
silicons, line 38: you may have noticed e_args isn't provided in most places

this is because everyone is uh, wrong. like. you're supposed to call attack_hand from clickchain, but people just lazy-out and have it called from literally fucking everywhere

list/params doesn't work for that reason.

so we're continuing the pattern; we'll slowly bring things around.

TheLordME (dismissed):

BlueWildrose (approved):

PR #6725, opened 2024-09-02
/datum/emote - emote refactor

silicons:
datumizes and standardizes all emotes because that file is in the fucking way of my saycode rewrite


code/__HELPERS/text/format.dm
```diff
@@ -15,4 +15,4 @@
 		return
 	// todo: optimize this if needed, maybe dynamic regex?
 	for(var/key in keys)
-		. = replacetext_char(., "$[key]", keys[key])
+		. = replacetext_char(., "%%[key]%%", keys[key])
```
silicons: it was either this or `${THIS}`; kinda leaning to the latter but not hard to regex for it later

`$` is simply too common to use as a replacement.

TheLordME (dismissed):

TheLordME (approved):

PR #6741, opened 2024-09-10
Projectiles reorganization and cleanup, impact effects, subtle EMP tweaks

silicons:
No gameplay changes except for EMP rebalancing.

Emps are now deterministic
Ion rifles now always hit for strength 2 instead of a probability tree of strengths
EMP damage slightly lessened as this makes a lot of things guaranteed to hit with strength 2 instead of 3


timothyteakettle (dismissed):

code/modules/mob/living/silicon/pai/defense.dm
silicons, line 1: I decided to just kill the instant death because emps are gonna be more common soon anyways

BlueWildrose (approved):

PR #6776, opened 2024-09-27
Lamps separated into on and off, Library lamps start on and Mining Gateway Medikits aren't empty

washikarasu:
## About The Pull Request
Makes lamps actually start on and replaces the empty medikits in surt gateway on both atlas and triumph

## Why It's Good For The Game
The library on the Atlas has always been super dim, and yet the lamps were obviously coded to be on. Something wasn't working, I guess this fixes it, and separates them into an on and off definition. As for the surt gateway medikits, it makes no sense to place down empty ones, rather remove them entirely.

## Changelog
adds on state for both lamp types, replaces empty medikits with regular medikits on atlas and triumph

:cl:
add: Added /obj/item/flashlight/lamp/on
add: Added /obj/item/flashlight/lamp/green/on
tweak: tweaked /obj/item/flashlight initialize proc
/:cl:


washikarasu:
I'd also repair the one line thing on the R-UST but I feel like that'd break the scope

silicons (approved):

PR #6790, opened 2024-10-08
atmos touchups: thermodynamics enforcement week

silicons:
- comments old code
- Cleans some stuff up
- fixes a thermodynamics / math issue in heat exchange pipes
- Greatly simplifies heat exchange operations against turfs


code/game/turfs/turf.dm
```diff
@@ -15,6 +15,10 @@
 	 * - an atmosphere id (use defines please)
 	 */
 	var/initial_gas_mix = GAS_STRING_TURF_DEFAULT
+	/**
+	 * Act like a specific temperature for heat exchanger pipes.
+	 */
+	var/temperature_for_heat_exchangers
```
TheLordME, line 21: why are we moving this to /turf instead of /turf/simulated ?
The code that uses this var is only ever run on /turf/simulated

code/__DEFINES/atmospherics/constants.dm
```diff
@@ -1,15 +1,34 @@
...
+#define R_IDEAL_GAS_EQUATION 8.31
+
+//* Math Helpers - Pressures *//
+
+/// 1atm of earth-standard pressure, in kPa
+#define ONE_ATMOSPHERE             101.325
+
+//* Math Helpers - Temperatures *//
+
+// todo: T_0C?
```
TheLordME, line 21: todo, make a decision

TheLordME (changes requested):

code/__DEFINES/atmospherics/constants.dm
silicons: this PR is massive already, are you absolutely sure you want me to start regexing here too?

code/game/turfs/turf.dm
silicons: so we can access it via `turf` level
it's not really about when it's used, it's about the fact that this is an inherent override to atmos behavior
everything else gets air from turf via API like return_air
there's no reason to have this one snowflake thing require `/simulated`; it's just against the usual atmos API to do so
a machine should not have knowledge of how the simulation for its turf is done, it should just do its job with the turf API

at some point we'll have `air_status` on `/turf` instead of `/turf/simulated` and `/turf/unsimulated`

after we evaluate what other optimizations unsim'd turfs have that simulated turfs don't anyways

TheLordME (approved):

TheLordME (approved):

PR #6816, opened 2024-10-24
reagents: reactions

silicons:
reworks chemical reactions


code/modules/reagents/chemistry/reactions/paint.dm
TheLordME, line 1: shouldnt these use the color defines?

TheLordME (approved):

TheLordME (approved):

PR #6826, opened 2024-10-29
Arbitrary gun attachment system 

silicons:
does not currently support projectile modification, that's coming later

Includes adding support for attaching maglights to certain guns


TheLordME (approved):
mfw silicons becomes a guncoder

BlueWildrose (approved):

PR #6828, opened 2024-10-30
Updates tgui backend libraries (typescript, eslint, compiler tools, etc)

silicons:


tgui/packages/common/collections.ts
```diff
@@ -57,6 +57,8 @@ type MapFunction = {
  *
  * If collection is 'null' or 'undefined', it will be returned "as is"
  * without emitting any errors (which can be useful in some cases).
+ *
+ * todo: investigate if this is needed; javascript contains native Collection.map().
```
LetterN, line 61: js has no collection.map, do you mean Array.map()? this map iterates thru Object key/values

TheLordME (approved):

PR #6843, opened 2024-11-06
proper rustg defines

LetterN:
## About The Pull Request
make rustg swap to rust_g64[.dll] when OPENDREAM
updates some misc documentations on the rustg_ procs
compile options properly set defines & properly warns


LetterN:
reftracking broken not my problem (ci checks for reftrack since theres a test for that one too)

silicons:
looking into reftracking myself in a bit

silicons (approved):

PR #6844, opened 2024-11-06
update sql

LetterN:
## About The Pull Request
title, update sql to support more than 1 connections, add ezdb, remove `%PREFIX%` on non-prefixed one


BlueWildrose:
setting it to draft so I don't mistakenly testmerge it until @LetterN is confident they fixed the issues caused with this PR

SQL/database_schema.sql
```diff
@@ -1,8 +1,6 @@
 /**
  * make sure to bump schema version and mark changes in database_changelog.md!
  *
- * default prefix is rp_
```
silicons, line 4: the reason this is prefixed is not because it's meant to be used as such
rename the file to `database_schema-template.sql` or something
the reason this file is here is so generating the actual `.sql`'s is as easy as find-replacing `%_PREFIX_%` instead of having to edit two files simultaneously. this results in a lot less typo's.

silicons:
this'll sound quirky but EZDB should never be automated on the DD instance.
the most we should have is a python script to automatically set up a mariaDB instance.
I think that's good for contributors.
having the DD instance spin it up/down is too far, we **want** people touching DB to be exposed to the DB itself. DB isn't used on citrp for trivial purposes, we need anyone touching it to know what they're doing.

as a side note; we generally have a policy that server should not require DB to run and operate, for the most part.
the only planned major backends/mechanics that will require DB are

- persistence 

.. and that's about it.
as part of that, data-stores for

- pictures / photographs
- paper / faxes / books 
- newscasters
- map persistence
- character persistence
- prototype repositories (which store persistent reagents/materials/etc)

will require DB. they will return null instances if queried when DB is down; this is intentional and *fine*, as anything not persistent would be loading hardcoded datums in.

PR #6846, opened 2024-11-08
[s] update the thing

LetterN:


code/modules/client/onboarding/security_checks.dm
```diff
@@ -2,50 +2,61 @@ GLOBAL_LIST_INIT(blacklisted_builds, list(
...
+	var/breaking_build = CONFIG_GET(number/client_error_build)
+	var/warn_version = CONFIG_GET(number/client_warn_version)
+	var/warn_build = CONFIG_GET(number/client_warn_build)
+
+	if (byond_version < breaking_version || (byond_version == breaking_version && byond_build < breaking_build)) //Out of date client.
+		to_chat_immediate(src, SPAN_DANGER("<b>Your version of BYOND is too old:</b>"))
+		to_chat_immediate(src, CONFIG_GET(string/client_error_message))
+		to_chat_immediate(src, "Your version: [byond_version].[byond_build]")
+		to_chat_immediate(src, "Required version: [breaking_version].[breaking_build] or later")
+		to_chat_immediate(src, "Visit <a href=\"https://secure.byond.com/download\">BYOND's website</a> to get the latest version of BYOND.")
```
silicons, line 42: other than the spoofing, is there a reason this is not using security procs / browse()?

it's not a critical security issue for someone to have wallhacks for a second before they're booted and logged
our old chat looks like absolute ass and it's a big UI/UX issue to have people be DC'd and forced to eyestrain to see why
also, all of these should log somewhere

PR #6855, opened 2024-11-12
admin stuff update

LetterN:
title
- admin stuff use href token now
- update VV, now (properly) supports appearance_visual (you can see the appearance!)
- update ahelp, support for screaming at admins if there are none. webhook changes so its formatted
- send2irc -> send2tgs


code/__DEFINES/_protect.dm
```diff
@@ -1,6 +1,5 @@
-/**
- * Completely occludes a path from view variable interactions.
- */
+///Protects a datum from being VV'd or spawned through admin manipulation
```
silicons, line 1: will dmdoc work when it's placed above the ifndef instead of the define?

code/__DEFINES/_protect.dm
```diff
@@ -11,9 +10,19 @@
...
 ##Path/can_vv_mark(){\
 	return FALSE;\
 }
+#else
+#define VV_PROTECT(Path)
+#endif
+// we del instead of qdel because for security reasons we must ensure the datum does not exist if Read is called. qdel will not enforce this.
 
 /**
  * Makes a path read-only to view variables.
```
silicons, line 28: readonly probably needs /Read() override too

code/__DEFINES/_cooldowns.dm
```diff
@@ -62,6 +62,14 @@
 
 #define COOLDOWN_TIMELEFT(cd_source, cd_index) (max(0, cd_source.cd_##cd_index - world.time))
 
+// kevin...
+// the cd indexes arent supposed to start with cd_
+#define COOLDOWN_START_A(cd_source, cd_index, cd_time) (cd_source.cd_index = world.time + (cd_time))
```
silicons, line 67: this seems to do the exact same thing as `COOLDOWN_START` and `COOLDOWN_FINISHED`

also i started using `CD_INDEX_` for specifically timer indexes because i want to namespace defines where it makes sense to; there's just so many defines floating around the codebase that typing anything in uppercase gives you a ton of suggestions from different things

having grouping helps a lot (e.g. i don't have to guess if `ATOM_INITAILIZED` or `ATOM_FLAG_INITIALIZED` if we use a verbose version is an atom flag, whereas `NOCONDUCT` and `NOBLOODY` tells me absolutely nothing in the name. this is what leads to people putting clothing flags on item flags and vice versa, because we didn't use to have very obvious tells *in the name* that it's improper)

this does make devs have to type a bit more but i think it's worth it for long term making understanding easier

code/__DEFINES/is_helpers.dm
```diff
@@ -21,9 +21,12 @@
 
 #define ismutableappearance(D) (istype(D, /mutable_appearance))
 
+#define isweakref(D) (istype(D, /datum/weakref))
+
 #define isimage(D) (istype(D, /image))
 
-#define isweakref(D) (istype(D, /datum/weakref))
+GLOBAL_VAR_INIT(magic_appearance_detecting_image, new /image) // appearances are awful to detect safely, but this seems to be the best way ~ninjanomnom
```
silicons, line 28: this should be reconciled with `IS_APPEARANCE` macro, seen in typeids

alternatively, switch the typeid macros to `IS_TYPEID_*` instead of `IS_*`, since those are implementation defined and definitely not as safe as this (then again; this also seems to be implementation defined, lol)

code/__DEFINES/vv.dm
```diff
@@ -62,8 +66,7 @@
 #define VV_HK_TARGET "target"
 ///name or index of var for 1 variable targetting hrefs.
 #define VV_HK_VARNAME "targetvar"
-/// to view an appearance virtual object
-#define VV_HK_VIEW_APPEARANCE "vv_appearance"
+// VV_HK_VIEW_APPEARANCE is handled on vv natively now
```
silicons, line 69: just get rid of the comment, anyone touching vv here knows what they're doing / shouldn't need a comment to know why

code/_globals/lists/misc.dm
```diff
@@ -9,3 +9,19 @@ GLOBAL_LIST_EMPTY(wire_color_directory) // This is an associative list with the
...
+	var/obj/sacrifice = new
+	for(var/varname in sacrifice.vars)
+		var/value = sacrifice.vars[varname]
+		if(!islist(value))
+			if(!isdatum(value) && hascall(value, "Cut"))
+				output += varname
+			continue
+		if(isnull(locate(REF(value))))
+			output += varname
+	return output
```
silicons, line 27: i honestly really dislike doing this because we do have our own slightly more janky way of doing it, **and** because imo, doing this in the render loop for vv is better

code/modules/mob/mob-logout.dm
```diff
@@ -34,13 +34,8 @@
...
 	update_client_z(null)
 	log_access_out(src)
-	if(admin_datums[src.ckey])
-		if (SSticker && SSticker.current_state == GAME_STATE_PLAYING) //Only report this stuff if we are currently playing.
-			var/admins_number = GLOB.admins.len
 
-			message_admins("Admin logout: [key_name(src)]")
-			if(admins_number == 0) //Apparently the admin logging out is no longer an admin at this point, so we have to check this towards 0 and not towards 1. Awell.
-				send2irc("LOGOUT", "[key_name(src)] logged out - no more admins online.")
+	// admin logout shenanigans moved to client destroy!
```
silicons, line 38: dont need to comment it, anyone editing this file should know what they're doing enough to look

tgui/packages/tgui-panel/chat/constants.js
```diff
@@ -26,13 +26,13 @@ export const MESSAGE_TYPE_LOCALCHAT = 'localchat';
 export const MESSAGE_TYPE_RADIO = 'radio';
 export const MESSAGE_TYPE_INFO = 'info';
 export const MESSAGE_TYPE_WARNING = 'warning';
-export const MESSAGE_TYPE_HELPFUL = 'helpful';
```
silicons, line 29: is this verified to not be necessary?

code/__DEFINES/_protect.dm
LetterN: kinda usless since you can still tag (and potentially) json_encode to get its value

code/_globals/lists/misc.dm
LetterN: wdym render loop? it runs every vv UI call?

PR #6858, opened 2024-11-13
mc update

LetterN:
## About The Pull Request
it's that time of the year again

notable mc changes:
- stack_end_detector exists
- PreInit gets called during subsystem new() (still before Initialize)
- init stages
- tick skipping for (known) laggy events (ie: explotions causing sendmap spam)
- 

## Changelog

:cl:
del: old inactivity subsystem is moved to server maint
server: resume_after_initializations config
server: drift_dump_threshold and DRIFT_PROFILE_DELAY config for automatically profiling high tick drift situations
/:cl:


silicons:
we are not removing all of citrp's changes to the master controller because it is out of date. 
this PR tramples everything from organization to commentary added because of lessons learned from us tweaking subsystems.
out of date does not mean unusable. anything that is actually being used on a day to day basis should not be just replaced with another codebase's reference implementation like this without due thought and rationale.

PR #6865, opened 2024-11-15
mass turf repath

silicons:
enough is enough with these damn planet subtypes

also drops unused 140x140's


code/__DEFINES/mining/legacy.dm
```diff
@@ -1,19 +1,31 @@
...
+		/datum/ore/uranium::name = 4 + 2,
+		/datum/ore/hematite::name = 5,
+		/datum/ore/coal::name = 5,
+		/datum/ore/phoron::name = 8 + 1,
+		/datum/ore/silver::name = 8 + 1,
+		/datum/ore/gold::name = 8 + 1,
+		/datum/ore/diamond::name = 2 + 3,
+		/datum/ore/platinum::name = 3 + 3,
+		/datum/ore/verdantium::name = 1 + 3,
+		/datum/ore/lead::name = 4,
```
silicons, line 28: these are arguably a balance change but some tweaks had to be done to get rid of the random per-planet rarities that made no sense anyways

BlueWildrose (approved):

PR #6893, opened 2024-12-01
toml config, database storage for repositories / singletons

silicons:
see title
one step closer to persistence

implements repository persistent storage for designs


code/__DEFINES/controllers/dbcore.dm
```diff
@@ -17,6 +17,11 @@
  */
 #define DB_MINOR_VERSION 3
 
+//* Tables *//
+
+/// Prefixes are currently disabled.
```
silicons, line 22: tl;dr

1. citrp will not be doing cluster setups in the foreseeable future
2. even if it did, databases should be agnostic to which server is accessing it and be otherwise synchronous as we're kind of one-world
3. therefore there's no point in prefixing. it's bad practice to be sharing a database with other services anyways.

tools/ci/config.toml
Zandario, line 1: Empty?

Zandario (commented):
Seems about right. Though I'm still not much of a DB guy

tools/ci/config.toml
silicons: yep

TheLordME (approved):
yolo

PR #6897, opened 2024-12-04
Throws out color priority system

silicons:
- temporary is cheap enough to do with filters
- fixed and admin are the same really; admins shouldn't complain they can't edit color if they don't know how the thing they're editing works
- washable is cheap enough to do with filters


code/modules/vehicles/ridden/boat.dm
```diff
@@ -83,7 +83,7 @@
 	if(!material)
 		qdel(src)
 		return
-	add_atom_colour(material.icon_colour, FIXED_COLOUR_PRIORITY)
+	add_atom_color(material.icon_colour)
```
TheLordME, line 86: why does material.icon_colour keep the u?

TheLordME (changes requested):

TheLordME (approved):

PR #6909, opened 2024-12-08
adds a verb to invert yourself 180 degrees horizontally. also, refactors update transform.

silicons:
look, mom! components can hook transform updates now. surely this will be used responsibly and i won't have to revisit this in the future. 

when do we get structs so we can go full ECS pattern with appearance updates? having this fragmented across multiple procs is kinda annoying!


TheLordME (approved):
THat

PR #6916, opened 2024-12-12
Resurrect Cargonia

washikarasu:
## About The Pull Request
Cargo rebuilt from the ground up.

## Why It's Good For The Game
Endeavour's cargo is one of the things I originally wanted to avoid in making the map. This rectifies it, returning some dignity to the important department.

## Changelog
Tears down and completely replaces cargo. This also removes the **RIDICULOUS** stairs from science into the refinery. If you really need it with no cargo, break in like everyone else.

:cl:
del: stairs into refinery
del: floating buttons
tweak: tweaked cargo
/:cl:

PR #6917, opened 2024-12-13
Make Maints Great Again

GySgtMurphy:
## About The Pull Request

addressing maints flow issues, might be something of a big deal

## Changelog

:cl:
add: maints stuff
add: maints loot to decks 1 and 4 of Endeavour, more subdivisions, added some more access areas. 
/:cl:


silicons:
Since I can't leave comments on map PRs, I'll just make a single comment.

## Overall

**This PR should not have touched everything at once.**

How the mapmerge system works is it merges **touched tiles**.

There's no downside to doing it maint-by-maint. It causes more overall file conflicts, but makes it a lot easier to review.
Now, you have a massive PR that **will** conflict with a lot more map PRs, not on a file level, but on a content level. This is not good.

## Screening

deck 1: 

1. unused room?
2. unused room?

image

deck 2:

3. while more stairs is usually good (depends, sometimes; too many multiz cross-points can be a good _or_ a bad thing depending on how it's done), this adds a lot of empty space. having empty places with just stairs outside of maint is questionable (this counts as non-maint).

image

deck 3:

4. [nitpick] making the fitness room smaller gets a minor ❓ from me
5. there's an unreasonable amount of empty space created by this, imo; is there any way to address this?

image

deck 4:

6, 7: it's questionable at best to add this much empty space but i'll not fret about it because the shape constraint imposed by the old shape is arguably worse than too much empty space. it's recommended to do 'filler' rooms within reason when this happens, because too much unused space is not good.

image

silicons (changes requested):
see comment

silicons:
This PR will majorly conflict with #6916 on the north half of deck 3. That PR has review / merge priority because it was made first.

GySgtMurphy:
> This PR will majorly conflict with #6916 on the north half of deck 3. That PR has review / merge priority because it was finished first, **less atomizable**, and is potentially able to be addressed relatively quickly.

Yes, but that PR goes against the design intent of the ship, and is something I don't think should be included, due to increasing the size of cargo far too much. That PR also completely removes all maints access on that end of the ship, if not all the maints areas.

PR #6920, opened 2024-12-14
tgui update

LetterN:
- updates all backend (`/common`) stuff to use ts for types, does **not** change the logic of `map()()`, `filter()()`, `sortBy()()`
- update linters & rules up with tg, all `interface/` changes are related to this
- sync deps with tg (sonar version was incorrect it isnt even running correctly before)
- update (most) of components to be typed
- updated tguipanel to latest(ish)
- sass(scss) updated
- unused ui is moved **outside** of any `packages/*` so that it does not get compiled in
- **IE8-10 support is slowly being removed in this pr**, as tg is only supporting ie11


Zandario:
Oh boy

PR #6928, opened 2024-12-18
Endeavour deck 3 portside tweaking

GySgtMurphy:
## About The Pull Request

Adjusts the areas around Cargo on Deck 3 to enable more hidden movement through that section of the ship, and removes an unnecessary dorm around Cargo to add more maints area and gimmicks.

## Why It's Good For The Game

Re-enables hidden movement and semi-hidden movement from areas around the cargo area, and into the middle and starboard sides from the port side aft of Deck 3 to the middle. 

## Changelog

:cl:
tweak: moved areas around in cargo, made less cavernous entrance lobby in line with other departments.
fix: reconnected disposals loop to mail room, entire system bypassed the mail room
remove: gimmick cludge shuttlecraft bay, walled in for a maints room/loot.
tweak: moved mining prep area out of the corner it was in, and stretched it slightly to remove the sense of crampness. 
add: small maints 'tool room' behind mail room, with salvageable machines and frames for building new ones. 
add: small gimmick room next to cargo, forward of the lobby.
add: one small maints dorm, right next to the vault, an homage to a 'payday' heist that was interrupted before the ship was delivered. 
add: more maints spaces. 
remove: excessive tritium, specialist tox and oxy med kits from cargo prep area
tweak: reduced size of q-pad room, was too large for no particular reason
remove: three redundant lounge areas from cargo - 1 from in front of QM's office, 1 from break room, one from excess hallway in front of mail room.
tweak: shifted mail room forward to remove extra hallway from in front of cargo. 
/:cl:


silicons:
Removed areas marked. 

1. I honestly can't tell if we added spacepods or something or if those are the new fighter sprites because they look kinda bad compared to the old one. Shame the old ones were copyrighted. Could I know why they were removed?
2. Fine-ish.
3. So, I actually think removing this makes the hall into cargo too long instead of wide. This, while a bit bloaty, makes it feel a lot more open and less like a closed-off hallway you have to navigate.
4. Fine-ish; the break room does have the new amenities still.

image

As for the new ones, here's some food for thought.

5: Empty room. This can maybe go into the maint empty rooms, as readding empty rooms, much like carving maint out with cargo, isn't really great. 
6: Ditto
7. Two wide maint is pretty optimal nowadays, though back in the day(tm) we usually went for one. That said, I'm wondering if we can restore a tile to cargo, as it does feel a little squished.
8. Food for thought: The removed grass lounge whatever thingy and the pods could probably be put back if the cargo shuttle was moved to the right. This would also lessen the need _to_ have to trade space between cargo and maints.

image

### pods

Assuming that wasn't some weird science staircase, this was actually something I missed on the original PR. I'm not too sure adding this to cargo was the right move, but I'm also not entirely against it.

I do think that cargo should have a tight hanger with **two**, though. There's little need for more than that (even planned, long-term changes that could potentially make pods have standalone overmap flight would be tough-pressed to justify more than that).

Or maybe with one, and room for another. Either/or. 
Traditionally this is robotics content but I also have been kinda going at this with a 'no one should own content' attitude lately so arguments for both are appreciated.

### break room

The break room is a bit weirdly laid out. 4 chairs instead of a sofa, the presence of a very small room next to it with just vendors and no maint access but having a 2-wide door in the break room.

I'm wondering if we have some alternatives for it.

GySgtMurphy:
> 3\. So, I actually think removing this makes the hall into cargo too long instead of wide. This, while a bit bloaty, makes it feel a lot more open and less like a closed-off hallway you have to navigate.

See attached, got something I think is more amenable. 

image

GySgtMurphy:
> Assuming that wasn't some weird science staircase, this was actually something I missed on the original PR. I'm not too sure adding this to cargo was the right move, but I'm also not entirely against it.

wasn't supposed to be a direct access to cargo, more to the maints leading in cargo's direction for semi-secure transfer of ordered gear, and possible shenanigans by players

silicons:
> > 3. So, I actually think removing this makes the hall into cargo too long instead of wide. This, while a bit bloaty, makes it feel a lot more open and less like a closed-off hallway you have to navigate.
> 
> See attached, got something I think is more amenable.
> 
> image

looks good; would recommend having the plants in 2 2x1's, so you don't have a divided hallway that feels too linear

GySgtMurphy:
> > > 3. So, I actually think removing this makes the hall into cargo too long instead of wide. This, while a bit bloaty, makes it feel a lot more open and less like a closed-off hallway you have to navigate.
> > 
> > 
> > See attached, got something I think is more amenable.
> > image
> 
> looks good; would recommend having the plants in 2 2x1's, so you don't have a divided hallway that feels too linear

Done

BlueWildrose (approved):

PR #6938, opened 2024-12-24
add minHealth which controls when carbons die so we can set it in species

timothyteakettle:
## About The Pull Request
adds support for controlling the ~~speed~~ point at which ~~crabs~~ carbons die
required for holospheres because their hologram form has very low health which isn't possible if they die at -100

this is set in mob/living because that's where maxHealth is currently handled but right now only carbons care about minHealth because everything else ignores the old config option and dies at 0

changes are replicated to also store crit and softcrit thresholds

## Why It's Good For The Game
better control over the point at which things die

## Changelog

:cl:
add: add minHealth which controls when carbons die so we can set it in species
/:cl:

### DIFF
--- a/code/__DEFINES/mobs/life.dm
+++ b/code/__DEFINES/mobs/life.dm
@@ -55,3 +55,7 @@
 #define ORGAN_DECAY_PER_SECOND_BRAIN			(60 / (60 * 10))				// brain decays entirely over 10 minutes, or to lethal degrees in 5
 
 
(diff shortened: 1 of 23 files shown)


TheLordME (dismissed):
why can crithealth not be set via species? if I set my minimum health higher than the crit health I never go into crit, which might be intentional at times, but also unintention at other times

TheLordME (approved):

PR #6964, opened 2025-01-10
holospheres (TESTMERGE ONLY)

timothyteakettle:
## About The Pull Request
IN SO FAR;
- 20 max health species that displays as a hologram
- all clothing and items they wear display as holographic, the same as their body
- hair and body alpha values can be customized
- other slots can be equipped like backpacks/belts/etc however they appear holographic (for now, might change this!)
- your loadout/job clothing copies onto your body, priotizing job clothing first
- if you use the enable-hologram verb or die as a hologram you will be turned back into your sphere, dying means you are stunned for 10 seconds and then you turn back into a hologram
- dying loses you nutrition and you also passively regenerate hologram health over time using nutrition, this can be regained using a recharging station
- being a hologram they are unable to be healed by normal things, they don't have a mouth, can't get diseases, etc, this is counteracted by having incredibly low health and their 'death' being quite punishing (loss of nutrition that requires a recharger, hardstun that leaves their shell vulnerable)
- cable coil heals burns on the sphere, welding tool heals brute on the sphere
- you can pick the sphere up by clicking on it
- you can revive the sphere using a multitool, this takes 10 seconds and requires the sphere to be full hp
- eye colour on the sphere is the character's eye colour
- if hologram is gibbed/dusted then the shell is also deleted
- action buttons for the verbs
- new clothing flag NO_UNEQUIP stops you unequipping the item
- attempting to strip the holosphere shell will strip the hologram (you can't avoid getting your items taken off you)

PLANNED:
testmerge this for a round or two and just get a better feel for their balance

## Changelog

:cl:
add: holospheres
/:cl:


Dimasw99:
I fear 20HP is a bit too low. For refrence, A spider does 18 to 30 damage, a 9mm does 20 damage, a laser beam does 40 damage

PR #6531, opened 2024-06-19
factions cleanup - removes most factions from station

silicons:
as discussed in suggestions on discord (thread was left up for days)

tldr for those git trawling in a few months:

it's nonsensical to maintain a dozen factions rather than focus on the dynamics of NT letting on a select few
we want to shift towards a secure research ship, it doesn't make sense to have everyone clown car-ing in from every different faction without strong ties to NT or other IC reason for them to be there.

PR #6980, opened 2025-01-27
Show contractors Employing Company+Allow alt titles

MarsM0nd:
## About The Pull Request

This fixes that contractor IDs (The faction not being NT ones, not the Traveler title) do show which other company they are employed by. At the same time, it allows them to use alternative titles for the jobs.

## Why It's Good For The Game

Showing which company someone belongs to on the ID was intended, and did just break. Alternative titles were explicitly disabled, due to confusing titles coming with some factions, but as there no longer are any, we can allow the full range of known station alternatives. A Vey-Med nurse probably should stay a nurse, and not become a full doctor when working at an NT station, at least title wise, as example.

## Changelog

:cl:
tweak: Allows contractors to have alt-titles
fix: Contractor IDs show the employing company
/:cl:

### DIFF
--- a/code/game/objects/items/id_cards/contractor_ids.dm
+++ b/code/game/objects/items/id_cards/contractor_ids.dm
@@ -1,14 +1,17 @@
 /obj/item/card/id/contractor
 	var/employing_coperation = ""
-	var/extern_title = ""
+	//var/extern_title = ""
+	//There are no faction specific alt titles, so they don't have any external title, and are just using the stations
 	//var/expiry_date = ""
 	icon_state = "chit"
 
 /obj/item/card/id/contractor/dat()
-	. += "<b>Employing Company:</b> [employing_coperation]"
-	. += "<b>External Job Title:</b> [extern_title]"
+	var/dat = list()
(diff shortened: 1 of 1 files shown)


TheLordME:
Contractor dont get Alt-titles because certain lore contributors gave their company specific alt-titles nobody understands. That is why Alt titles disapear when you are a contractor

MarsM0nd:
#6531 did remove all of the alternative titles (which were part of necropolis and the naramid guilds), so currently, it's not required. I didn't fully remove it, just comment out the parts to indicate that it should stick to station default names. 
As alternatives, either a list needs to be made for the default titles/no-default titles, to assign the rank instead, or a full refactor so the alt_title prototype gets handed there, being able to check if the background is set. The background handling could also be removed in ``code\modules\jobs\alt_title.dm`` and the hooks for it from ``code\modules\jobs\job.dm``, to make it more likely such won't get added in the future.

TheLordME:
ah, carry on then

TheLordME (approved):

PR #6991, opened 2025-02-10
Using the is_avoiding_ground() check: Flying no longer has you touching the ground.

MarsM0nd:
## About The Pull Request

Moves a few checks if someone's flying/hovering to use the is_avoiding_ground check. It also removes the slowdown of a tile if you aren't actually stepping on it. This check does include more, like if you are only riding on something that's lifting you up enough, but unless it flies itself, it'll trigger what you may be avoiding, and in turn, will be affected by it (i.E falling down the chasm, there is no delay in my test, so no Yoshiing over a single tile.) 
Draft for now to see if there are more places to check for it, and as the balance change is big enough to warrant additional thoughts.

## Why It's Good For The Game

Flying people no longer leave footprints, nor touch the ground for water or lava. Logically, they aren't even slowed down by the tiles. This heavily buffs flying, at least the ignoring the slowdown part, so that may be removed for the sake of balance. The rest should be more flavorful, or just making sense.

## Changelog

:cl:
tweak: If you aren't touching the ground, don't interact with things there.
/:cl:

### DIFF
--- a/code/game/objects/effects/debris/cleanable/blood.dm
+++ b/code/game/objects/effects/debris/cleanable/blood.dm
@@ -85,7 +85,7 @@ var/global/list/image/splatter_cache=list()
 
 /obj/effect/debris/cleanable/blood/Crossed(mob/living/carbon/human/perp)
 	. = ..()
-	if(perp.is_incorporeal())
+	if(perp.is_incorporeal() || perp.is_avoiding_ground())
 		return
 	if (!istype(perp))
 		return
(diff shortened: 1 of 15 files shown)


MarsM0nd:
Given the slowdown can be flagged on turfs so it'll still apply, even if flying, the balance problem can be somewhat handled, and with the (codersprite) drips trail, I'm not sure if there are any logical inconsistencies left, so it seems ready to me.

TheLordME (dismissed):

PR #7024, opened 2025-03-03
Broadcasts(like crew arrivals) are restricted to their map/overmap entity

TheLordME:
## About The Pull Request
almost all global broadcasts, like crew arrivals, Supermatter delams and so on, can only be heard on their local Z-Level and its connected Levels. 

## Why It's Good For The Game
Why should the station hear a trader thawing, or a trader hear someone arrive on station?

## Changelog

:cl:
tweak: most broadcastings that used to be global, to broadcast on its respective map
/:cl:

### DIFF
--- a/code/modules/multiz/basic_legacy.dm
+++ b/code/modules/multiz/basic_legacy.dm
@@ -1,6 +1,9 @@
 
 GLOBAL_LIST_EMPTY(connected_z_cache)
 
+/proc/GetStationZlevels()
+	return (LEGACY_MAP_DATUM).station_levels
+
 // todo: legacy code
 /proc/GetConnectedZlevels(z)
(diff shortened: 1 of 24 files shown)


code/modules/multiz/basic_legacy.dm
```diff
@@ -2,18 +2,15 @@
 GLOBAL_LIST_EMPTY(connected_z_cache)
 
 // todo: legacy code
-/proc/GetConnectedZlevels(z)
-	return SSmapping.get_z_stack(z)
```
Putnam3145, line 6: This is not equivalent to what it was changed to, comms goes down regularly and RCON does not work between Z-levels anymore

silicons:
brokey

PR #7034, opened 2025-03-12
shapeshifter copy and reset verb

lolman360:
## About The Pull Request

gives a verb to proteans and xenochimera that let them mimic another person within range.

this copies:
gender and identity

skin colour, skin tone and eye colour
hair and gradients
ears, horns, tail and wings
markings

target's icon base (for xenochimera ONLY. this, again, cannot work with proteans due to how limbs work. protean limbs are always robotic)
robot limbs (for proteans ONLY. this, again, cannot work with xenochimera due to how limbs work)

gives a verb to proteans and xenochimera that let them reset appearance to the currently-selected character slot.

## Changelog

:cl:
add: mimic-appearance verb, which lets you mimic the appearance of someone in range. for xenochimera and proteans only, because they're the only real shapeshifting races we have)
/:cl:

### DIFF
--- a/code/modules/species/protean/protean_powers.dm
+++ b/code/modules/species/protean/protean_powers.dm
@@ -55,6 +55,7 @@
 
 	//Organ exists, let's reshape it
 	var/list/usable_manufacturers = list()
+	usable_manufacturers["Default - Protean"] = null
 	for(var/company in GLOB.chargen_robolimbs)
 		var/datum/robolimb/M = GLOB.chargen_robolimbs[company]
 		if(!(choice in M.parts))
@@ -108,6 +109,7 @@
 		return
 	if(swap_not_rebuild == "Reshape")
 		var/list/usable_manufacturers = list()
+		usable_manufacturers["Default - Protean"] = null
 		for(var/company in GLOB.chargen_robolimbs)
 			var/datum/robolimb/M = GLOB.chargen_robolimbs[company]
 			if(!(BP_TORSO in M.parts))
@@ -294,6 +296,7 @@
 	if(new_species)
 		impersonate_bodytype_legacy = new_species.get_bodytype_legacy()
 		impersonate_bodytype = new_species.default_bodytype
+		impersonate_species_for_iconbase = new_species
 		regenerate_icons() //Expensive, but we need to recrunch all the icons we're wearing
 
 ////
@@ -357,7 +360,7 @@
 
 	for(var/mob/living/carbon/human/M in oview(7))
 		valid_moblist |= M
-	
+
 	var/mob/living/carbon/human/target = input(src,"Who do you wish to target?","Mimic Target") as null|anything in valid_moblist
 
 	if(!istype(target))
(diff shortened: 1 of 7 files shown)


MarsM0nd:
May be a bit out of scope, but it probably should allow you to also load a character slot, both to be able to reset to your old look, and to be able to load an alternative one. Currently, I think only xenochimeras could do that, and they would need to use their full regeneration, being dead for a while and starting feral.
Another species (and even more scope creep) would be the holospheres, they could fully copy one, including their look as a holograph.

BlueWildrose (commented):
You might as well make this applicable to all shapeshifters and not just proteans and xenochimera. That means prometheans and holospheres in particular. There's no reason why it shouldn't be that way as far as I know.

Otherwise, in the interest of not prefbreaking, there's a toggle currently present in Species Customization for characters called "Prevent Body Impersonation". Currently, this is meant to apply to resleeves, but I imagine players have other thoughts regarding what this checkmark means. You should see if you can make it prevent people from impersonating those with the resleeve_lock setting active.

lolman360:
> You might as well make this applicable to all shapeshifters and not just proteans and xenochimera. That means prometheans and holospheres in particular. There's no reason why it shouldn't be that way as far as I know.

holospheres are (as far as I'm aware) still unfinished/WIP, which is why I haven't rolled it out to them at the moment. prometheans i intentionally left out, but they can easily be added.

> Otherwise, in the interest of not prefbreaking, there's a toggle currently present in Species Customization for characters called "Prevent Body Impersonation". Currently, this is meant to apply to resleeves, but I imagine players have other thoughts regarding what this checkmark means. You should see if you can make it prevent people from impersonating those with the resleeve_lock setting active.

i'm aware. the main problem with resleeve impersonation is that it copies **the character's name** as well, making it almost indistinguishable who is actually playing the character. this specifically does not, so i think it's fine not touching the pref?

lolman360:
prefs lock implemented

BlueWildrose (dismissed):
Holospheres can probably come after that species is in a stable state, but thanks for making the changes.

Athena148:
> > You might as well make this applicable to all shapeshifters and not just proteans and xenochimera. That means prometheans and holospheres in particular. There's no reason why it shouldn't be that way as far as I know.
> 
> holospheres are (as far as I'm aware) still unfinished/WIP, which is why I haven't rolled it out to them at the moment. prometheans i intentionally left out, but they can easily be added.
> 
> > Otherwise, in the interest of not prefbreaking, there's a toggle currently present in Species Customization for characters called "Prevent Body Impersonation". Currently, this is meant to apply to resleeves, but I imagine players have other thoughts regarding what this checkmark means. You should see if you can make it prevent people from impersonating those with the resleeve_lock setting active.
> 
> i'm aware. the main problem with resleeve impersonation is that it copies **the character's name** as well, making it almost indistinguishable who is actually playing the character. this specifically does not, so i think it's fine not touching the pref?

I'd be careful giving this to any non WL species for obvious reasons. There's also some lore problems with this as Xenochimera are incapable of completely copying another individuals appearance.

lolman360:
> I'd be careful giving this to any non WL species for obvious reasons. There's also some lore problems with this as Xenochimera are incapable of completely copying another individuals appearance.

you're right, it's literally mentioned explicitly on their lore page that they can't copy other individuals, so I'm removing the functionality from xenochim.  i'm also against it  prometheans/hspheres/other non-WL races getting it, but ultimately is up to maintainers.

code/modules/species/protean/protean_powers.dm
```diff
@@ -343,6 +346,107 @@
...
+	for(var/mob/living/carbon/human/M in oview(7))
+		valid_moblist |= M
+
+	var/mob/living/carbon/human/target = input(src,"Who do you wish to target?","Mimic Target") as null|anything in valid_moblist
+
+	if(!istype(target))
+		return FALSE
+
+	if(get_dist(src,target) > 7)
+		to_chat(src,"<span class='warning'>There's nobody nearby to use this on.</span>")
```
TheLordME, line 370: shouldnt this be like: "[target] is to far away to mimic them"?

code/modules/species/species_shapeshift.dm
```diff
@@ -554,3 +554,263 @@ var/list/wrapped_species_by_ref = list()
...
+	var/mob/living/carbon/human/target = input(src,"Who do you wish to target?","Mimic Target") as null|anything in valid_moblist
+
+	if(!istype(target))
+		return FALSE
+
+	if(target.client.prefs.resleeve_lock)
+		to_chat(src,"<span class='warning'>This person has enabled Prevent Body Impersonation, you cannot copy them!</span>")
+		return FALSE
+
+	if(get_dist(src,target) > 3)
```
TheLordME, line 582: you build the list from oview(7) but check for dist > 3?
and you should propably change the failure text

TheLordME (changes requested):

TheLordME (approved):

PR #7061, opened 2025-03-27
random backend updates

LetterN:
## About The Pull Request
title

## Changelog

:cl:
fix: world_topic status host is correct now
code: updated progressbar
code: updated stack trace
code: updated callback
code: updated weakref
code: updated topic request, now supports returning json format
code: datum browser doesnt hold atom owner directly
code: updated stat tracking (debugging, not player stat)
admin: updated del-all verbs
admin: allow f12 browser inspect (516 only)
/:cl:


silicons (approved):

code/__DEFINES/_lists.dm
```diff
@@ -25,13 +25,28 @@
...
-/// Null-safe L.Copy()
-#define LAZYCOPY(L) (L? L.Copy() : null)
-/// Reads L or an empty list if L is not a list.  Note: Does NOT assign, L may be an expression.
+///Returns the list if it's actually a valid list, otherwise will initialize it
 #define SANITIZE_LIST(L) ( islist(L) ? L : list() )
 #define SANITIZE_TO_LIST(L) ( islist(L) ? L : list(L) )
+/// Performs an insertion on the given lazy list with the given key and value. If the value already exists, a new one will not be made.
+#define LAZYORASSOCLIST(lazy_list, key, value) \
+	LAZYINITLIST(lazy_list); \
+	LAZYINITLIST(lazy_list[key]); \
```
silicons, line 47: are the LAZYADD/REMOVE's for numbers / strings only? I feel like there's going to be confusion

defines are nice but what's wrong with
```
LAZYINITLIST(xyz)
xyz[thing] += val
```

i don't really like defines being too close to each other if they mean very different things semantically (LAZYADD vs LAZYADDASSOC)

Also please don't call that last one `LAZYOR` as it's operating on a list and not a binary value

code/__DEFINES/stack_trace.dm
```diff
@@ -0,0 +1,4 @@
+/// gives us the stack trace from CRASH() without ending the current proc.
+#define stack_trace(message) _stack_trace(message, __FILE__, __LINE__)
+
+#define WORKAROUND_IDENTIFIER "%//%"
```
silicons, line 4: can you comment what this is for?
we have `SOME_MAGIC_VALUE` 

also personally we should consider

`/magic_marker/xyz` or something at this point because we have a _lot_ of unique value-types that are just .. strings, unless we need this to be a string for some reason?

silicons: nevermind this **is** a string

what?
wack.

Put a comment saying what this is for please

silicons (approved):
whatever ill edit later

PR #7067, opened 2025-04-01
Allows leash laws to be enforced

MarsM0nd:
## About The Pull Request

Despite the availability of collars, and the presence of dogs, there are no leashes. This means you can't bring them to the indoor park, while keeping them leashed to not bother other guests of the park. Nor can you go outside, without the dog running of to chase the wildlife. This fixes that. 

## Why It's Good For The Game

April fools out of the way, it's mostly a port from [Virgo](https://github.com/VOREStation/VOREStation/pull/16813), with a few tweaks to make it work here, and some others to allow more flexibility. Variable length to allow some VV shenanigans, tested with the limited possibilities for the cable ones, and allowing simple dogs to be leashed (which could be expanded to more passive simple animals).

## Changelog

:cl:
add: Leashes
/:cl:


TheLordME (changes requested):
is this a genuine PR, or just an april fools one?
If the former, please address my comment

MarsM0nd:
> is this a genuine PR, or just an april fools one? If the former, please address my comment

April fools being the reason to make it, but it can be genuinely merged, in a way a port from a prior upstream, from how it was? In any case, check moved.

TheLordME (approved):

PR #7091, opened 2025-04-15
Less invasive chimera regenerate

MarsM0nd:
## About The Pull Request

Makes the chimera regenerate only replace missing organ/limbs, and fully heal them all, rather than deleting and respawning everything. This makes them not lose implants from it. It's done by making restore_all_organs actually restore ALL organs, and not just bodyparts. It's only used for aheals otherwise, and that already replaces every organ before, so there aren't any changes. The other part is copying the create_organs part where they are actually made, after checking the chimera would be missing a limb they should have. Should they somehow have another limb grafted to them, they'll heal it, but not regenerate if it gets amputated again. They will still heal robotic limbs, but that's not avoidable without more checks.

## Why It's Good For The Game

An ability that needs 10 seconds of their concentration currently removes any implant, which includes tracking ones. Lore wise, they would be required to have one, and it probably wouldn't make sense if it's that easily (and probably more likely accidentally) removed like that. There is also no indication this ability would do such, neither before or after, like a nif being thrown to the ground.

## Changelog

:cl:
tweak: Xenochimera regenerate is less destructive, only regenerates what's needed, and heals the rest.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human-damage-legacy.dm
+++ b/code/modules/mob/living/carbon/human/human-damage-legacy.dm
@@ -343,7 +343,7 @@ This function restores the subjects blood to max.
 This function restores all organs.
 */
 /mob/living/carbon/human/restore_all_organs(var/ignore_prosthetic_prefs)
-	for(var/obj/item/organ/external/current_organ in organs)
+	for(var/obj/item/organ/current_organ in organs)
 		current_organ.rejuvenate_legacy(ignore_prosthetic_prefs)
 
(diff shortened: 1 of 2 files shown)


MarsM0nd:
Found a problem: If a limb gets detached leaving a stump, that one doesn't get replaced.

MarsM0nd:
Handled by letting the removed() call handle doing so, and then just regenerating the part, the same way it turned into a stump in the first place, code wise.

TheLordME (approved):

PR #7093, opened 2025-04-15
Allows implant augments to be used

MarsM0nd:
## About The Pull Request

They are internal organs, not external organs, but the augment menu was searching there. 

## Why It's Good For The Game

Allows something that is there to be actually used, may give robotics something more to do.

## Changelog

:cl:
fix: augment implants can actually be activated
/:cl:

### DIFF
--- a/code/game/objects/items/melee/types/transforming/energy/ionic_rapier.dm
+++ b/code/game/objects/items/melee/types/transforming/energy/ionic_rapier.dm
@@ -14,6 +14,7 @@
 	lrange = 2
 	lpower = 2
 	lcolor = "#0000FF"
+	can_combine = FALSE
 
 	passive_parry = /datum/passive_parry{
 		parry_chance_default = 60;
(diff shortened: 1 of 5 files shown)


MarsM0nd:
A quick change for the medical/multiselect thing too, better ability to switch around (no need to select the tool again to store it, either drop it or automatically retract it choosing the option)

MarsM0nd:
Things will always get put into the left hand thanks to ``code\modules\mob\inventory\inventory-hands-put.dm``, put in hands for left and right is the same, for some reason, a bit outside of a quick fix

MarsM0nd:
Changed the equip in hand, given the step 2, it should have only tried to equip to the first hand slot, which becomes the left, so it shouldn't break anything that tried to put things in the left, and maybe have that then overflow to the right with that proc.

TheLordME (approved):

TheLordME (approved):

PR #2917, opened 2021-03-29
migrates to github actions + sync tgui to main's, removes a bunch of dead maps and code and reworks how submap loading is done

silicons:

PR #7107, opened 2025-04-21
Repairs the gateway away mission selection

Sorrowfulwinds:
## About The Pull Request
Turns out the Gateway is supposed to pick a random away mission at the start of every round. You know that stargate rip off in the back of explo?
Looks like it's been loosely broken since #2917 4 years ago. Though I don't remember anybody ever using the gateway since before that.

This PR fixes the system and simplifies it quite a bit. 
code/modules/awaymissions/zlevels.dm is now gateway_maps.dm

The method for loading and picking a map has been reduced to a hardcoded list in gateway_maps.dm instead of using a hardcoded txt file loaded using a bunch of nonsensical string processing.

There are 4 functional maps I've fixed and added back in:
Carpfarm
Listening  post
Snowfield
Snow outpost

Zoo would be really nice to have but it has a number of issues that are outside the scope of this PR so I've set it as a TODO.
Similarly I'm sure a bunch of the archived missions might be quite fun.

For those unfamiliar with the gameplay loop of the Gateway (considering its never worked):
1. The gateway has a ~30 minute timer roundstart where it is performing "warpspace triangulation". (Click on it it will tell you)
2. After that timer is over you can activate the Gateway and walk into it.
3. This will teleport you to the Gateway mission level randomly.
4. Find the Gateway in the mission level and smack it with a multitool to calibrate it and return.
5. Once you calibrate the return gateway you can teleport between the two freely.
!!All of these maps are explo HARDCORE mode and will absolutely kill you. This ain't miaphus!!

Devdocs:
If you want to make a gateway map it's extremely simple!:
1. Build a gateway, It's like 8 separate pieces just copy one from a station map.
2. Set the gateway console center bit to a: /obj/machinery/gateway/centeraway.
3. Sprinkle your map with /obj/landmark/gateway_scatter. 
- Players will be randomly teleported to these while the gateway is uncalibrated. Consider them your starting spawns.
5. Add a /datum/map_level/gateway/[mapname] and a /datum/map/gateway/[mapname] to your maps .dm file
- See maps/away_missions/carpfarm_140/carpfarm.dm for example.
6. Add the /datum/map/gateway/[mapname] path to the potentialGatewayLevels list in code/modules/awaymissions/gateway_maps.dm.

## Why It's Good For The Game
I don't particularly want the Gateway to even exist from a lore and game design standpoint.
But content today is better than content tomorrow.

## Changelog
-Rebuild gateway away mission maps
-Rebuild gateway away mission selection at roundstart
-Touch up away_mission folders and names

:cl: Shadowcat
refactor: refactored gateway mission code
fix: Fixed gateway random mission picker
fix: Fixed a few gateway missions to actually work
/:cl:


Sorrowfulwinds:
Whoops forgot to update base. Also I apparently deleted some archive files upon review. Going to fix that real quick.

Sorrowfulwinds:
Good to go. Confirmed the game compiles and launches.

silicons (approved):

PR #7113, opened 2025-04-24
Bundle of fixes

MarsM0nd:
## About The Pull Request

A few fixes ran into messing around with the code, every logical unit is one commit.

## Why It's Good For The Game

People aren't unable to escape some storage, but not other. (Logic wise, that probably could be handed further to the object, to have people not escape something like locked secure briefcases or so, but this is just a quick fix.)
Vore resisting works again, besides slightly less special logic for it.
Augment cooldown actually works, besides splitting the usability check from the action proc, less duplication
Fancy boxes don't get/spawn invisible.
Runtime fix for the resonate analyzer (available as augment implant), technically usable.
Plant analyzer doesn't seem to be bigger than the others, so put into the same size/weightclass

## Changelog

:cl:
fix: Able to escape every storage
fix: vore resist belly interactions
fix: augment cooldowns
fix: Cigar, chocolate, and crayon/marker/chalk boxes don't become (partly) invisible
fix: resonate analyzer can be used
tweak: plant analyzer is now small
/:cl:


code/modules/vore/eating/belly_obj_vr.dm
```diff
@@ -518,10 +518,11 @@
 	return null
 
 //Handle a mob struggling
-// Called from /mob/living/carbon/relaymove()
-/obj/belly/proc/relay_resist(var/mob/living/R)
+// Called from /mob/living/proc/process_resist()
+/obj/belly/contents_resist(var/mob/living/R)
```
MarsM0nd, line 522: As far as I can tell, the linter is unhappy with this proc, only change is that it's now related to a proc that is annotated to not sleep. It's called from where it should have been (if the logic weren't broken, returning before reaching it), so if it's just an expectation, and not a functional change, it seems to be fine?

silicons (approved):

code/modules/vore/eating/belly_obj_vr.dm
silicons: oh
hm
let me check in a moment

BlueWildrose: > oh hm let me check in a moment

Have you checked?

TheLordME (approved):

silicons (approved):

PR #7146, opened 2025-05-08
removes most species restrictions from suits

silicons:
we use auto adapting bodysets now; cyclers aren't going to be removed anytime soon but we're no longer in the era where we're going to restrict humanoids from wearing.. humanoid suits.


lolman360:
you should lock phoronoid suits to phoronoid 🙏

silicons:
> you should lock phoronoid suits to phoronoid 🙏

why?
are they physically unable to be worn by humans from their shape?

TheLordME (approved):

PR #7148, opened 2025-05-08
Update icon parent first

MarsM0nd:
## About The Pull Request

I just ctrl+shift+f for ``/update_icon_state(``, and manually added ``.=..()`` at the beginning if it didn't have that. ``return ..()`` was also removed if at the end of the function, otherwise turned to plain returns. Tested only so far it compiles, and that it should be the correct logic instead.

## Why It's Good For The Game

Objects should have the icon_state and icon sheet matching, which is best known by the full path, than the parent-> modify the icon_state after the parent already was called. That likely is the cause for "Emergency Core Eject"s to be invisible, as ``/obj/machinery/button/remote/driver`` use ``'icons/obj/objects.dmi'``, but their parent uses ``'icons/obj/stationobjs.dmi'``, and sets it to a state only found there.

## Changelog

:cl:
refactor: update_icon_state always call parent first
fix: objects should have their intended sprite from the refactored function
/:cl:

### DIFF
--- a/code/game/objects/items/id_cards/contractor_ids.dm
+++ b/code/game/objects/items/id_cards/contractor_ids.dm
@@ -35,7 +35,7 @@
 			icon_state = "chit_purple"
 		else
 			icon_state = "chit"
-	return 0
+	return 0 //TODO: Why 0, not parent?
 
 /obj/item/card/id/contractor/Initialize(mapload)
 	. = ..()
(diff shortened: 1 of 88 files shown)


code/game/objects/items/id_cards/contractor_ids.dm
```diff
@@ -35,7 +35,7 @@
 			icon_state = "chit_purple"
 		else
 			icon_state = "chit"
-	return 0
+	return 0 //TODO: Why 0, not parent?
```
MarsM0nd, line 38: As the comment says, only stumbled upon it doing the refactor, unsure if that was an intentional change if someone knows?

MarsM0nd:
Will get a merge conflict with #7113, as that fixed the same issues for storage/fancy specifically.

PR #7182, opened 2025-05-29
refactors whitelist to toml config

silicons:
as name implies

whitelist now works off of species (u)ids

code should be a bit cleaner too


code/game/world.dm
```diff
@@ -121,7 +121,7 @@ GLOBAL_LIST(topic_status_cache)
 	#ifdef UNIT_TESTS
 	HandleTestRun()
 	#endif
-
+
```
Zandario, line 124:
```suggestion

```

Zandario (changes requested):
You made one mistake and now I shall force your hand :)

Zandario (approved):
What a blessed and cursed thing to have, surely this won't bite us in the ass later

TheLordME (approved):

PR #7250, opened 2025-07-19
Make showing status in who verb on by default.

CaptionAlkali:
## About The Pull Request
Make showing status in who verb on by default.

## Why It's Good For The Game
Will make it clearer who's AFK, who's playing, who's observing, etc. without making people dig around in settings t turn it on specifically.


silicons:
The hilarious thing is I disabled the default on purpose.

Could you explain why making it clearer which players, specifically, are playing, is good for the server game and community?
I'm just checking here; downstreams are obviously able to just override this. The disable was due to latent problems on citrp specifically.

PR #7258, opened 2025-07-25
gives engineering a gun

lolman360:
coilgun for engineering. like the already extant improvised coilgun but not: 

1. boring
2. unusable (15% chance to explode in your face)

now:
different coils have different effects
and also gas cooling. it needs to be cold enough

:cl:
add: engineering coilgun 2
/:cl:


Putnam3145:
does not compile

PR #7261, opened 2025-08-02
adds architecture for new science system

silicons:
tl;dr new science system involving servers and mainframes.

this is an alternate attempt at making techwebs that isn't just timegates, instead making science into a proper system that will then be given to offmaps as we go.


code/controllers/subsystem/uniqueness.dm
```diff
@@ -0,0 +1,119 @@
+//* This file is explicitly licensed under the MIT license. *//
+//* Copyright (c) 2025 Citadel Station Developers           *//
+
+/**
+ * Nebula calls this a repository, I call this a subsystem.
+ * This is in charge of assigning non-duplicating IDs to things.
+ * * Also in charge of round and map mangling.
+ */
+SUBSYSTEM_DEF(uniqueness)
```
silicons, line 9: SSmapping has an obfuscation module but it doesn't really belong in mapping. We should move things over here as we go.

code/modules/research/experimental/computers/research_console.dm
```diff
@@ -0,0 +1,134 @@
+//* This file is explicitly licensed under the MIT license. *//
+//* Copyright (c) 2025 Citadel Station Developers           *//
+
+// TODO: finish this file!
```
silicons, line 4: this file (along with peripherals) are not ticked right now
they'll come in another PR
we're not replacing research just yet with this PR. i just wanted to get the new system in codebase and have working datastructures for things like node / design projection, what is the important part as the new system contains semantics for designs/tech being stored on separate servers from the actual 'mainframe' that hosts the techweb.

PR #7286, opened 2025-08-27
refactors click handling

silicons:
refactors click handling to be more modular, as we have rigsuits and mecha presumably on the way and I didn't feel like patching like four versions of the clickchain for this stuff.


TheLordME (dismissed):
whaaahoo

PR #7293, opened 2025-09-02
fixes keybindings unset issue (or rather, adds a catch to fix it automatically after a few seconds)

silicons:
adds a new client stability test for keybindings being set
also, kicks old hooks to legacy hooks because i hate untyped hooks and most of them shouldn't really be hooks imho

### DIFF
--- a/code/modules/client/game_preferences/middleware/keybindings.dm
+++ b/code/modules/client/game_preferences/middleware/keybindings.dm
@@ -13,14 +13,8 @@
 /datum/game_preference_middleware/keybindings/handle_reset(datum/game_preferences/prefs)
 	. = ..()
 
-	prefs.keybindings = list()
-	// don't change their hotkey mode.. unless it was never there.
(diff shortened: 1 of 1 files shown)


code/controllers/hooks.dm
```diff
@@ -1,39 +1,58 @@
+//* This file is explicitly licensed under the MIT license. *//
+//* Copyright (c) 2025 Citadel Station Developers           *//
+
 /**
- * @file hooks.dm
- * Implements hooks, a simple way to run code on pre-defined events.
+ * Typed hooks system.
```
silicons, line 5: this is the key of why i'm replacing the old one, btw

i don't like untyped procs at all. or weird 'call all procs on this datum'. the byond compiler won't save you most of the time but the times it can is the reason why i want typed procs with parents. that, or, atleast the linter can do something about it & find all implementations work.

TheLordME (approved):

PR #7294, opened 2025-09-04
fixes mech control consoles lagging the server

silicons:
but why

### DIFF
--- a/code/modules/vehicles/sealed/mecha/mecha_control_console.dm
+++ b/code/modules/vehicles/sealed/mecha/mecha_control_console.dm
@@ -28,7 +28,7 @@
 	var/list/data = ..()
 
 	data["beacons"] = list()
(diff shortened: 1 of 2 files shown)


code/modules/vehicles/sealed/mecha/mecha_control_console.dm
```diff
@@ -28,7 +28,7 @@
 	var/list/data = ..()
 
 	data["beacons"] = list()
-	for(var/obj/item/vehicle_tracking_beacon/TR in world)
+	for(var/obj/item/vehicle_tracking_beacon/TR in GLOB.vehicle_tracking_beacons)
```
Putnam3145, line 31: GLOB.vehicle_tracking_beacons is going to only contain /obj/item/vehicle_tracking_beacon anyway, so an `as anything` (or whatever you do for that) is appropriate here

TheLordME (approved):

PR #7299, opened 2025-09-06
A couple minor fixes

Shadowtail117:
## About The Pull Request

- If you start a `me` message with a comma or an apostrophe, there will no longer be a space between your name and the message. I.e. "**John Trasen** , being hungry, eats." -> "**John Trasen**, being hungry, eats."
- Fixes reversed logic with dropping things with a broken hand. Your character will now correctly cry in pain when trying to do this.
- Also potentially fixes reversed logic with collapsing with a broken leg and screaming in pain.
- In both of the above cases, it will deal halloss (pain) damage, since using a broken bone tends to hurt a lot.

## Why It's Good For The Game

Small fix and improvement.

## Changelog

:cl:
fix: Fixed erroneous space when starting a me message with a comma or apostrophe.
fix: Your character will now cry in pain when trying to hold something with a broken hand.
/:cl:

### DIFF
--- a/code/_globals/regexes.dm
+++ b/code/_globals/regexes.dm
@@ -21,4 +21,4 @@ GLOBAL_PROTECT(filename_forbidden_chars)
 GLOBAL_DATUM_INIT(multi_space_splitter, /regex, regex("\[ \]+"))
 
 //Finds if a string started with "," or "'", used for me messages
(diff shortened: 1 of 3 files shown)


Putnam3145:
It's really, really pedantic, but the apostrophe/comma change is an improvement/addition, not a fix. The fix is only for bugs; this was the me verb working as intended, just not having some functionality that might be desired.

PR #7322, opened 2025-10-04
Fix space dust event

Sorrowfulwinds:
## About The Pull Request
Space dust has had an incorrect icon_state for 7+ years, this makes it visible again. Updates the space_dust event to actually work if triggered by the GM events. It was using specific strings while being fed integer defines instead.

Cleans up some older code, a commented out call to space_dust that is older than the hard sync and can just go away. Removes space_dust not damaging singulo engine equipment since we don't use the singulo and its a skill issue if your singulo gets sniped by space dust.

I think the entire event system on its own may have issues actually turning on but this event can at least be manually triggered by GMs and work now.

## Why It's Good For The Game
Lowers legacy tech debt, increases shareholder value.

## Changelog
-Fix the icon for space dust.
-Modernize severity value to use EVENT_LEVEL_* defines. 
-Add EVENT_LEVEL_EXTREME define for a severity 4 choice.
-Remove old code for protecting singulo equipment. 
-Remove commented out code for dust swarm older than the hardsync. 
-Update gm event system to correctly call dust_swarm.

:cl:
fix: fixed space dust
/:cl:

### DIFF
--- a/code/__DEFINES/misc.dm
+++ b/code/__DEFINES/misc.dm
@@ -122,6 +122,7 @@ Will print: "/mob/living/carbon/human/death" (you can optionally embed it in a s
 #define EVENT_LEVEL_MUNDANE		1
 #define EVENT_LEVEL_MODERATE	2
 #define EVENT_LEVEL_MAJOR		3
+#define EVENT_LEVEL_EXTREME		4
 
 /// General-purpose life speed define for plants.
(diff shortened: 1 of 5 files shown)


TheLordME (approved):
Why are there different dust numbers and dust types?

silicons (approved):

PR #7377, opened 2025-10-26
Z-Mimic: Speculative glass layering fix

Lohikar:
changes:
- Wide mobs standing on the edge of a glass turf should no longer layer under the glass turf.


deathride58:
hell yeah gamer

TheLordME (approved):

PR #7585, opened 2026-05-13
Saw traps no longer automatically chop off legs and feet

Shadowtail117:
## About The Pull Request

Changed the saw traps to apply their damage directly to their targeted limb instead of immediately severing it.

This PR has been tested locally.

## Why It's Good For The Game

The saw traps in some exploration areas, for some reason, automatically sever your feet/legs when stepped on, on top of applying their actual damage. This seems overly harsh especially given the chaotic nature of exploration, so I have changed them to simply apply their damage to a leg/foot when stepped on. Their damage is enough to significantly injure someone with even a couple steps, and repeatedly walking into them if you are careless will eventually break/sever the limb anyway.

## Changelog
:cl:
balance: Saw traps will no longer immediately sever limbs when stepped on.
/:cl:

### DIFF
--- a/code/game/objects/effects/traps.dm
+++ b/code/game/objects/effects/traps.dm
@@ -611,8 +611,7 @@ if (istype(AM, /mob/living))
 		var/list/target_limbs = list(BP_L_LEG, BP_R_LEG, BP_L_FOOT, BP_R_FOOT)
 		var/selected = pick(target_limbs)
 		var/obj/item/organ/external/target = M.get_organ(selected)
-		M.apply_damage(damage, DAMAGE_TYPE_BRUTE)
-		target.droplimb()
+		target.inflict_bodypart_damage(damage, 0, DAMAGE_MODE_SHARP | DAMAGE_MODE_SHRED, "spinning saw blade")
 		M.visible_message("<span class='danger'>[M] is ripped by the whirling sawblades!</span>", \
 						"<span class='userdanger'>You are ripped open by the whirling sawblades!</span>")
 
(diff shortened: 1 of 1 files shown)


AlphaM01:
Is it really "overly harsh" or are you upset your character hit one?
Exploration is not actually that "chaotic", your character was just ignoring any attempt at caution and running in ahead of everyone despite being told not ICly not to do that.
I think consequences are good, actually.

Shadowtail117:
> Is it really "overly harsh" or are you upset your character hit one? Exploration is not actually that "chaotic", your character was just ignoring any attempt at caution and running in ahead of everyone despite being told not ICly not to do that. I think consequences are good, actually.

This trap (and actually another "bladed pillar" trap that I just addressed) is one of the only cases in the entire codebase where a limb is straight-up removed from a mob. The literal only other time this proc is called is during amputation surgery (and a niche case as a promethean reaction to corophizine). In any other case where a limb is severed, it is as a result of damage sustained to it, which I will wager is the intended avenue of having a limb traumatically amputated.

Given how debilitating the outright removal of a limb is, and given that there is quite literally no amount of armor that can protect against it, the consequences of this trap do feel out of proportion. I am not saying there shouldn't be consequences, obviously, but they need to be reasonable. Applying the trap's stated amount of damage to the target limb instead of directly removing it seems fair and in line with not only the other traps that exist but the overall health system as well.

TheLordME (approved):

silicons (approved):

PR #7587, opened 2026-05-23
scori crafting stuff

GySgtMurphy:
## About The Pull Request
Touches scorian materials a bit. Buffed bronze to make it more effective as a weapon than bone, nerfed bone's tech levels since they were ridiculous and removed hematite processing from magma forges in favor of crafting lead bars into metal rods.

## Why It's Good For The Game

Bronze weapons should be better than bone weapons. Metal rods change was something charles_wedge expressed a desire for. You should not get to skip half the tech tree from fighting one goliath.

## Changelog
:cl:
Nerfed tech levels on bones
Made scorians build houses out of lead instead of primitive steel
Made bronze slightly more effective for weapons
/:cl:


code/modules/materials/definitions/metals/lead.dm
```diff
@@ -21,3 +21,12 @@
 	nullification = MATERIAL_RESISTANCE_LOW
 
 	material_constraints = MATERIAL_CONSTRAINT_RIGID
+
+/datum/prototype/material/lead/generate_recipes()
+	. = ..()
+	. += create_stack_recipe_datum(
```
TheLordME, line 27: this allows for lead to steel conversion, I need to check how feasable it is, but maybe a subtype of /obj/item/stack/rods eg /obj/item/stack/rods/lead that converts back to lead might make sense here

silicons (approved):

PR #7588, opened 2026-05-23
Using the ballista now triggers an ATC message

Shadowtail117:
## About The Pull Request

Using the obstruction removal ballista will now trigger the ATC to broadcast an advisory about it within the next 20-100 seconds. This includes giving the approximate coordinates of where the beam came from.

This PR was tested locally.

## Why It's Good For The Game

Using the ballista is a bit of a big deal since it's basically a big space cannon. It makes sense that the ATC would warn other vessels about it, especially considering it could be used offensively in the wrong hands. Plus, it helps make the ATC a bit more meaningful and not just fluff.

## Changelog
:cl:
add: Using the obstruction removal ballista will now prompt the local ATC to issue a travel advisory about it.
/:cl:

### DIFF
--- a/code/modules/overmap/legacy/disperser/disperser_fire.dm
+++ b/code/modules/overmap/legacy/disperser/disperser_fire.dm
@@ -46,6 +46,11 @@
 	handle_overbeam()
 	qdel(atomcharge)
 
+	spawn(rand(20, 100) SECONDS)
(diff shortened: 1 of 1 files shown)


TheLordME (commented):
not quite sure how I feel about atc magically knowing about the fire state fo all ballistas but we can test merge it

TheLordME (approved):

